PR #19, opened 2026-07-10
fix CJK text width estimation

### LINKED ISSUE #14: [P2] CJK text width estimation may be inaccurate
## Problem

The SKILL.md states: "CJK text is measured at double width automatically". However, when building a diagram with Chinese labels (写入, 审批通过, 注入提示词, etc.), I encountered multiple label-vs-component overlap errors that may have been caused by inaccurate width estimation.

## Observation

The `textUnits()` function in `shared/utils.mjs` is responsible for measuring text width. If CJK characters are not properly detected and counted at double width, the estimated label width will be too small, causing labels to overflow their components and overlap adjacent elements.

## Suggested Investigation

1. Verify `textUnits()` correctly detects CJK Unicode ranges (U+4E00–U+9FFF, U+3400–U+4DBF, U+3000–U+303F, etc.)
2. Add test cases with pure CJK labels, mixed CJK+ASCII labels, and full-width punctuation
3. Consider exposing a `--debug-measure` flag that outputs estimated vs actual text widths for each label

## Context

- archify v2.6, architecture mode
- Diagram with ~18 components, ~60% Chinese labels
- 3 rounds of iteration to resolve all label overlap issues
- Some overlaps may have been caused by coordinate placement, but CJK width underestimation likely contributed

reallav0:
## Summary

- replace the incomplete hard-coded width ranges with Unicode script-aware CJK detection
- cover supplementary Kana, Tangut, Nushu, Khitan, and vertical CJK punctuation
- preserve single-width handling for halfwidth Katakana and Hangul forms
- add pure CJK, mixed ASCII/CJK, punctuation, and supplementary-plane regression tests
- rebuild and smoke-test the distributed `archify.zip`

## Root cause

The previous classifier covered common BMP Han ranges and supplementary unified ideographs, but skipped several supplementary East Asian scripts and U+FE10–U+FE19 vertical punctuation. Those glyphs were counted as one unit, which could underestimate label and collision-box widths.

## Validation

- `npm run check:validators`
- `node --test test/*.test.mjs` — 111 tests passed
- packaged `archify --help`, `archify doctor`, and architecture validation smoke checks

Closes #14


tt-a1i:
Thanks for the PR — looked at this carefully against #14.

For the Chinese labels called out in that issue (`写入`, `审批通过`, `注入提示词`, mixed ASCII/CJK), `textUnits` already returns the same widths on `main` as with this change. So this doesn't appear to fix the reported overlap symptom.

What it does change is broader Unicode coverage (vertical punctuation, Tangut/Nushu/Khitan, etc.). That also shifts width for ~tens of thousands of code points, including some that shouldn't be double-width in our monospace stack (e.g. middle dot `·` U+00B7 goes 1 → 2).

So I'm going to hold off on merging this as a `#14` fix. If you want to continue:

1. Please drop `Closes #14` (or retarget) — this shouldn't close that issue.
2. If the goal is still #14, a repro diagram where labels overflow under current `textUnits` would help a lot more than expanding script coverage.
3. If the goal is supplementary-script / vertical-punctuation coverage on its own, please keep the diff narrowly scoped and avoid over-counting non-fullwidth characters.

Happy to take another look with either a repro or a tightened version.

PR #21, opened 2026-07-11
fix: harden validation and packaged CLI

ShiroKSH:
## Summary

- honor explicit architecture `pos` overrides in grid layouts instead of treating them as a shared `undefined,undefined` cell
- reject zero and negative component dimensions in both the schema and renderer backstop
- ship the renderer used by `archify examples`, verify it in `doctor`, and cover an installed dependency-free skill
- validate diagram types before allocating temporary output directories
- make generated-validator and golden-render freshness checks insensitive to CRLF/LF checkout differences

## Root cause and impact

Grid validation only skipped components that had both `pos` and `row`/`col`, despite the documented `pos`-wins contract. Architecture sizes reused the unconstrained coordinate tuple, allowing invisible or invalid SVG rectangles. The public `examples` command called a script excluded from the release archive, while byte-for-byte freshness checks falsely failed on Windows checkouts. Unknown validation types also exited from inside the temporary-directory lifetime and bypassed cleanup.

The fixes stay at those ownership boundaries and leave renderer layout behavior unchanged for valid inputs. The dependency-free archive is rebuilt with the shipped example renderer and standalone validators.

## Checks

- `npm test` — all checks passed (114 Node tests plus golden/schema checks)
- dependency-free archive smoke — `doctor`, `examples`, and validation for all five diagram types
- archive content verification — 35 expected files, with development dependencies and tests excluded
- `git diff --check`


tt-a1i:
Hey, thanks a lot for this — really solid catch on several real bugs.

I went through it carefully and locally `npm test` is green (117), and the committed `archify.zip` matches a fresh `build-zip.sh` rebuild. The grid `pos`-wins fix matches the docs, moving `render-examples` out of `test/` is a genuine packaging bug (zip excludes `test/`), and the unknown-type temp-dir leak on `validate` is real too (`process.exit` skips `finally`). Nice work.

A few small things I'd like cleaned up before we merge:

1. **The new size cases in `layout-rules.test.mjs` aren't actually hitting the renderer backstop.** With `size: [0, 60]` etc., AJV fails first (`must be > 0`), so those tests never reach the `greater than 0` message in `render-architecture.mjs`. Either adjust the tests so they exercise the layout rule (if there's a path that bypasses schema), or drop them from that file and keep the schema coverage elsewhere — right now they give a bit of false confidence.

2. **CI `package-smoke` still doesn't run `archify examples`.** Since fixing the installed-skill `examples` path is a big part of this PR, it'd be great to add that one command to `.github/workflows/ci.yml` so we don't regress it again.

3. **Optional:** a short CHANGELOG note under Fixed would be nice, but not a blocker if you'd rather skip it.

No rush — happy to merge once 1 and 2 are sorted. Thanks again for digging into the packaging + validation edges 🙏

ShiroKSH:
Addressed in `7dea2d1`:

- moved the three non-positive `size` cases out of `layout-rules.test.mjs` and into the golden harness's explicit schema-enforcement section, so they no longer imply renderer-backstop coverage
- added `node bin/archify.mjs examples` to the dependency-free `package-smoke` job

`npm test` passes: 114 Node tests plus all golden/schema checks. I left the optional changelog entry to the maintainer-side merge workflow.

tt-a1i:
LGTM — thanks for cleaning up the test boundary and the package-smoke coverage. Merging now.

kbumrungcharoen-cloud (changes requested):
RequestChang

Issue #24, opened 2026-07-20
Architecture mode: connections can route through unrelated components — render/validate accept it silently


hee-c:
## Summary

In architecture mode, the built-in router can produce a path that passes **through** an unrelated component's bounding box. Both `render` and `validate` accept the result without any error or warning.

Because arrows are painted before component boxes (correct per the design system), the offending segments hide behind the obstacle component and the arrow visually appears to **originate from the obstacle**, misrepresenting the topology.

Workflow mode already guards against this since v2.8 ("workflow rendering now rejects edges that cross through non-endpoint nodes" — status update in #10). Architecture mode has no equivalent check. This report is a concrete, minimal real-world case for the "shared obstacle detection across architecture/dataflow/lifecycle" follow-up mentioned there.

## Repro

`repro.architecture.json` — 3 components, 1 connection, no route hints:

```json
{
  "schema_version": 1,
  "diagram_type": "architecture",
  "meta": {
    "title": "Repro",
    "subtitle": "arrow passes through an unrelated component",
    "output": "repro.html"
  },
  "components": [
    { "id": "api", "type": "backend", "label": "API", "pos": [400, 280], "size": [160, 76] },
    { "id": "cache", "type": "database", "label": "Cache", "pos": [645, 130], "size": [130, 60] },
    { "id": "queue", "type": "cloud", "label": "Queue", "pos": [880, 130] }
  ],
  "connections": [
    { "from": "api", "to": "queue", "variant": "dashed" }
  ],
  "cards": []
}
```

```bash
node bin/archify.mjs render architecture repro.architecture.json repro.html   # succeeds
node bin/archify.mjs validate architecture repro.architecture.json --json    # passes
node bin/archify.mjs check repro.html                                        # ok: true
```

## Actual

The only connection in the file is `api → queue`, but it renders as if it were `cache → queue`:

![api → queue rendered as if cache → queue](https://gist.githubusercontent.com/hee-c/34ac57b7b3d546277a6143945c647d6c/raw/39e1f56b54db10a1de06649a42bbe9225c96b000/repro-before.png)

Generated path:

```
M 560 318 L 712 318 Q 720 318 720 310 L 720 168 Q 720 160 728 160 L 880 160
```

[...]

PR #30, opened 2026-07-22
test: cover auto-routed pass-through in architecture Clean Flow gate

hee-c:
Refs #24 — follow-up to the maintainer's reply there: "Your exact `auto`-route case is better regression coverage than the simpler case we currently have, so we'd like to include it before release."

## What

Two regression tests in `archify/test/layout-rules.test.mjs`, next to the existing Clean Flow Gate cases:

1. **`architecture: Clean Flow Gate rejects an auto-routed connection through an unrelated component`** — the exact #24 repro: no route hints at all; the auto-router itself picks a detour whose corner sits inside an unrelated component, so the arrow hides behind the obstacle box and reads as "cache → queue" instead of "api → queue". The existing architecture case exercises an author-requested `route: 'straight'`; this one asserts the gate judges the router's own output.
2. **`architecture: via waypoints routed around the obstacle satisfy the Clean Flow Gate`** — the same document with the workaround from the issue (`fromSide`/`toSide` + `via` above the obstacle) renders clean, locking in that the gate rejects the bad route, not the topology.

## Verification

- `npm test` from `archify/`: 352 pass, 0 fail (Node v24.11.0)
- The reject test genuinely depends on the gate: the same document renders successfully today when `quality_profile` is omitted (which skips the gate), so the assertion would fail if the gate regressed.

## Notes

Test-only change — no renderer code touched — so it should compose cleanly with the in-flight Clean Flow work; feel free to cherry-pick into that branch if it lands first.

### DIFF
--- a/archify/test/layout-rules.test.mjs
+++ b/archify/test/layout-rules.test.mjs
@@ -355,6 +355,44 @@ test('lifecycle: Clean Flow Gate rejects a transition through an unrelated state
   assert.match(stderr, /col\/yOffset/);
 });
 
+// Regression coverage for #24. The explicit-straight case above covers a
+// route the author asked for; here the document gives NO route hints and the
+// auto-router itself picks a detour whose corner sits inside an unrelated
+// component — the arrow hides behind the obstacle box and reads as
+// "cache -> queue" instead of "api -> queue". The gate must judge the
+// router's own output, not just author-requested routes.
+function passThroughRepro(connection) {
+  return {
+    schema_version: 1,
+    diagram_type: 'architecture',
+    meta: { title: 'Repro', quality_profile: 'standard' },
+    components: [
+      { id: 'api', type: 'backend', label: 'API', pos: [400, 280], size: [160, 76] },
+      { id: 'cache', type: 'database', label: 'Cache', pos: [645, 130], size: [130, 60] },
+      { id: 'queue', type: 'cloud', label: 'Queue', pos: [880, 130] },
+    ],
+    connections: [connection],
+  };
+}
+
+test('architecture: Clean Flow Gate rejects an auto-routed connection through an unrelated component', () => {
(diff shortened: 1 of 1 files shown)


tt-a1i (approved):
Thank you — this is exactly the real-world regression coverage requested in #24. The auto-routed failure and explicit-`via` passing case are both focused, use production rendering, and compose cleanly with the shared Clean Flow work. I also reproduced the important remaining gap locally: profile-less default delivery still accepts the misleading route, so I am extending the invariant to default delivery in the stability branch while preserving your regression pair.

tt-a1i:
Thank you — this exact #24 reproduction was genuinely useful and directly shaped the production fix.

Main now includes the same three-component topology as a regression, but the expected behavior has advanced: `auto` first keeps the existing H-V-H route when it is clear, then tries the complementary bounded V-H-V route when the first crosses an unrelated component. Explicit `via` is still covered, explicit routing remains authoritative, and both auto candidates being blocked still fails closed through Clean Flow.

That shipped in #33 with all five CI checks green. Because this PR locks in the superseded hard-failure expectation, I’m closing it rather than asking you to rework a test-only contribution against a behavior that is now implemented. Thanks again for isolating the real topology.

hee-c:
Completely fair call — the reroute-first behavior in #33 is strictly better than the hard-failure expectation this PR locked in, and seeing the exact topology land as regression coverage in main is the best outcome #24 could have had. Thanks for the thoughtful close and the warm welcome — happy to keep contributing.

tt-a1i:
Thank you again — your report and regression work directly improved Archify, even though this PR was not merged verbatim. The care you put into isolating the problem, validating the behavior, and keeping the contribution focused was genuinely valuable.

We'd be very happy to have you continue contributing and help maintain Archify with us. Whether that means improving routing and validation, reviewing issues, testing real-world repositories, or shaping future behavior, you're warmly welcome to join in wherever you're interested.

Really glad to have you here, and looking forward to building more together.

PR #72, opened 2026-08-14
feat(drawio): export Archify diagrams as editable draw.io files

kidoln:
## Problem and value

Archify artifacts are self-contained HTML, but readers who want to keep editing a diagram in standard tooling had no way out. This adds draw.io export to every artifact, in two variants:

- **draw.io** — editable: draw.io-native shapes per component kind, real source/target edge bindings (dragging a box keeps its connections), authored route waypoints preserved.
- **draw.io strict** — 1:1: exact Archify corner radii, composited fill/stroke colors, stroke widths, dash patterns, and edge anchor points, always resolved to the light palette so the file reads correctly under any draw.io theme.

Sequence diagrams route through a dedicated UML-lifeline builder so the timeline survives export: participants become `shape=umlLifeline` cells spanning header→foot, activation bars become lifeline children, and each message's authored y survives as an exit/entry fraction. A headless path is included: `archify export-drawio <type> <input.json> [output] [--strict] [--json]`.

## Scope

- What changed:
  - New converter `archify/renderers/shared/svg-to-drawio.mjs` (zero-dependency regex parsing of the self-describing SVG; the SVG is the sole geometry source, coordinates never recomputed) plus `archify/scripts/export-drawio.mjs` and the `export-drawio` CLI subcommand.
  - Viewer Export menu gains the two draw.io entries (`archify/assets/template.html`), mirroring the Node converter; artifacts now carry `data-diagram-type` so the viewer can dispatch sequence exports.
  - `render-sequence.mjs` stamps activation bars with `data-graph-role="activation"` + participant identity.
  - Examples: all 12 sample HTMLs re-rendered with the new shell; a `.drawio` + `-strict.drawio` pair for each of the 10 static samples; the five per-type demo pairs refreshed (byte-identical); gallery artifacts and showcase proof regenerated; `archify.zip` rebuilt.
  - Post-review hardening: consistent attribute escaping on edge labels, frame-id-based (not positional) boundary↔wraps matching, sequence foot fallback against empty participant sets, `-strict.drawio` default naming, documented attribute-order contract, and a parity test that extracts the viewer converter from the live template and asserts byte-identical output with the Node converter.
  - Existing SVG/PNG/JPEG/WebP/share-card export paths, rendering/layout output for existing diagrams, JSON IR schemas, and the viewer's canonical-state receipt semantics.
  - No `mxfile`/`diagram` wrapper (draw.io opens the bare `mxGraphModel` directly) and no new runtime dependencies.
  - Delta (before/after compare) artifacts are not draw.io-exportable: their states are viewer-driven, and static SVG extraction yields an empty shell — exporting one would ship a broken diagram.
- No unrelated changes: the only files touched outside the exporter surface are regenerated artifacts (examples, gallery, showcase proof, zip).

## Stability impact

- Compatibility and migration risk: additive only. No schema or typed-JSON contract changes; artifacts without the new menu entries remain fully functional. `archify check` passes on every re-rendered sample.
- Renderer, validator, package, or generated-artifact risk: renderers gain one attribute (activation identity) and the HTML shell gains `data-diagram-type`; both are consumed only by the exporter. The packaged skill (`archify.zip`) is rebuilt so installed skills match the source tree; release identity still checks `2.14.0`.
- Failure behavior and rollback path: unresolvable edge endpoints are skipped (never dangling cell references); a missing SVG raises with a clear message; viewer failures surface through the existing toast/alert + `data-last-export-*` receipt attributes; the CLI exits non-zero. Rollback is reverting the branch — no data or contract migrations.

## Tests run

- `cd archify && npm test` → **666/666 pass** (validators, release identity, golden, full suite; includes 39 drawio tests: parsing units, builder units, 5-type CLI integrations, strict/sequence regressions, viewer↔Node parity).
- `node --test archify/test/export-drawio.test.mjs` → 39/39 pass.
- `./scripts/build-zip.sh archify.zip && node scripts/check-release-identity.mjs` → `release identity ok: 2.14.0`.
- `xmllint --noout` on all 30 example `.drawio` files → all well-formed.
- After the hardening fixes, all ten sample pairs re-exported and diffed against the committed files → byte-identical (no unintended output change).

## Visual evidence

Visual review **passed** during development: each feature commit documents its draw.io-viewer comparison (strict always-light palette under light and dark themes; sequence timeline, activations, segments, labels, and return dashes restored via UML lifelines). The committed `examples/*.drawio` pairs can be opened directly for re-verification; screenshots can be attached on request.

## Generated artifacts

- `examples/`: 12 sample HTMLs re-rendered, 20 new per-sample `.drawio` pairs, receipt for the delta sample.
- `docs/`: gallery artifacts (11), `docs/gallery.html`, `docs/gallery/manifest.json`, showcase proof (`docs/assets/archify-live-proof.gif` + `.json`) — all rebuilt from the new shell.
- `archify.zip`: rebuilt in a dedicated commit; previously it did not ship the exporter at all.

## Checklist


tt-a1i (changes requested):
Thanks for taking on draw.io export - this capability is still absent from `main`, so I would like to keep moving it forward.

However, this branch has diverged too far to safely rebase as-is. It now contains 74 changed files and about 95% of the textual diff is generated output. A merge simulation produces seven direct conflicts, including `archify.zip`, gallery/proof assets, and delta artifacts. The package is still based on Archify 2.14.0 while `main` is already 2.15.0.

Could you rebuild the change from current `main`, either by replacing this branch or opening a clean follow-up PR? Ideally split the work into:

1. converter, CLI, and tests;
2. viewer integration and documentation;
3. regenerated examples, gallery, and ZIP only once, after the source is stable.

Before that version is ready, I found several correctness issues that need addressing:

- In `archify/renderers/shared/svg-to-drawio.mjs:1183-1186`, strict export writes a second `<mxGeometry as="geometry">` inside the edge's existing geometry. The sequence implementation at line 796 puts label coordinates on the single geometry element instead. XML parsing alone does not verify that diagrams.net applies the nested geometry, so this needs a real import or round-trip assertion.
- Current `main` supports authored brand marks, but the converter only reads and emits `id`, `kind`, `label`, `sublabel`, and `tag` (`svg-to-drawio.mjs:182-187`, `552-560`, `1108-1117`). Merging this implementation would silently drop brand identity, including in the advertised strict 1:1 export.
- The Node converter and the large inline viewer converter are hand-maintained copies. The parity test explicitly covers only one non-strict architecture artifact and excludes strict and sequence paths (`export-drawio.test.mjs:624-647`). Please derive both surfaces from one canonical implementation, or add parity coverage for all five diagram types, strict/non-strict, and browser/CLI paths.

Two smaller follow-ups:

- All boundaries currently use `parent="1"` while nodes are assigned only to their innermost boundary. In nested region/security-group diagrams, moving the outer region therefore does not move the inner group and its nodes.
- Please document the viewer export and `archify export-drawio` command in the English/Chinese README and packaged Skill docs.

The 39 draw.io-specific tests pass locally. The remaining request is mainly to reduce generated churn, rebuild the implementation semantically on current `main`, and add diagrams.net-level correctness coverage.

YunyueLi:
Thank you for the substantial work on editable draw.io export. The capability remains valuable and is still absent from `main`; the recommendation below is about creating a safe path to merge it, not rejecting the product direction.

The current branch should not be rebased or extended further. It is based on an older Archify release, now conflicts with `main`, changes 74 files, and contains a very large amount of generated output. Continuing to patch this branch would make it difficult to distinguish the converter's behavior from stale generated artifacts and conflict resolutions.

Please open a clean follow-up PR from the latest `main`, reference #72, and use this PR as implementation history. A staged delivery would make the work reviewable:

1. **Converter, CLI, and correctness tests.**

   Introduce one canonical SVG-to-draw.io conversion implementation and the `archify export-drawio` CLI seam. Keep viewer integration and bulk generated artifacts out of this first step. Add a real diagrams.net import or round-trip assertion for strict output; XML parsing alone cannot prove that diagrams.net applies edge geometry and label placement correctly.

2. **Viewer integration from the same implementation.**

   The browser export and Node CLI must not remain two independently maintained converters. Generate or share the browser implementation from the same canonical source, and add parity coverage for all five diagram types in strict and non-strict modes. Sequence diagrams need dedicated coverage for lifelines, message positions, activations, segment labels, and return-message dashes.

3. **Documentation and generated artifacts only after the source is accepted.**

   Document both the viewer export and CLI command in the English README, Chinese README, and packaged Skill. Regenerate examples, Gallery artifacts, proof assets, and `archify.zip` once from the final accepted source rather than carrying old generated output through the review.

The clean implementation should also meet these correctness requirements:

- emit a single valid `mxGeometry` contract per edge; do not rely on nested geometry elements that diagrams.net may ignore;
- preserve authored brand marks, node and relationship identity, labels, routing, dash semantics, and strict-mode geometry;
- preserve nested boundary hierarchy so moving an outer boundary also moves its inner boundaries and contained nodes;
- cover architecture, workflow, sequence, data-flow, and lifecycle with representative import-level fixtures;
- use the current package and release identity rather than rebuilding from the previous 2.14.0 state;
- keep the initial source diff focused and defer generated files until the behavior is stable.

Once the clean follow-up is open, the current PR can be closed in favor of it. This preserves the design work already completed while giving the feature a realistic route to approval and merge.

tt-a1i (changes requested):
Additional review of unchanged headdef87141869e158cb536252a881151d8a8f58178. The existing strict geometry/nested groups/brand/parity/documentation findings remain; this is a newly reproduced data-loss path.

[P1] Ensure the default draw.io destination can never equal the source (archify/scripts/export-drawio.mjs:56-58,82-83). The case-sensitive replace only handles lower-case .json/.html. With a valid IR file named diagram.JSON, or an extensionless valid IR file, `archify export-drawio architecture <input> --json` has no explicit output, computes outputPath===inputPath, exits0/ok:true, and overwrites the original JSON with draw.io XML. Both real CLI fixtures reproduced the loss. Derive a guaranteed distinct .drawio filename, apply input-alias protection to explicit and implicit outputs, and add preservation tests for uppercase/extensionless/symlink cases.

All39 draw.io tests pass and all64 ZIP payload files match canonical source; those checks do not exercise this default-path case. Native diagrams.net round-trip acceptance is not claimed. No source edits or merge.

tt-a1i:
Queue update: `def87141869e158cb536252a881151d8a8f58178` is unchanged from the data-loss review. The uppercase/extensionless implicit-output path must preserve the input, alongside explicit-output and symlink protections. The earlier converter parity, nested geometry/ownership and actual diagrams.net import concerns also remain unaddressed at this head.

Editable export remains useful. Please prepare the focused converter/CLI follow-up from current `dev`, reference this implementation, and use the canonical Viewer source when adding the browser surface. The earlier suggestion to start from `main` is superseded by our dev integration workflow. Keep this PR as history until a concrete replacement exists; do not close it merely because it is old. This update reuses the unchanged-head review and does not claim a fresh diagrams.net round trip. Author owns the next revision/replacement; maintainers will re-review it.

Issue #11, opened 2026-06-30
[P2] Expand component types (cache, queue, monitor, gateway, storage)


tt-a1i:
## Problem

The `componentType` enum currently has 7 types:
`frontend`, `backend`, `database`, `cloud`, `security`, `messagebus`, `external`

Real-world architecture diagrams often need more semantic granularity. For example:
- Redis/Memcached are semantically different from PostgreSQL — they are caches, not primary stores.
- SQS/RabbitMQ are queues, which are a subset of message buses with different visual expectations.
- Prometheus/Grafana/Datadog are monitoring systems.
- API Gateway/Kong/Envoy are gateways, distinct from generic backends.
- S3/GCS are object storage, semantically different from databases.

## Suggested Addition

```
cache       — Redis, Memcached, Varnish
queue       — SQS, RabbitMQ, Kafka (when used as queue)
monitor     — Prometheus, Grafana, Datadog, ELK
gateway     — API Gateway, Kong, Envoy proxy
storage     — S3, GCS, Blob Storage (object/file storage)
```

## Acceptance Criteria

- Each new type gets its own CSS class (`c-cache`, `c-queue`, etc.) with a distinct color in both themes.
- The legend auto-includes new types when used.
- Backward compatible — existing diagrams without these types continue to work.

tt-a1i:
Closing: no external replies. Can reopen if needed.

PR #91, opened 2026-08-19
fix(text-fit): count BMP emoji-presentation symbols as two units

NgoQuocViet2001:
## Problem and value

`textUnits` is the width model behind `fittedNodeFontSize` and `minimumNodeTextWidth`. It already counts supplementary-plane emoji as two units, but the symbols that carry emoji presentation in the BMP were still one:

```
"✅" -> 1     "⭐" -> 1     "⚡" -> 1     "⌛" -> 1     "⛔" -> 1
"😀" -> 2     "中" -> 2     "Ａ" -> 2
```

Those all render at the same square advance, so a node label like `✅ Done` was measured at roughly half the width it actually takes. `fittedNodeFontSize` then shrinks too little and `minimumNodeTextWidth` under-reports, which is precisely the failure `text-fit.mjs` says it exists to close — the text overruns its neighbours while the layout receipt still reads clean.

UAX #11 classifies this set (U+231A..U+2B55) as East Asian Wide, alongside the emoji blocks the regex already covers. Hangul Jamo Extended-A (U+A960..U+A97F) had the same gap while every other Hangul block was listed.

## Scope

- What changed: the `FULLWIDTH_RE` range list in `renderers/shared/utils.mjs`, plus tests.
- What deliberately did not change: `widthFactor`, `horizontalPadding`, any renderer, any schema, any authored JSON. The pre-existing over-coverage inside U+2E80..U+A4CF is left alone.
- No unrelated changes.

## Stability impact

- Compatibility and migration risk: none for typed JSON. Diagrams whose labels contain these symbols now measure wider, so their node text shrinks a little more or is reported as needing a wider box — which is the corrected behaviour. No checked-in example, recipe, or gallery source contains a character in the added ranges, so no generated artifact changes.
- Failure behavior: unchanged; this only moves a number the existing fit and validation paths already consume.
- Rollback path: revert the one line.

## Tests run

- `node --test test/geometry.test.mjs` → 53 pass, 0 fail (52 pre-existing + 1 new group).
- `npm test` → the set of failing tests is identical before and after this change (Windows-only symlink and packaging tests); pass count goes up by the new test.

New coverage asserts the added symbols count as two, that a real label (`✅ Done`) sums correctly, and that narrow neighbours in the same blocks — `→` (U+2192), `☎` (U+260E), `①` (U+2460) — stay at one unit, so the ranges cannot quietly widen further.

## Visual evidence

Not applicable — no authored artifact contains a character in the added ranges, so no rendered output changes.

## Generated artifacts

None. No source of a generated artifact changed.

## Checklist

- [x] I used a minimal focused change and preserved existing typed JSON behavior.
- [x] I added a regression test for the behavioral change.
- [x] I checked generated artifacts — no gallery, guide, README proof, or `archify.zip` source is affected.
- [x] No secrets, private repository content, or customer data in fixtures.


archify/renderers/shared/utils.mjs
```diff
@@ -165,7 +165,13 @@ export function applyTemplate(template, { title, subtitle, svg, cards, visualPre
...
+// Code points that take two columns of advance width: East Asian Wide and
+// Fullwidth per UAX #11. That takes in the BMP symbols carrying emoji
+// presentation (U+2705, U+2B50, U+26A1, U+231B, ...), which render at the same
+// square advance as the supplementary-plane emoji already listed here, and
+// Hangul Jamo Extended-A, wide like every other Hangul block. Spelled out as
+// ranges because V8 has no \p{East_Asian_Width=W} property escape.
+const FULLWIDTH_RE = /[\u1100-\u115F\u231A-\u231B\u2329-\u232A\u23E9-\u23EC\u23F0\u23F3\u25FD-\u25FE\u2614-\u2615\u2648-\u2653\u267F\u2693\u26A1\u26AA-\u26AB\u26BD-\u26BE\u26C4-\u26C5\u26CE\u26D4\u26EA\u26F2-\u26F3\u26F5\u26FA\u26FD\u2705\u270A-\u270B\u2728\u274C\u274E\u2753-\u2755\u2757\u2795-\u2797\u27B0\u27BF\u2B1B-\u2B1C\u2B50\u2B55\u2E80-\uA4CF\uA960-\uA97F\uAC00-\uD7A3\uF900-\uFAFF\uFE10-\uFE19\uFE30-\uFE6F\uFF01-\uFF60\uFFE0-\uFFE6\u{16FE0}-\u{18DFF}\u{1AFF0}-\u{1AFFF}\u{1B000}-\u{1B2FF}\u{1F000}-\u{1FAFF}\u{20000}-\u{3FFFD}]/u;
```
tt-a1i: Thanks for filling this BMP gap. There is one regression in the current code-point model: `textUnits()` still counts U+FE0F as one unit, so widening the base now changes common emoji-presentation sequences from two units to three. I reproduced `textUnits("⭐️") === 3`, `textUnits("✅️") === 3`, and `textUnits("☕️") === 3`, while the rendered glyph remains one square emoji; `✈️` happens to remain 2 only because its base is still counted as 1. Please handle VS15/VS16 sequences explicitly (rather than only extending the base-character regex) and add regression cases for `⭐️`, `✅️`, and `✈️`. Also, if this comment intends to track current UAX #11, please pin/document the Unicode version: Unicode 16/17 classify U+2630–U+2637 and U+268A–U+268F as Wide, but those ranges are absent, while U+A97D–U+A97F are currently unassigned and included by this range.

tt-a1i (changes requested):
Thank you for the focused fix and for adding boundary coverage around the BMP emoji-presentation symbols. The diagnosis is useful, the change is nicely scoped, and I confirmed the targeted geometry suite passes 53/53; the Node 18/20/22/24, package-smoke, and WebM CI lanes are also green. There are two correctness/release issues to address before merge:

1. `archify.zip` is now stale. `utils.mjs` is part of the distributed Skill runtime, so the archive must be rebuilt even though no checked-in rendered example changes. The current `zip-freshness` CI failure reproduces this exact mismatch.
2. Emoji variation-selector sequences need handling before widening their base code points; see the inline comment.

Once those are fixed and CI is green, this should be straightforward to re-review.

archify/renderers/shared/utils.mjs
NgoQuocViet2001: You're right, and thank you for reproducing it with numbers — that made it unambiguous. Fixed in the push above.

**The regression.** `textUnits` was counting U+FE0F as a unit of its own, so widening the base turned a two-unit sequence into three. Your three cases, before and after:

| | before | after |
|---|---|---|
| `⭐️` U+2B50 U+FE0F | 3 | 2 |
| `✅️` U+2705 U+FE0F | 3 | 2 |
| `☕️` U+2615 U+FE0F | 3 | 2 |
| `✈️` U+2708 U+FE0F | 2 | 2 |

**The rule, rather than the patch.** A variation selector carries no advance of its own — it re-presents the character before it. So the pair is now measured *from the selector*, not from the base:

- VS16 (U+FE0F) asks for emoji presentation → the pair renders as one square → 2
- VS15 (U+FE0E) asks for text presentation → renders narrow → 1
- U+FE00–U+FE0F never contributes width on its own

That is why I did not just special-case the widened bases. As you noted, `✈️` was landing on 2 by accident: its base is narrow and the selector added the second unit. Under the new rule it is 2 *because VS16 forces the square*, which is the same reason `⭐️` is 2 — and `⭐︎` (VS15) is now correctly 1, which the old code also got wrong in the other direction.

One deliberate call: a selector after a base that cannot take emoji presentation is malformed input, and I measure it wide. Over-measuring pads a box; under-measuring spills the label out of it, which is the failure this PR exists to prevent. It is commented in the source.

**On pinning UAX #11** — good catch, and both of your specifics were real. The comment now states it tracks **Unicode 17.0**, and:

- Added U+2630–U+2637 (trigrams) and U+268A–U+268F (monogram/digram). Unicode 16.0 moved these from Neutral to Wide and I had missed them.
- Narrowed Hangul Jamo Extended-A to **U+A960–U+A97C**. U+A97D–U+A97F are unassigned, and unassigned code points outside the CJK ranges UAX #11 enumerates default to Neutral, not Wide — so including the block's tail was wrong.

**Tests.** New case `textUnits measures a variation-selector sequence from the selector` covers `⭐️`, `✅️`, `☕️`, `⚡️`, `✈️`, `❤️`, the VS15 forms `⭐︎` / `✈︎`, a bare selector (0), and sequences in context (`✅️ Done` → 7, `⭐️⭐️` → 4). The wide-boundary case gained `☰` `☷` `⚊` `⚏`, plus `ꥼ` → 2 and `꥽` → 1 to pin the new Extended-A edge.

`node --test archify/test/geometry.test.mjs` → 54/54.

**Known limitation, not addressed here:** a ZWJ sequence such as `👨‍💻` still measures as the sum of its parts rather than one square. That predates this PR rather than being introduced by it, so I left it out to keep the diff reviewable — happy to take it as a follow-up if you want it.

NgoQuocViet2001:
Both points addressed — thanks for the careful review, the variation-selector one was a genuine regression I introduced and your repro numbers pinned it exactly.

**1. `archify.zip` rebuilt.** You're right that `utils.mjs` ships in the skill runtime, so the archive has to move with it. Rebuilt with `scripts/build-zip.sh` and verified with the same check `zip-freshness` runs:

```
unzip -q <fresh>.zip -d /tmp/fresh
unzip -q archify.zip  -d /tmp/checked
diff -rq /tmp/fresh/archify /tmp/checked/archify
```

Against the previously committed archive the only difference reported was `renderers/shared/utils.mjs` — so the archive is now current and nothing unrelated drifted into it. Against the rebuild it is clean.

One note in case it bites another Windows contributor: building on Windows silently produces a CRLF archive that fails this job even after a rebuild. I built from an LF export (`git -c core.autocrlf=false -c core.eol=lf archive`) and ran `build-zip.sh` under Linux. Happy to add a line to the contributing notes if you'd find that useful.

**2. Variation-selector sequences.** Fixed at the rule level rather than by special-casing the widened bases — details and the before/after table are in the [inline reply](https://github.com/tt-a1i/archify/pull/91#discussion_r3820054508). Short version: the pair is now measured from the selector, so VS16 → 2 and VS15 → 1, and `U+FE00–U+FE0F` never adds width on its own.

I also acted on the UAX #11 note in the same comment: the table now states it tracks Unicode 17.0, gained the two ranges you named (U+2630–U+2637, U+268A–U+268F), and Hangul Jamo Extended-A stops at U+A97C so the unassigned tail is no longer claimed as Wide.

**Verification**

- `node --test archify/test/geometry.test.mjs` → **54/54** (was 53; one new case).
- `archify/test/start-page.test.mjs` fails for me locally both with and without this change, i.e. on `upstream/main` too — it looks environmental on Windows and CI is green on it, so I left it alone rather than touch something outside this PR. Flagging it only so you know I saw it.

Ready for another look whenever you have time.

NgoQuocViet2001:
CI is green on `9354780` — all nine checks, including the `zip-freshness` job that reproduced the stale archive:

```
zip-freshness              success
test (18) (20) (22) (24)   success
package smoke (ubuntu) (macos) (windows)  success
webm-artifact              success
```

That closes both items from your review, so this is ready whenever you are.

tt-a1i (approved):
Thank you for the careful follow-up. The variation-selector handling, Unicode range corrections, regression coverage, and rebuilt archive address the earlier review comments cleanly. I re-ran the focused geometry suite (54/54) and the full test suite (657/657), confirmed the distributed utils.mjs matches archify.zip, and all nine CI checks are green. This is a focused and useful text-fitting fix. Approved — thanks for the contribution!

PR #107, opened 2026-08-25
fix(viewer): dim/restore arrowheads on focus and chapter hover preview

StvLi:
## Problem

SVG marker arrowheads are drawn **independently of the referencing path** — path `opacity` never dims or lights up the arrow itself. So in guided-view focus (`setMany`) and chapter hover preview (`showChapterPreview`), the line dimmed to `0.13`/`0.1` but **every arrowhead stayed fully lit**, making chain highlighting look broken (the focused chain's arrows never visibly "light up").

## Fix

- **`renderers/shared/utils.mjs`**: add four translucent **dim marker variants** (`arrowhead[-emphasis|-security|-dashed]-dim`, opacity `0.28`).
- **`assets/template.html`**:
  - `setMany` (guided-view focus / node selection): unmatched edges swap `marker-end` to the dim variant; matched edges keep/restore their authored arrowhead. `clear()` restores all.
  - `showChapterPreview` (chapter rail hover): in-chapter edges get `data-chapter-preview-role="edge-live"` + their authored arrowhead (CSS keeps line *and* arrow fully lit); all other edges dim. `clearChapterPreviewPresentation` restores markers and roles.

## Verification

- `node --check` on all inline scripts.
- Upstream `validate --quality showcase` on a real architecture spec: **9/9 checks, 0 errors / 0 warnings**.
- Runtime check via headless Chrome on the generated HTML: hovering chapter ② dims 7 non-chain arrowheads and keeps the 2 chain arrowheads fully lit (`LIVE=2 / DIM=7 / OTHER=0`).
- Test suites `chapter-delta-preview`, `architecture-delta`, `focus`, `story`: **22/22 pass**.

### DIFF
--- a/archify/assets/template.html
+++ b/archify/assets/template.html
@@ -4646,6 +4646,12 @@
       opacity: 0.28 !important;
       filter: grayscale(0.72) !important;
     }
+    /* The chapter's own edges stay fully lit (line + arrowhead) so the
+       hovered chain visibly lights up against the dimmed rest. */
+    svg[data-chapter-preview] [data-edge-from][data-chapter-preview-role="edge-live"] {
+      opacity: 1 !important;
+      filter: none !important;
+    }
     svg[data-preset="blueprint"][data-chapter-preview] [data-node-id] { filter: none !important; }
 
     /* Dashed boundary variants */
@@ -8409,7 +8415,11 @@
           node.removeAttribute('data-focus-selected');
           node.setAttribute('aria-pressed', 'false');
         });
-        edges().forEach(function (edge) { edge.removeAttribute('data-focus-match'); });
+        edges().forEach(function (edge) {
+          edge.removeAttribute('data-focus-match');
+          var originalMarker = edge.getAttribute('data-marker-original');
+          if (originalMarker) { edge.setAttribute('marker-end', originalMarker); edge.removeAttribute('data-marker-original'); }
+        });
         chip.hidden = true;
         label.textContent = '';
         detail.textContent = '';
@@ -8483,10 +8493,24 @@
         normalized.forEach(function (id) { selected[id] = true; related[id] = true; });
         var selectionMode = options.mode === 'selection' || normalized.length > 1;
 
+        function dimMarker(markerEnd) {
+          if (typeof markerEnd !== 'string') return markerEnd;
+          var m = markerEnd.match(/url\(#([^)]+)\)/);
+          if (!m || /dim$/.test(m[1])) return markerEnd;
+          return 'url(#' + m[1] + '-dim)';
+        }
         edges().forEach(function (edge) {
           var from = edge.getAttribute('data-edge-from');
           var to = edge.getAttribute('data-edge-to');
           var match = selectionMode ? selected[from] && selected[to] : selected[from] || selected[to];
+          // SVG markers do not inherit path opacity; swap the arrowhead to a
+          // translucent dim variant so focused chains visibly light up.
+          if (edge.getAttribute('data-marker-original') == null) {
+            edge.setAttribute('data-marker-original', edge.getAttribute('marker-end') || '');
+          }
+          edge.setAttribute('marker-end', match
+            ? (edge.getAttribute('data-marker-original') || '')
+            : dimMarker(edge.getAttribute('data-marker-original') || ''));
           if (!match) return;
           edge.setAttribute('data-focus-match', '');
           if (!selectionMode) { related[from] = true; related[to] = true; }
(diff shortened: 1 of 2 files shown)


YunyueLi:
Thanks for taking the time to investigate the focus and chapter-preview presentation. I reproduced this branch in a clean checkout and generated the real `examples/web-app.architecture.json` artifact. I am closing this version because the proposed fix is based on an unproven premise and introduces persistent viewer/export regressions.

## 1. The stated root cause does not reproduce in Chrome

The PR says that path `opacity` does not dim the referenced SVG marker. That is not what Chrome renders, and it is inconsistent with SVG object-opacity compositing: markers are painted as part of the shape's rendered result, while object opacity is applied to that result as a whole. See the SVG 2 rendering model:

- https://www.w3.org/TR/SVG/render.html#ObjectAndGroupOpacityProperties
- https://www.w3.org/TR/SVG/render.html#PaintingShapesAndText

I verified this with an isolated Chrome pixel test using a red path and a red `marker-end` over a white background. At a marker-only pixel outside the path stroke:

- path `opacity="1"` produced `[255, 0, 0, 255]`
- path `opacity="0.1"` produced `[255, 228, 228, 255]`

So the referencing path's opacity already dims the arrowhead in the browser used by the PR's own headless verification. Adding `opacity="0.28"` to a second marker then compounds the opacity instead of fixing an independently bright arrow:

- chapter preview: `0.10 × 0.28 = 0.028`
- focus: `0.13 × 0.28 = 0.0364`

That makes unmatched arrowheads substantially fainter than their lines.

The reported `LIVE=2 / DIM=7 / OTHER=0` check only counts the `marker-end` URLs after the mutation. It proves that the new code swapped marker references; it does not compare baseline pixels or demonstrate that the original browser rendering left arrowheads fully lit. A before/after visual or pixel assertion on the unmodified implementation is required before adding a compatibility workaround here.

## 2. Chapter preview cleanup leaves the preview's markers behind

On the generated Sample Web App artifact, I focused the second chapter button and then moved focus to the theme button to end the preview.

During preview:

- `data-chapter-preview="identity-and-cache"`
- 7 drawable edges referenced `*-dim` markers
- 2 drawable edges retained authored markers
- 15 edge-related DOM elements had `data-marker-original`

After preview ended:

- `data-chapter-preview` was removed
- all chapter preview roles were removed
- the same 7 drawable edges still referenced `*-dim` markers
- all 15 `data-marker-original` attributes remained

This follows directly from the patch: `clearChapterPreviewPresentation()` removes preview attributes and roles but never restores `marker-end` or releases the saved marker state. The PR description says that cleanup restores markers, but the implementation does not do that.

This is more than a missing one-line reset. A chapter preview may run while an active guided chapter already owns focus marker state. Leaving the preview must restore the previous focus distribution, not always restore all authored markers. The shared `data-marker-original` attribute does not model those two state owners, so focus and preview can overwrite one another.

## 3. Transient marker state leaks into canonical SVG exports

After the chapter preview was cleared, I invoked the real SVG export. The downloaded SVG contained:

- 7 `marker-end` references to `*-dim` definitions
- 15 `data-marker-original` attributes

Despite that leaked viewer state, the runtime reported:

```text
data-last-export-format="svg"
data-last-export-canonical="true"
```

The same issue occurs when exporting during ordinary focus. Focusing `api` and exporting produced 4 dim marker references and all 15 `data-marker-original` attributes while still reporting `canonical=true`.

`serializeSvg()` removes the existing focus/chapter attributes from its clone, but this patch does not restore authored `marker-end` values or remove `data-marker-original`. It also does not include the new state in the `canonicalStateClean` predicate. The exported diagram can therefore differ from the authored diagram while the receipt claims the export is canonical.

## 4. The selector mutates non-drawable edge metadata elements

`edges()` / `[data-edge-from][data-edge-to]` matches 15 elements in the sample, but only 9 are drawable paths with markers. The other 6 are edge-related label elements without `marker-end`.

The patch saves an empty original marker and writes `marker-end=""` to those elements. `Archify.focus.clear()` checks `if (originalMarker)`, so it never removes the empty saved value. After focus is cleared, all authored path marker URLs are visually restored, but 6 elements permanently retain both:

```html
data-marker-original=""
marker-end=""
```

Any marker-state implementation must be restricted to drawable shapes that actually have an authored marker.

## 5. Existing tests do not cover the behavior introduced here

This PR changes only `assets/template.html` and `renderers/shared/utils.mjs`; it adds no regression test. The existing 22 passing assertions do not exercise:

- baseline marker pixels before the workaround
- preview enter → leave restoration
- focus set → clear restoration
- active focus → chapter preview → leave restoration
- export during focus or after preview
- canonical cleanup of marker state
- edge metadata elements without markers

That is why the reported suites can pass while the real page remains mutated and the exported SVG is contaminated. The branch is also currently behind `main` and has no status checks, but the runtime and export failures above are independently blocking.

## What would be needed for a replacement PR

Please start by attaching a reproducible artifact and before/after visual or pixel evidence from a named browser showing that the current implementation actually leaves marker arrowheads bright. If this turns out to be a browser-specific compatibility issue, scope the workaround to the demonstrated behavior.

A replacement implementation would also need to:

1. operate only on drawable edges with an authored `marker-end`;
2. centralize marker-state calculation instead of duplicating URL rewriting in focus and chapter preview;
3. restore the previous owner state when preview ends (focused distribution when focus is active, authored distribution otherwise);
4. keep transient marker state out of canonical exports and extend `canonicalStateClean` defensively;
5. add browser/runtime tests for every enter/leave/clear/export sequence listed above and all four marker variants;
6. update onto current `main` and pass the full check suite.

Closing this revision rather than iterating on a workaround whose underlying defect is not currently demonstrated. A new PR is welcome if concrete cross-browser evidence establishes the original problem and the state/export lifecycle is covered end to end.

PR #114, opened 2026-08-26
fix(svg): emit XML-safe marker attributes

### LINKED ISSUE #110: deliver output renders as an XML error page in browsers (valueless data-* attributes in inline SVG)
# `deliver` output renders as an XML error page in browsers (valueless `data-*` attributes in inline SVG)

## Description

The HTML produced by `archify deliver` does not render. Browsers that parse the inline SVG strictly abort with an XML error and show the partial document instead of the diagram. Example from Firefox:

```
This page contains the following errors:
error on line 52 at column 33: Specification mandates value for attribute data-node-label
Below is a rendering of the page up to the first error.
```

The root cause is that the renderers emit SVG attributes **without values**. A valueless `data-foo` attribute is legal in HTML syntax but invalid in XML, and the inline SVG inside the delivered document fails any XML-strict parse.

## Reproduction (v2.15.0)

```bash
git clone https://github.com/tt-a1i/archify && cd archify
node archify/bin/archify.mjs deliver architecture archify/examples/brand-aware-delivery.architecture.json out.html --quality showcase --json
# open out.html in a browser, or verify programmatically:
python3 -c "import re, xml.etree.ElementTree as ET; s=open('out.html').read(); ET.fromstring(re.search(r'<svg\b.*?</svg>', s, re.S).group(0)); print('well-formed')"
# -> SyntaxError: not well-formed (invalid token)
```

No `npm install` is required.

## Affected locations

All valueless attributes in emitted SVG, five source files:

- `archify/renderers/architecture/render-architecture.mjs:1011` — `<text data-node-label${hasSub ? ' data-detail-anchor' : ''} …>`
- `archify/renderers/workflow/render-workflow.mjs:656` — same pattern
- `archify/renderers/sequence/render-sequence.mjs:321` — same pattern
- `archify/renderers/lifecycle/render-lifecycle.mjs:462` — same pattern
- `archify/renderers/shared/legend.mjs:199` — `rootAttributes = hasInteractiveEntries ? ' data-legend data-legend-bridge' : ' data-legend'`

`data-node-label` is emitted valueless on every node label `<text>` element across all five diagram types; `data-detail-anchor` additionally when the node has a sublabel; `data-legend`/`data-legend-bridge` on the legend root element when a legend is present. So every `deliver` output with at least one node is affected, and the bundled examples ship the same defect.

## Suggested fix

Give each attribute an explicit value, e.g. `data-node-label=""` / `data-detail-anchor=""` (or `="1"`), wherever the renderers emit them valueless. For comparison, `archify/renderers/shared/cli.mjs:198` already writes `data-node-label="${esc(label)}"` with a value, so the valueless emission is an inconsistency rather than an intentional pattern.

A regression guard would be a deliver-level check that the emitted document's inline SVG parses as well-formed XML (e.g. via `DOMParser` in the existing visual-check pipeline, or `xmllint --noout`), which would have caught this at the exact point of failure the reader hits.

## Environment

- archify v2.15.0 (installed via `npx skills add`)
- Node v22.23.2
- Verified in Firefox and via `xml.etree.ElementTree` (Python stdlib)

sunsunsun-java:
## Problem and value

Closes #110.

Raw inline SVG extracted from delivered HTML contains valueless `data-*` marker attributes. Those attributes are valid in HTML syntax but are not XML-well-formed, so strict XML consumers cannot parse the raw SVG source.

## Scope

- What changed:
  - Emit explicit empty values for `data-node-label`, `data-detail-anchor`, `data-boundary-label`, `data-legend`, and `data-legend-bridge`.
  - Cover Architecture, Workflow, Sequence, Dataflow, and Lifecycle through installed-skill `deliver` regression cases.
  - Parse raw SVG source with the dev-only `saxes` XML parser, without browser or HTML-parser normalization.
  - Verify tracked generated artifacts contain XML-well-formed inline SVG.
  - No schema, geometry, layout, Viewer, export implementation, or renderer behavior beyond explicit marker values.
  - The existing `fast-uri` lock version and its upstream audit finding remain unchanged; dependency remediation is outside #110.

## Stability impact

- Compatibility and migration risk: low. Empty-valued `data-*` attributes preserve HTML DOM presence-selector behavior while making raw SVG source valid XML.
- Renderer, validator, package, or generated-artifact risk: renderer output bytes and derived hashes change; topology, geometry, schemas, Viewer code, and validation contracts do not.
- Failure behavior and rollback path: no new runtime failure path. Reverting commit `7ac7766` restores the previous serialization.

## Tests run

- `node --test --test-name-pattern="cli: deliver works from an installed skill without node_modules|cli: deliver XML guard" test/cli.test.mjs` — 2 passed, 0 failed.
- `node --test test/generated-artifact-xml.test.mjs` — 1 passed, 0 failed; more than 20 tracked artifacts checked.
- `npm test` — 726 tests: 710 passed, 0 failed, 16 skipped because Chrome is opt-in for the default suite.
- `ARCHIFY_CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" node --test test/viewer-chrome-layout.test.mjs` — 11 passed, 0 failed, 0 skipped.
- Headless-Chrome Viewer SVG export probe across all five checked-in diagram types — 5/5 exported roots were `svg`, each with 0 XML parser errors.
- `npm ci` — clean install succeeded. It reports the pre-existing upstream `fast-uri` audit finding; this PR leaves the lock version unchanged.

## Visual evidence

Visual review: passed with real Chrome. This is a serialization-only change; no pixel, geometry, layout, or Viewer source changed.

## Generated artifacts

Regenerated only outputs whose authoritative renderer source changed:

- `archify/examples/*.html` for all five diagram types.
- Root `examples/*.html` renderer outputs and the affected Checkout receipt.
- `docs/gallery.html`, `docs/gallery/manifest.json`, and `docs/gallery/artifacts/*.html`.
- `docs/cases/mco-runtime.architecture.html`.
- `experiments/mco-showcase/mco-runtime.html`.
- `docs/assets/archify-live-proof.json` artifact hashes only; the GIF was not regenerated.
- `archify.zip`, rebuilt because packaged renderer and example bytes changed.

## Checklist


tt-a1i (approved):
Reviewed the focused XML-serialization change and regression coverage across all five diagram types. The explicit empty values preserve existing DOM presence semantics while making raw inline SVG XML-well-formed. Targeted XML tests and all required CI checks pass.

nalzfyry176-glitch:
> Reviewed the focused XML-serialization change and regression coverage across all five diagram types. The explicit empty values preserve existing DOM presence semantics while making raw inline SVG XML-well-formed. Targeted XML tests and all required CI checks pass.

PR #122, opened 2026-08-26
fix(brand-capture): read a bounded head instead of capping the whole page

korjavin:
## Problem and value

Fixes #102. `brands capture` reads the HTML page to find declared icon `<link>` tags, but `readLimited(page.response, MAX_HTML_BYTES)` counted the *entire* response body against the 256 KiB cap. A page with a tiny `<head>` (a handful of icon links) failed with `brand asset is too large` whenever its total body — content the icon scan never even looks at — pushed the response past the cap, before a single icon candidate was inspected.

Confirmed against the issue's own repro:

```
$ node bin/archify.mjs brands capture https://xquik.com --json
```

- On `main`: `brand capture failed: brand asset is too large` (exit 2).
- After this change: succeeds, returns a digest-pinned PNG.

## Scope

- What changed: `renderers/shared/brand-marks.mjs` — the HTML-page read path now uses a new `readHtmlHead()` instead of `readLimited()`. It streams the response and stops as soon as `</head>` appears in what's been read so far, before ever counting or downloading the rest of the body.
- What deliberately did not change: `readLimited()` itself (still used unchanged to guard image bytes via `imageData()`), the icon-candidate scan (`iconCandidates()`), the fetch/redirect/DNS-pinning/deadline machinery, and the failure mode when the byte budget runs out *before* a head close is seen — that still throws the same `brand asset is too large` error and exit code, matching the issue's explicit "keep the existing failure when no closing head appears within the byte limit."
- No unrelated changes.

## Stability impact

- Compatibility and migration risk: none — this only affects how much of a page is read before parsing icon candidates; the parsed contract (`iconCandidates`, capture result shape) is untouched.
- Renderer, validator, package, or generated-artifact risk: none — brand capture is a CLI/network path, not part of the typed JSON renderer contract.
- Failure behavior and rollback path: unchanged failure path and error message for the cases that should still fail (head itself exceeds the cap, or is missing). Revert is a single-file change if needed.

## Tests run

```
cd archify
npm ci
npm test
```
729 passed, 0 failed, 16 skipped (real-Chrome `visual-check` tests, gated behind `ARCHIFY_CHROME`, unset in this environment — pre-existing, unrelated to this change).

Also ran the exact repro from the issue directly against the live site (see above), and confirmed via `git stash` that it fails on unpatched `main` and succeeds with the fix applied.

## Visual evidence

Not applicable — this is a network/parsing fix, no rendered-diagram or Viewer output changes.

## Generated artifacts

None regenerated; no generated-artifact sources changed.

## Checklist

- [x] I added or updated a regression test for behavioral changes — three new CLI-level tests in `test/brand-marks.test.mjs`: small head + large body captures; `</head>` split across chunks; head-itself-too-large still fails closed. Verified all three fail on unpatched `main` before writing the fix, per CONTRIBUTING's evidence requirement.
- [x] I removed secrets, private repository content, and customer data from fixtures and screenshots — test fixtures use only a local `http.createServer` and a 1x1 PNG fixture already used elsewhere in the test file; the live repro against xquik.com uses only its public HTTP response, no credentials.

## Maintainer integration and boundary repair

Candidate `a8a6f24` includes dev `b98a414`. During integration, two public-CLI regressions reproduced on the proposed reader: an oversized head with its closing tag in the limit-crossing chunk was incorrectly accepted, and an icon after the closing head in the same chunk was used. The reader now stores at most256KiB, scans each byte once, recognizes a complete case-insensitive closing tag across chunks, and returns only bytes through that tag. An exact-budget head with a large body remains accepted. The existing image limits, network checks, and digest pinning remain.

Official Node22.23.2/macOS, final combined source: `node --test archify/test/brand-marks.test.mjs`32 passed,0 failures/skips; new regression run before correction:2 failed/1passed; canonical ZIP rebuilt; `git diff --check` passes. Full remote CI is pending. Integration targets dev, not main.


tt-a1i (changes requested):
Thanks for fixing the bounded-head behavior; the focused brand-mark suite passes on this head (28/28). The blocker is the distributable: `archify/renderers/shared/brand-marks.mjs` differs from the copy inside the committed `archify.zip`, and this PR does not update the archive. Installed Skill users would therefore retain the old whole-page cap and miss the fix. Please rebuild `archify.zip` with the canonical Node 22 build, verify exact freshness, and rerun the current package-smoke/zip-freshness gates after syncing with main.

korjavin:
Good catch — thanks. `archify.zip` is rebuilt now:

- Rebased onto current `main` first (picked up the zip-determinism work from the recent `build(zip)` commits).
- Rebuilt with the canonical Node 22 toolchain via `scripts/build-zip.sh` (confirmed `node -p 'process.versions.node'` → 22.23.2 here).
- Verified byte-exact freshness the same way the `zip-freshness` CI job does: `scripts/build-zip.sh /tmp/fresh.zip && cmp -s /tmp/fresh.zip archify.zip` — exact match.
- `npm run check:brand-marks` / `check:validators` — clean, so the archive wasn't also stale from unrelated generated-artifact drift.
- `npm test` in `archify/` — 841 passed, 0 failed, 27 skipped (browser-gated, unchanged).
- Replicated the `package-smoke` job directly: unzipped the rebuilt `archify.zip` and ran `scripts/package-smoke.mjs` against the extracted package — passed.

The rebuilt archive now carries `readHtmlHead()` (confirmed by grepping the extracted package's `renderers/shared/brand-marks.mjs`), so an installed Skill picks up the fix.

tt-a1i (changes requested):
Thanks for rebuilding the archive. Reviewed current head 09ed28d675bf857def411bef8db071ded4177368: all 73 packaged files now match source, so the previous stale-ZIP request is resolved. Two runtime cases remain in readHtmlHead (archify/renderers/shared/brand-marks.mjs:284-294).

1. [P2] Enforce the byte limit against the actual head-close offset, not after an includes check. A localhost page whose head is 286806 bytes (budget262144) is captured successfully when the closing marker shares the last chunk. The same oversized content can instead fail when split differently. Locate the actual closing marker and reject if its end lies beyond the retained budget; test both coalesced and split chunks.
2. [P2] Do not interpret a script string as the closing head tag. A small valid head sends `<script>const template = "</head>";</script>` followed by its real icon link. The reader cancels before the link arrives, falls back to favicon.ico and fails404 when that fallback is absent. Use a bounded HTML-aware scan that handles raw-text/comment contexts.

Both reproduced with real local HTTP and the public capture command. The existing 28 brand-marks tests pass; those tests do not cover these cases. No external-site write, source edit, or merge.

PR #125, opened 2026-08-26
Sandbox CLI output paths (Fixes #124)

### LINKED ISSUE #124: [Bug]: CLI deliver output path can overwrite non-HTML files (meta.output sandboxed; CLI is not)
### Archify version or commit

`8b542d60e5eb2c552a3bb5bc1651d076be312aec` (`archify/package.json` reports `2.16.0-dev.0`)

### Installation method

Git clone

### Diagram type

Workflow (shared `resolveOutputPath` — same behavior for other `deliver` types)

### Exact command

```bash
cd archify
printf 'MARKER=do-not-destroy' > ../marker.env
node bin/archify.mjs deliver workflow examples/agent-tool-call.workflow.json ../marker.env --json
```

Absolute output paths behave the same way.

No crafted IR is required. Unchanged checked-in input:

`examples/agent-tool-call.workflow.json`

The failure is in the **CLI output argument**, not authored diagram content. (`meta.output` is sandboxed and is not the path that fails.)

`deliver --json` returned success (`"ok": true`) with `"output"` pointing at `../marker.env`.

After the command:

- previous contents of `marker.env` were gone
- file size became ~728304 bytes
- file began with `<!DOCTYPE html>`

So delivery committed an HTML artifact onto a non-`.html` path outside the intended “diagram output” contract.

### Expected behavior

CLI `[output]` should follow the same safety contract as `meta.output` (documented/enforced in `resolveOutputPath`):

1. require a resolved `.html` extension, and/or
2. require the path to stay inside the current working directory after symlink resolution, and/or
3. refuse to overwrite an existing non-`.html` file without an explicit force flag

A failed delivery should return a non-zero exit and a stable `diagnostics[]` code (Agent-readable), not a successful receipt.

### Actual behavior

In `archify/renderers/shared/output-path.mjs`, `source === 'meta'` gets absolute / `.html` / cwd checks. `source === 'cli'` only does `path.resolve(cwd, rawOutput)` and then `deliver` commits the artifact there.

That lets a CLI output argument replace arbitrary existing files (including non-HTML marker files outside cwd).

### Optional final-artifact screenshot

N/A — not a visual renderer defect. Happy to attach a redacted `--json` receipt if useful.

### Environment

- OS: Windows 10/11
- Node.js: 18+ (ran via `node bin/archify.mjs`)
- Client: Cursor (evaluating Archify as an Agent skill / CLI)
- No access tokens, credentials, private repository contents, or personal/customer data included

### Notes for maintainers

CONTRIBUTING.md says security issues should use GitHub private vulnerability reporting. That endpoint returned “private vulnerability reporting is not enabled” for this repository, so I am filing through the bug form instead as a **CLI contract / unsafe overwrite** issue (Agent-first blast radius if an Agent is steered into a bad output path).

Suggested fix: apply the `meta.output` constraints to CLI `requestedOutput`, plus a regression that asserts `deliver … ../marker.env` fails with a stable diagnostic.

I can retest a patch quickly.

alivirgo:
## Summary
- Apply the same cwd + extension sandbox used for `meta.output` to CLI output paths, so `deliver`/`render`/`compare` cannot overwrite arbitrary files (e.g. `../marker.env`).
- Keep compare receipts as `.json` via `requiredExtension: '.json'` while HTML artifacts still require `.html`.
- Add regression tests for CLI escape, non-`.html` deliver targets, and non-`.json` receipts.

Fixes #124

## Test plan
- [x] `node --test test/output-path.test.mjs` (non-symlink cases; symlink tests need Windows Developer Mode / elevated privileges)
- [ ] `archify deliver workflow <input> ../outside.env` fails and leaves the outside file untouched
- [ ] `archify deliver workflow <input> ../escaped.html` fails with `output/cli-outside-cwd`
- [ ] `archify compare ... --receipt foo.html` fails with `output/cli-extension`
- [ ] Relative in-cwd `.html` / `.receipt.json` outputs still succeed

Made with [Cursor](https://cursor.com)

### DIFF
--- a/archify/bin/archify.mjs
+++ b/archify/bin/archify.mjs
@@ -472,6 +472,7 @@ async function commandCompare(args) {
       defaultOutput: compareReceiptPath(outputPath),
       inputPaths: [basePath, headPath],
       otherOutputPaths: [outputPath],
+      requiredExtension: '.json',
     }));
   } catch (error) {
     const outputDiagnostic = error.archifyDiagnostics?.[0];
@@ -645,6 +646,7 @@ async function commandCompare(args) {
         defaultOutput: compareReceiptPath(currentOutput),
         inputPaths: [basePath, headPath],
         otherOutputPaths: [currentOutput],
+        requiredExtension: '.json',
       });
     } catch (error) {
       const outputDiagnostic = error.archifyDiagnostics?.[0];
--- a/archify/renderers/shared/output-path.mjs
+++ b/archify/renderers/shared/output-path.mjs
@@ -247,6 +247,11 @@ export class OutputPathError extends Error {
   }
 }
 
+function normalizeRequiredExtension(extension) {
+  const value = String(extension || '.html').trim().toLowerCase();
+  return value.startsWith('.') ? value : `.${value}`;
+}
+
 export function resolveOutputPath({
   requestedOutput,
   authoredOutput,
@@ -255,9 +260,20 @@ export function resolveOutputPath({
   inputDescription = 'an input',
   otherOutputPaths = [],
   cwd = process.cwd(),
+  // Artifact outputs default to .html. Compare receipts pass '.json'.
+  requiredExtension = '.html',
 }) {
   const rawOutput = requestedOutput || authoredOutput || defaultOutput;
   const source = requestedOutput ? 'cli' : (authoredOutput ? 'meta' : 'default');
+  const extension = source === 'meta' ? '.html' : normalizeRequiredExtension(requiredExtension);
+  const label = source === 'meta' ? 'meta.output' : 'CLI output';
+  const relativeFix = source === 'meta'
+    ? `set meta.output to a relative ${extension} path inside the current working directory`
+    : `pass a relative ${extension} path inside the current working directory`;
+  const sandboxSource = source === 'meta' || source === 'cli';
+
+  // meta.output must stay authored-relative. CLI absolute paths are still
+  // resolved, then rejected by the shared cwd containment check below.
   if (
     source === 'meta'
     && (path.isAbsolute(rawOutput) || path.posix.isAbsolute(rawOutput) || path.win32.isAbsolute(rawOutput))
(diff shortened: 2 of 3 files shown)


jordandunmire97-ai (approved):
Expanding

chiang21fcb (approved):
Thanks for the PR @alivirgo. This is a clean, well-scoped fix.

## What I reviewed

- `output-path.mjs`: The `sandboxSource` unification is elegant. All three safety gates (extension, resolved-extension, cwd containment) now apply to both `meta` and `cli` paths. The `requiredExtension` parameter keeps the contract flexible for compare receipts.
- `archify.mjs`: The two `requiredExtension: '.json'` additions for receipt paths are exactly right.
- `output-path.test.mjs`: Three new regression tests cover the exact scenarios from #124. The two updated existing tests correctly adapt to the new check order (extension gate runs before alias gate).

## Confirmed

- The `../marker.env` overwrite from #124 is blocked by the extension check
- Compare receipts are gated by `.json` extension independently
- Stable diagnostic codes follow the existing `output/meta-*` pattern
- All existing tests are updated to match the new behavior

## One note

The cwd containment for CLI paths is a reasonable safety choice. If maintainers want to allow absolute CLI paths for power users, that can be relaxed later without changing the extension gate, which is the critical fix.

No blockers. This is ready for maintainer review.

tt-a1i (changes requested):
Thanks for the focused sandboxing fix. The source-level direction is sound, but the distributable is stale on this exact head. I compared the committed files with the corresponding entries in `archify.zip`: both `archify/renderers/shared/output-path.mjs` and `archify/bin/archify.mjs` differ from the packaged copies. As a result, users installing the release artifact would not receive the output-path protections this PR adds, even though the repository source has them. Please rebuild `archify.zip` with the repository's canonical Node 22 build, verify byte-for-byte freshness, and run the current package-smoke/zip-freshness gates after syncing with main.

jordandunmire97-ai (approved):

tt-a1i (changes requested):
Additional review of unchanged head e9653bae8f7456ad58ac0742d7d752858594cb60: the prior stale-ZIP finding still applies, but this is a newly reproduced public CLI regression, not a duplicate package reminder.

[P1] Separate trusted internal renderer targets from user-supplied output paths (archify/renderers/shared/output-path.mjs:268-273,288-315). From archify/, `node bin/archify.mjs validate architecture examples/web-app.architecture.json --json` now fails output/cli-outside-cwd because commandValidate renders into os.tmpdir while retaining the project cwd. `node bin/archify.mjs inspect architecture examples/web-app.architecture.json` fails output/cli-extension because its internal renderer target is /dev/null. Parent-versus-head resolver comparison confirms the regression. This blocks ordinary validation/inspection of valid diagrams, not an attempted unsafe user write.

Preserve the external sandbox while giving internal temporary/no-output calls an explicit safe seam, and add successful public validate/inspect regressions. Focused output-path tests: 17 passed, 5 failed; the latter also include fixtures not updated for the extension restriction, so they are not presented as proof of remaining overwrite vulnerabilities. No source edits or merge.

tt-a1i:
Closing during maintainer backlog triage. The file-type protection is already delivered by #322 (ed7f4d4): HTML artifacts and JSON receipts are checked, including resolved symlink targets. Main deliberately preserves explicit absolute and parent-directory output paths, whereas this proposal would reject those supported uses. Verified the current-main output-path and CLI output-type suites: 32 passed, 0 failed, 0 skipped. Thank you for raising and working on this boundary; the applicable protection is already shipped.

PR #129, opened 2026-08-27
build(zip): make canonical archive reproducible

tt-a1i:
## Summary

- preserve the reproducible and atomic canonical ZIP work from #119
- retry the pinned DSH runtime install once, only for transient npm network failures
- recreate the install directory before every attempt and record the attempt count in the acceptance receipt
- keep all non-network failures fail-fast and retain the real install, skill-discovery, package-smoke, and uninstall checks

## Verification

- `node --test integrations/deepseek-harness/test/*.test.mjs` on Node 22.23.0: 23/23 passed
- `git diff --check`
- `node --check` for the modified scripts
- real local distribution acceptance exercised the retry path and produced a final receipt with `attempts: 2`; both external npm attempts timed out in the current network environment, so GitHub CI remains the independent full acceptance gate

This repository-local continuation preserves the original commits and authorship from #119 because that fork does not allow maintainer edits.

Supersedes #119.


tt-a1i:
@YunyueLi Could you please review this repository-local continuation of #119?

The original reproducible/atomic ZIP commits and authorship are preserved. The additional change replaces the DSH acceptance runtime's repeatedly timing-out `npm install` with a real pnpm install, keeps a bounded retry for transient network errors, and explicitly allowlists the five required dependency build scripts.

Local Node 22 contract tests pass 23/23, the complete local DSH distribution receipt is `ok: true`, and all GitHub checks are green. The GitHub DSH distribution acceptance improved from two 300-second timeouts to a successful 28-second full install/discovery/smoke/uninstall run.

Auto-merge is enabled after the required independent approval.

alivirgo:
Post-fix verification on Windows (Node 24.18.0):

- H1 atomic replace: deterministic writer, no pre-delete of output
- H2 unmerged index: ls-files --stage guard present
- H4 ZIP32: requireZip32 + temp rename in write-deterministic-zip.mjs
- H5 fail-closed: rejected build preserves trusted archive bytes (Node 22 gate on Node 24)


elease-package-gates.test.mjs passes locally (5 pass / 4 skipped awaiting Node 22).

Optional follow-up in alivirgo/fix/zip-reproducibility-116: Windows test runner uses Git Bash for build-zip.sh invocations so the Node-version rejection test works on win32 too.

PR #130, opened 2026-08-27
Improve workflow short-edge column diagnostics (Fixes #126)

### LINKED ISSUE #126: workflow: edge length between adjacent columns is non-uniform, so identical specs pass or fail depending on which column pair you picked
Authoring ~14 diagrams across three repositories with v2.16 (native release archive, Node 24). The layout and composition validators were consistently precise and genuinely useful — every diagnostic named a subject, evidence and supported fixes. This one is the exception: it is not predictable from the spec, and I could not find it in `SKILL.md` or `references/authoring-contract.md`.

## What happens

In a `workflow`, an edge between two nodes in the same lane and adjacent columns is rejected as too short — but only for some column pairs. Column spacing is not uniform, and the gap is not derivable from the column index.

Two nodes, one labelled edge. Only `col` changes between runs:

| columns | result |
|---|---|
| 0 → 1 | ok |
| 1 → 2 | `too short (12px; minimum 28px)` |
| 2 → 3 | ok |
| 3 → 4 | `too short (22px; minimum 28px)` |
| 4 → 5 | ok |

Not caused by `phases` or `groups` — the table is identical with both removed. Not affected by `meta.viewBox` width: 900, 1080, 1400 and 1600 produce the same coordinates in the diagnostics.

## Repro

Minimal spec, vary `col` on the two nodes:

```json
{
  "schema_version": 1,
  "diagram_type": "workflow",
  "meta": { "title": "repro", "quality_profile": "showcase", "viewBox": [900, 520] },
  "lanes": [{ "id": "agente", "label": "Agente" }],
  "nodes": [
    { "id": "a", "lane": "agente", "col": 1, "type": "backend", "label": "A", "sublabel": "x" },
    { "id": "b", "lane": "agente", "col": 2, "type": "backend", "label": "B", "sublabel": "y" }
  ],
  "edges": [{ "id": "ab", "from": "a", "to": "b", "label": "liga", "variant": "default" }]
}
```

```
node bin/archify.mjs validate workflow repro.json --quality showcase
```

## Why it costs authoring time

The suggested fix — *"drop its label or route it through a channel"* — asks the author to remove semantic information for a geometric reason, when the same relationship one column over is fine. Since the spacing is not documented, each legal pair is found by trial: I spent four validate rounds on a single six-step pipeline relocating steps between lanes.

## What would help

In order of preference:

1. Uniform column spacing, or spacing that always clears the 28px edge minimum.
2. Failing that, document the spacing model so authors can place nodes correctly the first time.
3. A diagnostic that names a legal alternative (*"col 2→3 would satisfy the minimum"*) instead of suggesting the label be dropped.

## Three smaller documentation gaps

From the same pass, if useful:

- `--repo-root` and `--layout-json` are `architecture`-only (the CLI says so on use). `SKILL.md` describes repository evidence without noting that the other four types cannot carry `meta.repository` or `sources` — that decides whether a multi-diagram set can be grounded at all.
- `workflow` accepts `col` 0..5 only; the limit surfaces as a schema error at validate time.
- In `lifecycle`, the event lane's column *N* aligns under the main rail's column *N+2*. `SKILL.md` says "align beneath later phases"; the exact offset is the part an author needs.

One more trap worth a line in the delivery contract: when `deliver` fails it preserves the previous artifact, so running `visual-check` immediately after measures the **old** file.

alivirgo:
## Summary
- Short same-lane edges no longer suggest dropping the label. The diagnostic names legal adjacent column pairs (`0→1`, `2→3`, `4→5`) for the node widths in play, or channel/width/skip-column repairs.
- Documents that workflow `colXs` are non-uniform (pairs `1↔2` / `3↔4` are too tight for default 92px nodes) in `SKILL.md` and `renderers/workflow/README.md`.
- Adds regression tests for the 12px / 22px tight pairs vs a wide pair that still validates.

This does not rebalance column spacing (that would break showcase layouts). It makes the existing constraint diagnosable and documented.

## Test plan
- [x] `node --test archify/test/workflow-column-spacing.test.mjs`
- [ ] Showcase workflow examples still validate as before
- [ ] Confirm short-edge message contains wider-pair hint and not "drop its label"

Fixes #126

### DIFF
--- a/archify/SKILL.md
+++ b/archify/SKILL.md
@@ -38,6 +38,8 @@ Do not read `renderers/shared/geometry.mjs`, renderer source, validator source, 
 
 Lifecycle note: phase columns `0..4` occupy the main rail; event/outcome columns `0..2` align beneath later phases. A recoverable state uses `type: "failure"` plus a real transition back to the active state.
 
+Workflow note: column centers are not uniform (`col` 0..5 at x 88 / 220 / 300 / 430 / 500 / 625). Same-lane adjacent pairs 1↔2 and 3↔4 are too tight for default 92px nodes: a straight edge there fails the 28px minimum even when an identical edge one column over passes. Prefer 0→1, 2→3, or 4→5; otherwise skip a column, reduce `width`, or route through a channel. Details: `renderers/workflow/README.md`.
+
 ## Type router
 
 | Type | Use for |
--- a/archify/renderers/workflow/README.md
+++ b/archify/renderers/workflow/README.md
@@ -81,8 +81,10 @@ backed by rendered nodes receive Semantic Legend controls.
 | Legend row | y = lane bottom + 44; viewBox height must be ≥ legend y + 18 |
 
 Column-center gaps are 132 / 80 / 130 / 70 / 125 px: columns 1↔2 (80px) and
-3↔4 (70px) cannot both hold default-width 92px nodes in the same lane — skip a
-column or reduce `width`.
+3↔4 (70px) cannot both hold default-width 92px nodes in the same lane with a
+straight labeled edge that clears the 28px minimum. Use 0→1, 2→3, or 4→5,
+skip a column, reduce `width`, or route through a channel. The short-edge
+diagnostic names those legal adjacent pairs for the node widths in play.
 
 ## Design Rules
 
--- a/archify/renderers/workflow/render-workflow.mjs
+++ b/archify/renderers/workflow/render-workflow.mjs
@@ -51,10 +51,69 @@ const layout = {
   laneH: 104,
   laneGap: 20,
   laneTitleH: 30,
+  // Non-uniform by design (phase/group rhythm). Adjacent gaps are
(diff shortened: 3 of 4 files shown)


sunsunsun-java:
Thanks for investigating this. The root-cause analysis is correct: the legacy workflow grid has non-uniform colXs, which makes some adjacent pairs fail the direct-clearance rule.
However, this scoped fix only documents the fixed-column workaround and improves the diagnostic message. It does not address the agreed design for issue #126: schema v2 must treat col as a logical rank and derive pixel positions through constraint-driven layout, while keeping v1 behavior and migration explicit.
The current implementation is covered by [PR #146](https://github.com/tt-a1i/archify/pull/146), which adds the readable-v2 compiler, causal v1 diagnostics, migration support, label/viewBox containment, and the complete adjacent-rank acceptance matrix. It also avoids suggesting a channel as a fix for node overlap.
Because the implementation touches the same renderer and documentation areas, merging both approaches would create conflicting geometry semantics. Please consider closing this PR as superseded. The regression-test ideas may still be useful as v1 compatibility coverage and can be adapted into the newer compiler-based implementation.

PR #136, opened 2026-08-27
fix(dsh): make Archify bundle source-auditable

AI-Scarlett:
## What this fixes

This makes the DeepSeek Harness integration reviewable from a fixed Git Commit without running an Archify build.

- Keeps the bounded static Skill payload in the integration package.
- Replaces the patch entry for an official package with an adapter owned by the Archify DSH package; the adapter delegates only to the public DSH filesystem Skill provider.
- Declares that provider as a peer dependency and retains one isolated archify-plugin provider.
- Adds static payload, package, tarball, and distribution-contract coverage.

The embedded payload contains the runtime files and JSON examples needed by the Skill instructions. It intentionally excludes test code, lifecycle scripts, development dependencies, generated tooling, and large rendered HTML examples. The package version is intentionally unchanged.

## Compatibility boundary

This PR does not claim current DSH compatibility. The existing evidence remains DSH 0.1.0-rc.6; later DSH releases need the same isolated acceptance run before being declared supported.

## Local verification

- node --test integrations/deepseek-harness/test/package-contract.test.mjs integrations/deepseek-harness/test/bundled-payload-contract.test.mjs
- git diff --check

I did not run real DSH installation or runtime acceptance locally. This PR's isolated CI remains the gate for that behavior.


tt-a1i (changes requested):
Thanks for making the bundle inspectable. Reviewed exact head 88dc7aa32febc60ae216455b605f513f12628403. The static projection itself matches all65 canonical source files (2032116bytes), and the70 ZIP payload entries also match. Two integration/acceptance defects remain.

## Standards
[P2] Wire the new eager peer import into the clean-checkout test environment (integrations/deepseek-harness/lib/index.js:3). The workflow contract job directly runs `node --test integrations/deepseek-harness/test/*.test.mjs` without installing integration peers. That exact command now yields25passed/1failed: adapter-security.test.mjs cannot import @deepseek-ai/dsh-skill-filesystem. Distribution acceptance depends on this job. Install the declared peer reproducibly for the test or expose a dependency-free resolver seam without weakening real adapter coverage.

## Spec
[P2] Actually compare the extracted ZIP on Windows (scripts/distribution-acceptance.mjs:440-446 under the integration). The branch extracts committed.zip but compares packedSkill only to checkedInPayload, never checkedDir/archify, then reports unzipContentsIdentical:true. Executing that exact branch with an intentionally stale valid ZIP still reports equality. Keep a normalized comparison with the extracted ZIP, or label it skipped; do not attest to bytes that were not compared.

Package/payload/tarball/discovery focused tests13/13 passed. Windows finding uses an isolated execution of the exact branch on macOS, not a claim of native Windows or live DSH acceptance. No source edits/merge.

AI-Scarlett:
I have pushed `1c05770` addressing both review findings:

1. The resolver now lives in a dependency-free module re-exported from the adapter. Clean-checkout tests import that seam; an additional isolated peer fixture imports the real adapter and checks its Config, skills injection and exact delegation.
2. Windows no longer claims ZIP content equality after comparing only packed and checked-in payloads. The receipt explicitly reports `not-asserted-on-windows` and `extraction-only-equality-skipped`; Linux retains the actual fresh-vs-committed checks.

Validation: `node --test integrations/deepseek-harness/test/*.test.mjs` passes in a checkout without installing integration peers. No native Windows or DSH distribution runtime pass is claimed.

I also checked current main. It now deliberately rebuilds the immutable `archify-dsh-v0.1.0` snapshot, while this older proposal changes that distribution to a checked-in source payload. That creates packaging conflicts; I preserved main's release policy instead of resolving them by silently replacing the released payload. The two requested fixes are ready for review here, but the source-distribution proposal still needs your decision on a separately versioned adapter release.

tt-a1i:
Closing this older source-payload distribution proposal after checking the current dev implementation and your revised-head notes. The source-auditability goal is now covered by the existing owned adapter plus `release.json` (immutable source commit, Skill version, DSH version), Git-HEAD adapter inputs, canonical clean staging, and tarball distribution acceptance. The prepared adapter remains separately versioned; no published release is being replaced. The latest dev DSH run35114578621 passed both package contracts and real distribution acceptance.

Checking a second full Skill payload into `.dsh-bundled-skills/archify` would add another runtime/Viewer copy to maintain and conflict with the current one-product-source packaging contract. We are retaining the pinned-source build approach rather than introducing that alternative distribution format.

Thank you for fixing the dependency-free resolver seam and accurately bounding the Windows claims. This closure does not claim that your proposal failed those checks, nor that pending v0.2.0 has been published. A future need for build-free source distribution should be proposed as a separately versioned adapter change with a concrete consumer requirement.

PR #132, opened 2026-08-27
test(zip): invoke build-zip.sh via bash on Windows

alivirgo:
## Summary
#116 was closed by [#129](https://github.com/tt-a1i/archify/pull/129). This PR is a small follow-up only.

On Windows, `release-package-gates.test.mjs` now invokes `build-zip.sh` through Git Bash so the Node-version rejection test captures stderr correctly (previously failed with `undefined` stderr when spawning the `.sh` directly on win32).

## Verification (Windows, Node 24.18.0)
- `node --test archify/test/release-package-gates.test.mjs`: 5 pass, 0 fail, 4 skipped

## Test plan
- [x] `node --test archify/test/release-package-gates.test.mjs` on Node 24
- [ ] CI `package smoke (windows-latest)` job

PR #145, opened 2026-08-27
fix(repo): check out tracked text files as LF

### LINKED ISSUE #144: [Bug]: Tracked files check out as CRLF on Windows, so working-tree bytes stop matching the index
### Archify version or commit

`2.16.0-dev.0` at `9a5060566c832832fb843e457e58c8ee6bac82fd`

### Installation method

Git clone

### Diagram type

Installation or packaging

### Exact command

```shell
git clone https://github.com/tt-a1i/archify.git
cd archify/archify
npm ci
npm test
```

### Minimal reproduction

No typed JSON is involved — the input is the checkout itself. On any machine where `core.autocrlf=true` (the default for Git for Windows), a fresh clone materializes tracked text files with CRLF while the index stores LF.

```console
$ git config core.autocrlf
true

$ git show HEAD:examples/checkout-platform-delta.html | head -c 16 | od -c
0000000   <   !   d   o   c   t   y   p   e       h   t   m   l   >  \n

$ head -c 16 examples/checkout-platform-delta.html | od -c
0000000   <   !   d   o   c   t   y   p   e       h   t   m   l   >  \r

$ git status --porcelain examples/checkout-platform-delta.html
                                   # clean — the divergence is invisible

$ git ls-files --eol | grep -c "w/crlf"
373
```

`.gitattributes` marks generated artifacts `linguist-generated=true` but sets no `text`/`eol` attribute, so nothing pins the working-tree line endings. 373 tracked files are affected: 140 `.mjs`, 99 `.md`, 66 `.json`, 46 `.html`, 7 `.yml`, 6 `.mmd`, and `scripts/build-zip.sh`.

### Expected behavior

A tracked file's working-tree bytes match its committed bytes, so tests that compare renderer output against a checked-in artifact are meaningful and `build-zip.sh` stages the committed bytes.

### Actual behavior

`npm test` on a fresh Windows clone:

```
# tests 735
# pass 699
# fail 15
# skipped 21
```

Eleven of those fifteen are byte-comparison failures caused purely by the line endings, for example:

```
not ok 34 - checked-in Checkout compare artifact is reproducible from its authoritative inputs
    + '<!doctype html>\n' +
    - '<!doctype html>\r\n' +
```

The full set that clears once line endings are pinned:

- README motion proof is compact, looping, and backed by current gallery artifacts
- checked-in Checkout compare artifact is reproducible from its authoritative inputs
- contributor and pull-request guides keep proof changes reproducible and stability-first
- explicit en and zh-CN preserve complete authored field inventories across all five modes
- generated proof gallery matches its sources, receipts, and checked-in artifacts
- guide page: ships bilingual recipes and syntactically valid interaction code
- narrow viewport preserves a contained fallback without adding a mobile product surface
- omitted locale preserves non-English authored content and the English Viewer contract in all five modes
- skill description is portable across 1024-character runtimes and remains searchable
- start page: offers five bounded bilingual starts without ingesting source content
- zh-CN localizes renderer-owned output across all five modes without translating authored content

### Impact beyond the test suite

This also touches the reproducibility guarantees hardened in #119 / #129. `scripts/build-zip.sh` stages tracked repository content, so on a CRLF checkout every staged text file carries different bytes and `archify.zip` gets a second byte representation of the same logical contents — the outcome the Node 22 pin in the zip builder exists to prevent. `build-zip.sh` is itself checked out with CRLF.

`windows-latest` is in the `package-smoke` matrix in `.github/workflows/ci.yml` and in `.github/workflows/dsh.yml`, so Windows is a supported target rather than an incidental environment.

### Environment

- Windows 11, Git for Windows 2.49.0 (`core.autocrlf=true`, the installer default)
- Node v22.15.1, npm 11.3.0

### Suggested fix

Pin text files to `eol=lf` in `.gitattributes` and mark binary payloads explicitly. Every tracked file is already stored as LF in the index (`git ls-files --eol | grep -c "i/crlf"` returns `0`), so this renormalizes nothing and produces no index churn — it only stops the checkout filter from rewriting bytes.

I have this implemented and verified, and will open a PR referencing this issue. After the change: 736 tests, 711 pass, 4 fail, 21 skipped. The 4 remaining failures are present on `main` before the change and are a separate Windows packaging concern (`archive build …`, `package smoke …`, `cli: preview …`), overlapping the area of #132.

ethanstoner:
## Problem and value

Fixes #144.

`.gitattributes` marks generated artifacts `linguist-generated=true` but sets no `text`/`eol` attribute. With `core.autocrlf=true` — the Git for Windows installer default — a fresh clone materializes 373 tracked files with CRLF while the index stores LF. `git status` stays clean, so the divergence is invisible:

```console
$ git show HEAD:examples/checkout-platform-delta.html | head -c 16 | od -c
0000000   <   !   d   o   c   t   y   p   e       h   t   m   l   >  \n

$ head -c 16 examples/checkout-platform-delta.html | od -c
0000000   <   !   d   o   c   t   y   p   e       h   t   m   l   >  \r
```

Two consequences. Tests that compare renderer output against a checked-in artifact fail on bytes that were never actually different in the repository. And `scripts/build-zip.sh` stages tracked content, so a CRLF checkout feeds different bytes into `archify.zip` — a second byte representation of identical logical contents, which is the outcome the Node 22 pin in the zip builder exists to prevent (#119, #129). `build-zip.sh` is itself checked out with CRLF.

`windows-latest` is in the `package-smoke` matrix in `.github/workflows/ci.yml` and in `.github/workflows/dsh.yml`, so this is a supported target.

## Scope

- **What changed:** `.gitattributes` pins text files to `eol=lf` and marks `.gif` / `.jpg` / `.png` / `.zip` as `binary`. Adds `archify/test/repository-line-endings.test.mjs`, which asserts no tracked file materializes with CRLF.
- **What deliberately did not change:** no renderer, schema, validator, CLI, viewer, or typed JSON behavior. No file contents were renormalized. The existing `linguist-generated` rules are untouched. The four pre-existing Windows packaging failures are left alone — they are a separate concern overlapping #132.
- **No unrelated changes:** confirmed. Two commits, two files.

## Stability impact

- **Compatibility and migration risk:** none for the index. Every tracked file is already stored as LF — `git ls-files --eol | grep -c "i/crlf"` returns `0` — so this renormalizes nothing and produces no diff churn. It only stops the checkout filter from rewriting bytes on the way out. Linux and macOS checkouts are already LF and are unaffected. No tracked `.bat`, `.cmd`, or `.ps1` files exist that would want CRLF.
- **Renderer, validator, package, or generated-artifact risk:** none. `archify.zip` is unchanged and remains fresh: `scripts/build-zip.sh` excludes `archify/test` and `archify/test/*` explicitly, and the archive contains no test files, so the added test cannot enter the package. No skill runtime, schema, renderer, or `SKILL.md` content changed.
- **Failure behavior and rollback path:** the new test fails closed, listing every offending path. Rollback is reverting the two commits; because the index is untouched, a revert restores the previous behavior exactly.

Contributors who already have a CRLF working tree need one refresh to pick up the corrected attributes:

```bash
git ls-files -z | xargs -0 rm -f
git checkout -- .
```

## Tests run

All on Windows 11, Node v22.15.1, npm 11.3.0, Git 2.49.0, `core.autocrlf=true`, base `9a50605`.

Baseline on `main`, before the change:

```
cd archify && npm test
# tests 735
# pass 699
# fail 15
# skipped 21
```

After the change:

```
cd archify && npm test
# tests 736
# pass 711
# fail 4
# skipped 21
```

Eleven failures cleared. The 4 that remain are present on `main` before this change and are untouched by it: `archive build excludes untracked files and external symlinks from the live working tree`, `archive build is byte-for-byte reproducible across caller time zones without system zip`, `cli: preview runs from an installed skill without node_modules and exits cleanly`, `package smoke rejects every dependency metadata field in a built package`.

Targeted regression test, verified red then green:

```
# with .gitattributes from main, one file re-materialized
cd archify && node --test test/repository-line-endings.test.mjs
# tests 1
# pass 0
# fail 1

# with this change
cd archify && node --test test/repository-line-endings.test.mjs
# tests 1
# pass 1
# fail 0
```

Working tree verified after the change:

```console
$ git ls-files --eol | grep -c "w/crlf"
0
$ git status --porcelain
                                   # clean
```

One unrelated flake to report honestly: `benchmark never accepts a visual pass without an identified reviewer` (`test/ordinary-model-floor.test.mjs`) failed once with `INTERNAL_ERROR: Unexpected end of JSON input` — the spawned `verify` process's stdout read back truncated under `--test-concurrency=2`. It passed 18/18 in isolation and did not recur in the final run. It is not related to this change, and I did not attempt to fix it here.

## Visual evidence

Not applicable. No rendered output changes; this affects only the bytes a checkout writes to disk. Visual review: **skipped** — no visible surface is touched.

## Generated artifacts

None regenerated. `archify.zip` remains fresh because no skill runtime, schema, renderer, or `SKILL.md` content changed, and `scripts/build-zip.sh` excludes `archify/test` so the added test is not packaged. No Gallery, guide, start-page, or README proof inputs changed.

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,14 @@
+# Checked-in artifacts are compared byte-for-byte against renderer output and
+# staged into the canonical archify.zip. A CRLF checkout rewrites those bytes,
+# so pin every text file to LF in the working tree regardless of core.autocrlf.
+* text=auto eol=lf
+
+# Binary payloads must never be line-ending converted.
+*.gif binary
+*.jpg binary
+*.png binary
+*.zip binary
+
 # Generated diagram artifacts and built documentation surfaces.
 # Keep renderer/viewer templates and the hand-authored landing page visible to Linguist.
 /archify/examples/*.html linguist-generated=true
--- a/archify/test/repository-line-endings.test.mjs
+++ b/archify/test/repository-line-endings.test.mjs
@@ -1,0 +1,24 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { execFileSync } from 'node:child_process';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), '../..');
+
+function trackedFilesConvertedToCrlf() {
+  const output = execFileSync('git', ['ls-files', '--eol'], { cwd: repoRoot, encoding: 'utf8' });
+  return output
+    .split('\n')
+    .filter((line) => /\bw\/crlf\b/.test(line))
+    .map((line) => line.slice(line.indexOf('\t') + 1).trim());
+}
+
+test('tracked text files check out as LF so working-tree bytes match the index', () => {
+  const converted = trackedFilesConvertedToCrlf();
+  assert.deepEqual(
+    converted,
+    [],
+    `checked out with CRLF, so these no longer match their committed bytes and cannot be compared against renderer output or staged into archify.zip:\n${converted.join('\n')}`,
+  );
+});


tt-a1i:
感谢这个修复方向。#202 已合并到 main，覆盖了同一个 Windows CRLF / core.autocrlf 问题，并采用仓库级 LF 规范与真实 Git checkout 回归测试（覆盖 true/input/false，同时验证二进制字节不受影响）。为避免维护两套重叠实现，这个 PR 作为已被 #202 取代关闭。

PR #146, opened 2026-08-27
feat(workflow): add constraint-driven compiler

### LINKED ISSUE #126: workflow: edge length between adjacent columns is non-uniform, so identical specs pass or fail depending on which column pair you picked

sunsunsun-java:
## Summary
- add readable-v2 constraint-driven workflow compilation and causal semantic diagnostics
- preserve v1 compatibility while unifying layout, routing, labels, and bounds
- add semantic contract tests and a verified Maka workflow proof artifact

## Verification
- `npm test` (855 passed, 18 skipped, 0 failed)
- Maka workflow showcase validation: 9/9 checks passed
- Chrome visual checks: 1440x900 and 2048x1320, light/dark themes passed

The generated Maka artifact is based on source revision `f8be426ba81bc2f62e9cf0e08bac85f576cc9261`.

Fixes #126


tt-a1i (approved):
Reviewed exact head 40dda49aed7f9e15caded8ea4f2febccd99aac8c against #126. The v1 byte-compatibility boundary, fail-closed v1-to-v2 migration, deterministic constraint compiler, semantic diagnostics, and committed proof receipt are all covered by the current green checks; the focused local compiler/migration/semantic suite also passed 133/133. I found no blocking correctness issue. The material tradeoff is the long-term maintenance surface of workflow schema v2, and the owner has explicitly accepted that product direction.

PR #150, opened 2026-08-27
feat(cognition): add cognition diagram type for reasoning-route diagrams

### LINKED ISSUE #149: New diagram type proposal: cognition — render how a reasoning system routed one question through its own knowledge
## Problem

All five current types (architecture / workflow / sequence / dataflow / lifecycle) render **external systems**. There is no type for the other half of the picture: **a system rendering its own reasoning** — how one question was routed through a conditional knowledge base, which cards accepted or rejected it, and where the knowledge ends (blindspots).

We hit this concretely in [Lingshu](https://github.com/FuRongJun-1999/CommonTrustProtocol), an open-source conditional-knowledge engine: retrieving "triangle interior angles" routes through domain lanes, an annotation index, negative routes, and a confidence chain. We first forced this into a `workflow` diagram with a sidecar trace file (the typed IR correctly rejected our extra fields via `additionalProperties: false`) — the semantics simply do not fit the workflow contract.

## Cognitive semantics with no home in current types

| Cognitive semantic | Routing data | Current best expression |
|---|---|---|
| Conditional narrowing | domain lanes + two-stage convergence | workflow lanes (approx.) |
| Deterministic index hit | `card_hit: true` + score | mainPath (approx.) |
| Negative route / low-score branch | excluded-by-unapplicable-conditions, score=2 branches | role: branch/error (approx.) |
| Conditional-space declaration | `edu_level: E1–E5` | **no equivalent** |
| Knowledge-card direct answer | `direct_answer` | cards (approx.) |
| Confidence chain | score 7/2/2/2 ordering | **no equivalent** |

## Proposal

Add a sixth type, `cognition`, with routing semantics as first-class IR fields:

- `node.verdict: accept | branch | defer | reject | blindspot` — drives node fill + legend
- `node.card_hit: boolean` — deterministic-hit evidence (auditability anchor)
- `node.confidence` + `edge.width` — the visible confidence chain
- `node.edu_level: E1–E5` — conditional-space declaration (cognition-specific)
- `node.direct_answer` — full-sentence answer, auto-aggregated into the viewer card area
- `edge.role: main | branch | defer | reject` — routing hop semantics
- `meta.question` / `meta.verdict` — question banner and overall outcome

## Evidence this fits the pipeline

A draft implementation (schema + typed renderer + wiring) validates and delivers cleanly through the existing artifact checks, and end-to-end on real retrieval output:

- `validate cognition` → 9 artifact checks, 0 errors, 0 warnings
- `deliver cognition --quality showcase` → 9/9 pass
- golden artifacts re-rendered byte-identical; real retrieval output (5 nodes, 4 edges, zh-CN) delivers on first pipeline pass

Happy to open the PR against this issue if the direction looks right. Non-goals we'd suggest: no Mermaid import mapping, no changes to existing five types' contracts.

FuRongJun-1999:
Closes #149

## User problem

Agent/reasoning systems need to explain **how they routed one question through their own knowledge** — which conditional cards accepted the question, which branches scored lower, which were excluded by negative routes, and where knowledge ends (blindspots). All five existing types render external systems; none can express these routing semantics. Our workaround (forcing them into `workflow` + a sidecar trace file) is documented in #149.

## What changes

New sixth type `cognition`:

- `schemas/cognition.schema.json` — verdicts, deterministic `card_hit`, `confidence`, conditional-space `edu_level` (E1–E5), `direct_answer`, `edge.role`, `meta.question`/`meta.verdict`; `additionalProperties: false` throughout, shared `$defs` reused

- `renderers/cognition/render-cognition.mjs` — shared layout/geometry/diagnostics pipeline reused; verdict-driven fills, deterministic-hit tick, question banner, confidence-mapped edge widths, auto-aggregated direct-answer cards

- CLI wiring (`TYPES`, usage, demo example), shared registries (`START_TYPES`, semantic/relationship collections), regenerated validators, i18n legend/context keys (en + zh-CN), type-router rows in `SKILL.md`, schema README, package description

- Example + golden artifacts (both copies), `render-examples`/`golden` table rows

- The five existing types: zero schema/renderer contract changes

- Mermaid import mapping (non-goal per #149)

- Existing validation strictness (all new fields are typed and enumerated; unknown fields rejected as usual)

Purely additive. Existing schema-v1 documents of the other five types are untouched; no migration needed.

New type fails through the standard diagnostics pipeline (schema + layout + artifact checks with machine-readable `supportedFixes`). Rollback = revert the single commit; no shared state is mutated.

## Regenerated artifacts

Rebuilt from the merged source after syncing `main`:

- `examples/cognition-knowledge-route.html`, `archify/examples/cognition-knowledge-route.html` — re-rendered. The checked-in copies predated `main`'s switch to embedded WOFF2 fonts and still carried the `2.16.0` generator badge, so a fresh render no longer matched them.

- `archify/renderers/shared/generated-validators.mjs` — rebuilt from the combined schema source. The merge conflict in this generated file was resolved by regeneration, not by hand, per CONTRIBUTING §Packages and generated artifacts.

- `archify.zip` — rebuilt with Node 22 so canonical bytes stay reproducible. It now ships the four cognition payload files (renderer, schema, example JSON, example HTML) that earlier builds omitted.

No other artifact changed: `check:brand-marks`, `check:validators`, `check:release-identity`, and `test/golden.mjs` all pass on this head, so outputs whose inputs did not change remain byte-identical.

## Review follow-ups

- **Legend contract** — dropped the unused `legend.cognition.subject` translation (the legend is verdict-driven, so that key was dead), removed `entries.external` from the cognition legend schema because no renderer category backs it, and documented the cognition legend-key set in `schemas/README.md`.

- **Rationale accuracy** — `docs/cognition-design-rationale.md` now presents recursive nested subgraphs as a design direction rather than a schema-v1 contract: there is no `subgraph` field, no `part_of` edge type, and no renderer `direction` parameter, and `additionalProperties: false` rejects all of them.

- **Viewport containment** — the desktop-reader browser gate caught that the cognition example overflowed every tested viewport (at 1440×900 it rendered 1307px tall). A 720px-wide `viewBox` is scaled up to the reader width, so the diagram grew taller than the page. The lane width and column centres are now on the same 1080 scale the other five types use. The example source also declares `quality_profile: showcase`, matching every other bundled example — earlier the checked-in HTML claimed showcase while its source had none, and lowering the HTML to `standard` (rather than raising the source) hid the problem until the viewport gate ran.
- **Deliberately not changed** — the root `required` list keeps `mainPath` optional. No diagram type requires it (`workflow` does not either), and `schemas/README.md` lists it as a supported structural array rather than a mandatory field. Making `cognition` require it alone would break the cross-type consistency this PR is built on.

## Tests and final-artifact evidence

- `node bin/archify.mjs validate cognition examples/knowledge-route.cognition.json` → **9 artifact checks, 0 errors, 0 warnings**

- `node bin/archify.mjs deliver cognition ... --quality showcase` → **9/9 pass** (`sha256 be37001765f9`)

- `node test/golden.mjs` → **all checks passed** (cognition rows added for both packaged and repo copies)

- `npm run check:validators` → fresh

- End-to-end on real retrieval output from our conditional-knowledge engine (`graph_retrieve("三角形内角和是多少")` → 5 nodes / 4 edges / zh-CN → delivered HTML): the IR converts losslessly because the fields are first-class, not squeezed into a workflow shape

- Environment note: `cli.test.mjs` preview test and `skill-metadata` frontmatter test fail identically on vanilla `main` in this Windows checkout (CRLF / SIGTERM exit-code differences), so they are excluded from the evidence set; `generate-validators.test.mjs`, `delivery-contract`, `render-output-checks`, `relationship-lens`, `guided-views`, `legend-contract`, `cli.test.mjs` (rest) all pass

## Rollout / deploy impact

None at runtime; the type becomes available to Skill users immediately upon merge and ships with the next release.


fb0448387-pixel:
Hi

codeCraft-Ritik (commented):
You’ve done a great job introducing the new cognition diagram type while keeping the existing five types completely untouched. The separation of schema, renderer, CLI wiring, registries, i18n, examples, and golden artifacts is especially clean.

tt-a1i (changes requested):
Thanks for tackling a real gap: a typed cognition-routing IR could be valuable, and keeping it additive is the right direction. I reproduced two contract blockers on this exact head:

1. The new first-class routing fields do not drive the artifact. Starting from `knowledge-route.cognition.json`, I changed `triangleCard.confidence` from `0.88` to `0.01` and changed edge `e1.role` from `main` to `reject`; the rendered `<svg>` remained byte-identical. `renderEdgePath()` only reads `edge.variant` and an author-supplied `edge.width`, while `node.confidence`, `edge.role`, and `edu_level` are otherwise unused. This contradicts the PR’s “confidence-mapped edge widths” and typed negative-route claims: two materially different routing decisions can produce the same diagram unless authors duplicate those semantics into unrelated presentation fields. Please either derive/validate the presentation from these fields or narrow/remove the claimed contract, with behavioral tests that fail when the semantic fields stop affecting output.

2. `card_hit: true` renders the same badge twice. `renderNode()` interpolates `hitBadge` once beside the sigil/brand and again after the label/sub-label; the checked-in example therefore contains two identical circles at the same coordinates for its single hit. Please render it once and add a focused assertion for one badge per hit node.

The current golden row proves only that one fixture reproduces its checked-in HTML; it does not protect these cognition-specific semantics. Because those semantics are the main value of adding a sixth diagram type, I would not merge until they have direct regression coverage. The branch is also currently conflicting with main, so the corrected head will need a fresh sync and current required checks.

tt-a1i (changes requested):
Additional review of unchanged head5fec546863eede6980249d2d783e2dcf2c8fadaa. The prior non-functional confidence/role/edu_level fields and duplicate card-hit badge findings still reproduce; no duplicate review is being added for those alone.

New delivery/integration findings:

- [P2] Register cognition in the preview runtime as well as the main CLI. `node archify/bin/archify.mjs preview cognition archify/examples/knowledge-route.cognition.json --no-open` exits1: Unknown diagram type cognition. archify/bin/preview.mjs:18 still has only the original five types, despite the CLI's new type registration. Add a successful preview regression for the new type.
- [P2] Deliver the new type in archify.zip. The archive lacks the cognition renderer/schema/JSON/HTML examples (four files) and retains eight older payload entries. ZIP users do not receive the advertised capability. Rebuild with canonical Node22 from the final source.
- [P2] Regenerate the root cognition golden from its actual source/default profile. examples/cognition-knowledge-route.html:5001 hard-codes showcase, while the source has no quality_profile and a fresh render plus the packaged example are standard/advisory. Golden currently reports29pass/1fail for that real content difference, not a CRLF-only mismatch.

Generated-validator check passes; a showcase validate can succeed despite the separately reproduced semantic-field failures. No source edits or merge.

FuRongJun-1999:
Both review rounds addressed in f0f3b22:

**B1 — routing fields now drive the artifact.** Changed fields change the render:
- `node.confidence` → fill opacity (`0.55 + 0.45 × confidence`, clamped) plus a readable `conf N` annotation when ≠ 1 — low-belief nodes are visibly demoted.
- `node.edu_level` → conditional-space tier tag (bottom-left, E1–E5).
- `edge.role` → routing-hop styling: `main` solid full-strength / `branch` dashed (6 3, 0.9) / `defer` dotted (2 3, 0.6) / `reject` dashed (5 3, 0.45).

Reproduced your exact probe on the new head: `triangleCard.confidence` 0.88→0.01 and `e1.role` `main`→`reject` now produce a **changed** artifact (confidence annotation flips to `conf 0.01`, the main edge picks up the reject dash + 0.45 stroke-opacity).

**B2 — preview runtime registered.** `archify preview cognition examples/knowledge-route.cognition.json --no-open` no longer exits 1.

**B3 — card_hit badge deduped** (was emitted twice per node).

Golden examples re-rendered and `test/golden.mjs` passes all checks. One pre-existing failure on this platform (`skill-metadata` YAML frontmatter probe, reproducible on the unmodified base 5fec546) is unchanged and out of scope for this fix.

The design intent: these fields are the *semantics* of a routing diagram, not metadata — a reader should see at a glance which route won, which were demoted, and how much belief backed each step. Happy to adjust the specific visual encodings if you'd prefer different mappings.

FuRongJun-1999:
Rebased onto current `main` (31 upstream commits, now 199360c) and resolved the conflicts:

- `SKILL.md`: kept your `schema_version: 2` workflow-authoring guidance and re-added `cognition` to the type list.
- `generated-validators.mjs`: regenerated from source so the cognition schema is included in the shared validator bundle.
- Golden cognition examples re-rendered against the new renderer infrastructure; `test/golden.mjs` passes all checks and the B1 probe (confidence 0.88→0.01, role main→reject changes the artifact) still reproduces on the rebased head.

The branch is now conflict-free and ready for re-review.

PR #108, opened 2026-08-25
feat(i18n): localize Viewer UI from meta.locale

sunsunsun-java:
## Problem and value

Closes #106.

Chinese diagrams currently mix authored Chinese content with fixed English Viewer copy. This change lets the Agent choose an explicit locale in source data and gives every renderer one consistent Viewer language contract.

## Scope

- What changed:
  - Added optional `meta.locale` with first-phase support for `en` and `zh-CN`; omitted locale remains English.
  - Centralized renderer-owned Viewer messages and applied them to all five diagram modes, including controls, dynamic states, export feedback, accessibility text, `<html lang>`, and page title.
  - Kept authored content under Agent control and documented the unsupported-language contract: authored content may use that language while fixed Viewer UI falls back to English with an explicit disclosure.
  - Updated schemas, standalone validators, Skill guidance, README variants, CI coverage, development identity (`2.16.0-dev.0` / `2.16`), generated examples, Gallery artifacts, public pages, and `archify.zip`.
- What deliberately did not change:
  - No additional Viewer locale such as `zh-TW`, Japanese, or German was introduced.
  - Technical identifiers, product names, code, protocols, API paths, and environment names remain authored verbatim.
- No unrelated changes: confirmed. Local Maka regeneration and research directories are excluded from the commit.

## Stability impact

- Compatibility and migration risk: low. `meta.locale` is optional, so existing v1 JSON remains valid and keeps the previous English Viewer behavior.
- Renderer, validator, package, or generated-artifact risk: the shared catalog touches every renderer, so regression coverage checks all five modes and the rebuilt package.
- Failure behavior and rollback path: unsupported locale values fail schema validation instead of silently selecting the wrong locale. The feature is contained in this commit and can be reverted atomically.

## Tests run

- `npm test` in `archify/`: 724 tests total — 708 passed, 0 failed, 16 skipped by environment contract.
- `ARCHIFY_CHROME=/Applications/Google Chrome.app/Contents/MacOS/Google Chrome ARCHIFY_CHROME_NO_SANDBOX=1 node --test test/i18n.test.mjs`: 9 passed, 0 failed, 0 skipped. This exercises real Chrome interactions across all five diagram modes.
- `scripts/build-zip.sh archify.zip`: rebuilt 87 packaged files.
- Extracted-tree freshness comparison against a second build: passed.
- `node scripts/package-smoke.mjs <extracted-skill>`: passed on macOS.

## Visual evidence

[...]

PR #154, opened 2026-08-28
feat(i18n): add zh-TW as a Viewer locale

### LINKED ISSUE #152: feat(i18n): add zh-TW as a Viewer locale
## Problem and value

PR #108 shipped first-phase Viewer localization for `en` and `zh-CN`, and listed
`zh-TW` under "what deliberately did not change". This issue proposes that follow-up.

Traditional Chinese readers (Taiwan, Hong Kong) currently have two options, both lossy:

- omit `meta.locale` -> authored Traditional content sits inside an English Viewer
- set `zh-CN` -> the Viewer renders Simplified characters and mainland terminology
  right next to Traditional authored labels, mixing scripts on one screen

Mixed Traditional/Simplified on a single screen reads as a defect to Traditional
readers in a way that Traditional/English does not.

## Why this is not just character conversion

Deriving zh-TW from zh-CN mechanically (OpenCC and friends) is not sufficient and
would produce wrong output for this catalog:

- one-to-many mappings: `复制` -> `複製`, never `復制` (复 = 複 / 復 / 覆)
- terminology, not glyphs:
  `数据库` -> `資料庫` (not `數據庫`), `云服务` -> `雲端服務`,
  `消息总线` -> `訊息匯流排`, `导出` -> `匯出`, `剪贴板` -> `剪貼簿`,
  `加载` -> `載入`, `默认` -> `預設`, `视口` -> `視埠`, `语义` -> `語意`

zh-TW therefore needs a hand-authored third column, not a derived one.

## Proposed scope

- `SUPPORTED_LOCALES` -> `['en', 'zh-CN', 'zh-TW']`
- `MESSAGE_PAIRS` tuples 2 -> 3 elements (518 keys)
- `common.schema.json` `locale` enum + regenerated `generated-validators.mjs`
- `SKILL.md` unsupported-language disclosure text
- `i18n.test.mjs` coverage, README variants, regenerated examples / gallery / `archify.zip`

## Non-goals

- No other locale (ja, de, ...)
- No change to how authored content is handled
- No change to the `omit locale -> English + explicit disclosure` contract
- Technical identifiers, product names, protocols, and API paths stay verbatim

## Evidence from a local proof

I patched `MESSAGE_PAIRS` in a local copy to serve Traditional Chinese for 106
demo-visible keys (toolbar, legend, guided views, nav dock, finder, common actions),
with untranslated keys falling back to English rather than Simplified.

- `deliver --quality showcase` -> 9/9 artifact checks, 0 errors, 0 warnings
- `visual-check` -> containment passes at 1440x900, 1600x1000, 1920x1080, 2048x1320
- character scan of the artifact -> 0 Simplified characters
  (control: the `zh-CN` build of the same source contains 32 distinct Simplified
  characters across several hundred occurrences, so the scan is not vacuous)

Happy to prepare the full 518-key PR if you agree on the scope and the
"hand-authored, not converted" constraint.

neoeric:
## Problem and value

Closes #152.

PR #108 shipped first-phase Viewer localization for `en` and `zh-CN` and listed `zh-TW` under "what deliberately did not change". This is that follow-up.

Traditional Chinese readers currently have two lossy options: omit `meta.locale` and put authored Traditional content inside an English Viewer, or set `zh-CN` and mix Simplified Viewer chrome with Traditional authored labels on one screen. Mixed script reads as a defect in a way that Traditional/English does not.

Opened as a **draft**: CONTRIBUTING asks for scope agreement before implementing a schema change, and #152 has not been answered yet. The code is here for technical feedback, and I will rework or drop it if you want a different shape.

## Scope

**What changed**

- `SUPPORTED_LOCALES` is `['en', 'zh-CN', 'zh-TW']`; all 518 `MESSAGE_PAIRS` tuples gain a third element.
- `common.schema.json` `locale` enum accepts `zh-TW`; `generated-validators.mjs` regenerated via `npm run generate:validators`.
- `SKILL.md`, `references/authoring-contract.md`, `schemas/README.md`, `README.md`, `README_EN.md`, `README_ZH.md` updated for the three-locale contract.
- The authoring contract now states that `zh-CN` and `zh-TW` are never substituted for one another; `skill-metadata.test.mjs` asserts that wording.
- `archify.zip` rebuilt with `scripts/build-zip.sh`.

**What deliberately did not change**

- No other locale (ja, de, …).
- No change to how authored content is handled — the renderer still never translates it.
- No change to the `omit locale → English + explicit disclosure` contract for unsupported languages.
- Technical identifiers, product names, protocols, and API paths stay verbatim.

## Why the catalog is hand-authored

Deriving `zh-TW` from `zh-CN` by script would be wrong here in two distinct ways:

1. **One-to-many mappings.** `复` maps to 複 / 復 / 覆. `复制` must become `複製`; a naive converter yields `復制`.
2. **Regional terminology, not glyphs.** `数据库` is `資料庫`, not `數據庫`. Also `云服务` → `雲端服務`, `消息总线` → `訊息匯流排`, `导出` → `匯出`, `剪贴板` → `剪貼簿`, `加载` → `載入`, `默认` → `預設`, `视口` → `視埠`, `语义` → `語意`.

A glyph-only conversion produces Traditional characters carrying mainland vocabulary, which is not what a Taiwanese reader expects.

Low risk. `meta.locale` stays optional and `zh-TW` is purely additive, so every existing schema-v1 document remains valid and unchanged. No renderer geometry, layout rule, diagnostic code, or export path is touched.

An unsupported locale still fails schema validation with a `/meta/locale` diagnostic (the existing test uses `fr` and `zh-HK`, both untouched). The catalog completeness loop still throws `Incomplete Archify i18n tuple` if any tuple is short, so a partial locale cannot ship. Rollback is reverting the two commits and rebuilding the ZIP.

## Tests and evidence

**Added coverage** (`archify/test/i18n.test.mjs`)

- `zh-TW carries no Simplified-only characters`
- `zh-TW is authored separately from zh-CN`
- `zh-TW uses Taiwan terminology rather than mainland equivalents`

The existing `every supported catalog is complete and preserves interpolation variables` iterates `SUPPORTED_LOCALES`, so it now covers `zh-TW` automatically. All 518 keys were checked to carry exactly the same `{placeholder}` set as their English source.

**Suite result — measured against a baseline, not in isolation**

Run on Windows 11 / Node v22.15.0:

| | tests | pass | fail |
|---|---|---|---|
| clean `main` | 744 | 690 | 26 |
| this branch | 747 | 693 | 26 |

The same 26 tests fail before and after, by name. They are environment failures on a Windows checkout — symlink-guard tests, archive/package gates, and the CRLF-sensitive frontmatter and locale tests — and are consistent with #144. **This branch introduces no new failure and adds three passing tests.** I did not attempt to fix the pre-existing 26; they look orthogonal.

**Final-artifact evidence**

Rendering an architecture diagram with `meta.locale: "zh-TW"`:

- `deliver --quality showcase` → 9/9 artifact checks, 0 errors, 0 warnings
- `<html lang="zh-TW">` emitted
- character scan of the artifact → 0 Simplified characters (control: the `zh-CN` build of the same source contains 32 distinct Simplified characters over several hundred occurrences, so the scan is not vacuous)

**ZIP**

`scripts/build-zip.sh` run twice produced byte-identical output (1,258,050 bytes, 70 files). Note for reviewers on Windows: the script needs `TMPDIR` set to a Windows-visible path, otherwise the inline `node -e` step resolves the `mktemp -d` result as `C:\tmp\…`. I did not change the script — flagging it in case it is worth a separate issue.


a951753abc:
Thanks for picking this up — the terminology reads as natural Taiwanese Traditional Chinese throughout (資料／匯出／剪貼簿／預設／支援／非同步), and I found no Simplified vocabulary leaking into the zh-TW column after scanning all 518 tuples. +1 on the scope in #152 as a Traditional Chinese reader.

One consistent issue: **38 zh-TW strings use half-width punctuation (`,` `:` `;`) directly adjacent to CJK text.** In Traditional Chinese UI copy these should be full-width `，` `：` `；` (the zh-CN column and the rest of the zh-TW column already use full-width `。`, so this looks like a conversion artifact rather than a style choice). Affected keys:

- `viewer.preset.current`: 視覺風格:{style}。選擇視覺風格
- `viewer.motion.yielding`: 暫停動態效果;目前讓位給{owner}
- `viewer.export.failed`: 匯出失敗:{message}
- `viewer.export.unknownVariant`: 未知的分享卡片類型:{variant}
- `viewer.export.routeRequired`: 請先追蹤路徑,才能匯出路徑分享卡片
- `viewer.export.reachRequired`: 請先追蹤編寫的可達範圍,才能匯出可達範圍分享卡片
- `viewer.export.routeFailed`: 路徑分享卡片匯出失敗:{message}
- `viewer.export.reachFailed`: 可達範圍分享卡片匯出失敗:{message}
- `viewer.export.copyFailed`: 複製失敗:{message}
- `viewer.export.card.routeSummary.one` / `.other`: 路徑:{source} → {target} · …
- `viewer.export.error.webmRequirements`: …追蹤動畫,以及瀏覽器支援 MediaRecorder
- `viewer.guided.intro`: 沿著精選路徑逐步檢視,不會改動原始圖表。
- `viewer.guided.storyTrail`: {label} 的故事軌跡:{count} 個段落
- `viewer.guided.chapter.open` / `.current` / `.current.title` / `.delta.aria`
- `viewer.guided.share.settled`: 路徑已定位,可供閱讀。
- `viewer.guided.share.aria`
- `viewer.guided.beat.aria.reverse` / `.multiple` / `.group`
- `viewer.guided.caption.direction`: 編寫方向:{from} → {to}
- `viewer.guided.beatLink`
- `viewer.guide.map.hint` / `viewer.guide.lens.hint` / `viewer.guide.present.hint`
- `viewer.finder.result.routeTarget`: 選擇{label}作為路徑終點,{links}
- `viewer.passport.relationship.help` / `.inspect`
- `viewer.passport.beacon.one` / `.other`
- `viewer.route.instructions`, `viewer.lens.instruction`, `viewer.radar.surface`, `viewer.radar.compacted`, `viewer.intent.summary`

A one-line check that catches all of them, in case you want to add it to `i18n.test.mjs` for the zh-CN/zh-TW columns:

```js
/[一-鿿][,:;!?]|[,:;!?][一-鿿]/.test(message)
```

Minor, non-blocking: 時序圖 is fine, though 循序圖 is the more common Taiwanese rendering of "sequence diagram" — either works.

neoeric:
Thank you — this was the right catch, and the count was exact.

**Fixed.** 51 zh-TW strings carried half-width `,` `;` `:`; all of them now use `，` `；` `：`. Running your expression over the catalog confirms your number: exactly 38 of those 51 matched it. The other 13 are the same conversion artifact with a `{placeholder}` on one side instead of CJK, so the character was never adjacent to a Han glyph — `viewer.export.copyFailed`-style keys where the punctuation sits between a placeholder and a word. Leaving those would have kept the column internally inconsistent and let a regression slip past any guard, so they were converted too.

**Guard added,** slightly stricter than what you proposed. Rather than matching adjacency, it asserts that neither Chinese column contains an ASCII `,` `;` `:` at all, and names the offending key on failure:

```js
test('Chinese catalogs use full-width sentence punctuation', () => {
  for (const locale of ['zh-CN', 'zh-TW']) {
    for (const key of catalogKeys()) {
      const message = translateMessage(locale, key);
      const found = [...message].filter((character) => ',;:'.includes(character));
      assert.deepEqual(found, [], `${locale} ${key}: ${message}`);
    }
  }
});
```

The adjacency form would not have caught those 13. Your exact regex now reports 0 keys in both `zh-CN` and `zh-TW`.

One exemption, deliberate: `viewer.nav.guide.title` keeps its half-width `?` — it names a keyboard key rather than ending a sentence, and `zh-CN` keeps it half-width for the same reason. That is why the guard covers `,;:` and not `?`.

**One width difference I did not change,** in case you disagree: all 19 keyboard-shortcut tooltips use half-width parentheses in zh-TW (`切換主題(T)`) where zh-CN uses full-width (`切换主题（T）`). Unlike the punctuation above, each column is internally consistent here, so it is a choice rather than an artifact — half-width avoids the wide gap around a single ASCII key. Happy to align with zh-CN if you would rather the two Chinese columns match.

On 時序圖 / 循序圖 — noted, and thank you for flagging it as non-blocking. Keeping 時序圖.

**Also in this push:** the branch is merged with `main` and no longer conflicts. 2.16.0 ships as `main` wrote it and the zh-TW entry moved up into `[Unreleased]`; `generated-validators.mjs` was regenerated from the merged schemas rather than hand-merged, so it carries both `locale: ["en","zh-CN","zh-TW"]` and `schema_version: [1,2]`; `archify.zip` was rebuilt from the merged tree.

One thing I deliberately left open: `npm run check:release-identity` now fails, because `[Unreleased]` has content again while `package.json` is still the stable `2.16.0`. `main` has not opened the next development cycle since the 2.16.0 release, so this branch would be the PR that does it — the way #108 moved to `2.16.0-dev.0` after 2.15.0. That is a maintainer call rather than something this PR should decide, so I have left it alone — happy to push the bump as soon as @tt-a1i names the core version.

sunsunsun-java:
Thank you for the thoughtful implementation and for addressing the terminology feedback. After reviewing the current product scope, we’ve decided not to expand the supported Viewer locale set beyond en and zh-CN at this time.
We’re closing this PR as a roadmap and maintenance-scope decision, not because the contribution lacks value. The implementation and discussion will remain available for reference if Traditional Chinese localization enters the roadmap later. No further rebasing, release-version work, or conflict resolution is needed.

PR #159, opened 2026-08-28
fix(site): preserve language across navigation

### LINKED ISSUE #157: fix(site): preserve the selected language across site navigation
## Problem

The selected site language is not preserved when navigating between Archify GitHub Pages surfaces.

Verified against `main` at `316a184` on 2026-08-28.

## Reproduction

1. Open https://tt-a1i.github.io/archify/.
2. Switch the landing page to Chinese.
3. Click the “验证作品集” / Proof Lab navigation link.

## Actual behavior

The landing page has `<html lang="zh-CN">`, but Proof Lab opens with `<html lang="en">` and English copy.

The same contract is inconsistent across other inner pages:

- landing page reads and writes `archify-lang`;
- Proof Lab reads and writes `archify-gallery-language`;
- Guide reads and writes `archify-guide-language`;
- Start uses an explicit `?lang=` value or browser language and does not inherit the selected site language.

## Expected behavior

A language selected on any Archify site page remains selected while navigating across the landing page, Proof Lab, Guide, and Start. An explicit supported `?lang=en|zh` value should take precedence when present.

## Root cause

The static site pages use independent persistence keys and fallback rules, so same-origin navigation does not share one site-level locale preference.

## Proposed scope

- use `archify-lang` as the canonical site language preference;
- retain legacy page-specific keys only as migration fallbacks;
- preserve explicit Start `?lang=` precedence;
- update generated Gallery, Guide, and Start pages from their templates;
- add regression coverage for persistence-key consistency and the cross-page navigation path.

## Non-goals

- no change to diagram-authored content;
- no change to Viewer `meta.locale` behavior;
- no new locale;
- no schema, renderer, package-version, or release-identity change.

## Acceptance criteria

- Chinese landing page → Proof Lab remains Chinese;
- Chinese landing page → Guide remains Chinese;
- Chinese landing page → Start remains Chinese when no explicit language override is supplied;
- explicit `?lang=en|zh` wins and becomes the subsequent site preference;
- English navigation remains English;
- generated pages are fresh and relevant tests pass.

YunyueLi:
## Problem and value

Closes #157.

Selecting Chinese on the landing page currently resets to English when navigating to Proof Lab because each GitHub Pages surface owns a different persistence key. Guide has the same split state, and Start ignores the selected site language unless a query parameter or browser preference happens to match.

This change gives the public site one predictable language preference while preserving explicit URL overrides and existing user choices.

## Scope

- What changed:
  - added a shared `ArchifySiteLanguage` runtime using `archify-lang`;
  - made supported `?lang=en|zh` values take precedence;
  - migrated existing Proof Lab and Guide preferences through read-only legacy-key fallbacks;
  - made Landing, Proof Lab, Guide, and Start read and write the same preference;
  - made storage failures fall back safely instead of aborting page initialization;
  - regenerated the checked-in Proof Lab, Guide, and Start pages;
  - added a focused runtime and site-integration regression test.
  - diagram-authored content, Viewer `meta.locale`, schemas, renderers, package identity, and release artifacts;
  - supported locales remain English and Simplified Chinese.

## Stability impact

- Compatibility and migration risk: low. `archify-lang` is already the landing-page key. Existing `archify-gallery-language` and `archify-guide-language` values remain readable as migration fallbacks and are copied into the canonical preference on use.
- Renderer, validator, package, or generated-artifact risk: no renderer, validator, schema, package, or ZIP bytes changed. Only the three site pages generated from edited templates were rebuilt.
- Failure behavior and rollback path: blocked or unavailable storage falls back to URL, browser language where explicitly enabled, or English. Reverting this commit restores the previous per-page behavior.

## Tests run

- Pre-fix regression: `node --test test/site-language-continuity.test.mjs` → 0 passed, 2 failed.
- Focused post-fix suite: `node --test test/site-language-continuity.test.mjs test/gallery.test.mjs test/guide-page.test.mjs test/start-page.test.mjs` → 10 passed, 0 failed.
- Full suite on Node.js 24.19.0: `npm test` from `archify/` → 720 passed, 0 failed, 28 environment-gated browser tests skipped.
- `git diff --check` → passed.

## Visual evidence

Visual review: passed in a real browser against a local static server.

- Before, production at `316a184`: Landing `zh-CN` → Proof Lab `en`.
- After, this branch: Landing `zh-CN` → Proof Lab `zh-CN` → Guide `zh-CN` → Start `zh-CN`.
- Explicit `?lang=en` on Proof Lab persisted English through Guide, Start, and Landing.
- Headings and `<html lang>` were checked at every step.
- No screenshot is attached because this is a navigation-state correction rather than a layout change, and the original report image contained personal chat identity that is intentionally not republished.

The locally generated current-main gallery artifacts emit an existing `MutationObserver` console error inside an embedded artifact. No gallery artifact file differs in this PR; the site-language runtime and top-level Guide/Start/Landing pages introduced no console error.

## Generated artifacts

Regenerated from their canonical templates:

- `docs/gallery.html`
- `docs/guide.html`
- `docs/start.html`

`docs/index.html` and `docs/assets/site-language.js` are direct site sources. `archify.zip` remains unchanged because no packaged Skill file changed.

## Checklist


sunsunsun-java:
Thanks for the focused fix. I re-reviewed the current head (`96c6aff`) and verified the normal Landing → Gallery → Guide → Start path in real Chrome: the shared preference works there. I found two reproducible functional issues and one coverage gap that should be addressed before merge.

1. **Custom builds omit the new runtime dependency.**

   `scripts/start-template.html`, `guide-template.html`, and `gallery-template.html` now load `assets/site-language.js`, but their builders do not copy that file to a custom output directory. For example:

   ```sh
   node scripts/build-start.mjs /tmp/site/start.html
   ```

   succeeds, but `/tmp/site/assets/site-language.js` is absent. Opening the generated page throws `ArchifySiteLanguage is not defined`, so Start interactions do not initialize. Guide and Gallery reproduce the same failure. The checked-in `docs/` deployment is unaffected because the asset already exists there, but arbitrary build/preview/deployment destinations are broken. Please either copy the runtime beside every generated output or inline it into the templates.

2. **A manual language switch is reverted on refresh when the page was opened with `?lang=`.**

   Reproduction in Chrome:

   - open `guide.html?lang=en`;
   - click 中文: the page becomes `zh-CN` and storage becomes `zh`, but the URL remains `?lang=en`;
   - refresh: the stale query wins again and storage is overwritten to `en`.

   Landing and Gallery have the same pattern. Start already avoids it by updating the URL from `render()`. Please update/remove the `lang` query when the user manually toggles, while preserving query precedence on initial load.

3. **The regression test does not execute the required cross-page path.**

   `site-language-continuity.test.mjs` regex-checks that each HTML file contains shared `read()`/`write()` calls, but it never loads pages sequentially, follows navigation, refreshes, or verifies that the generated runtime dependency exists. Both failures above therefore pass CI. Issue #157 explicitly asks for persistence-key consistency *and* cross-page navigation coverage. An artifact/browser-level test should cover Chinese and English navigation, explicit `?lang=en|zh`, refresh after a manual toggle, and a custom builder output.

All current GitHub checks are green, and generated Gallery/Guide/Start files are fresh. The remaining concerns are the two runtime edge cases above and the missing behavioral regression coverage.

YunyueLi:
Implemented a root-cause revision for the language continuity and cross-page navigation issue. The final implementation is in `d8c14cb`; the protected PR branch now points to `91c98cb` without rewriting its existing history.

### What was wrong

The first revision unified the storage key, but it did not establish one authoritative language state for the whole site:

- A stale `?lang=` value could continue to override a manual language selection after refresh.
- Generated Gallery, Guide, and Start pages referenced shared runtime assets without making their builders responsible for copying those assets into custom output directories.
- Navigation markup and styling were still owned independently by each page, so the pages could drift visually and functionally.
- The tests mostly checked source text and did not reproduce the actual entry, navigation, selection, refresh, custom-build, or responsive-layout journeys.

### Root-cause fix

1. **One durable language authority**
   - `archify-lang` is the only persisted preference.
   - `?lang=en` and `?lang=zh` are treated as one-time entry instructions: the value is stored, then the language parameter is removed from the URL with unrelated query parameters and the hash preserved.
   - After entry, page navigation, manual selection, and refresh all read the same persisted preference.
   - The default is explicitly English. Legacy keys and browser-language guessing were removed.

2. **Complete generated-site outputs**
   - The Gallery, Guide, and Start builders now copy every shared site asset required at runtime.
   - Custom output directories therefore contain both the shared language runtime and shared navigation stylesheet, rather than depending on the repository layout by accident.

3. **One navigation contract across all public pages**
   - Landing, Proof Lab, Guide, and Start now consume the same navigation stylesheet and expose the same primary destinations, language control, GitHub link, and Install Skill action.
   - Page-specific controls remain inside their own page content instead of changing the global navigation.

4. **Behavioral regression coverage**
   - Added artifact-level tests for all custom builders and their emitted shared assets.
   - Added a real Chrome journey covering explicit-language entry, Landing → Proof Lab → Guide → Start navigation, manual language changes, and refresh persistence.
   - Added desktop and mobile assertions for the shared navigation contract and computed layout parity.
   - Wired the browser regression into the existing Chrome-backed `test:webm` CI command, so no workflow-file change or additional GitHub Actions permission is required.

### Verification completed

- Real Chrome site regression: **4/4 passed**, including entry, navigation, selection, refresh, and responsive navigation checks.
- Full repository suite: **750 tests, 721 passed, 29 environment-gated skips, 0 failed**.
- Generated-page reproducibility and custom-output asset tests passed.
- Manual visual QA passed for all four pages at desktop `1440×900` and mobile `390×844` viewports.
- The branch is synchronized with the latest `main` content available before this update and has no net change to `.github/workflows/ci.yml`.

This addresses the three review findings directly and adds the missing cross-page navigation consistency requirement. Ready for re-review once the current CI run completes.

YunyueLi:
Follow-up on the first CI run (`686c018`): the product regression assertions were sound, but the initial CI placement exposed two execution-layer problems that are now corrected.

- The generic Node matrix ran the custom multi-page build integration concurrently with the repository suite. That additional process load exposed an existing timing-sensitive output-path test on every Node runner. The custom build integration is now isolated in the serialized Chrome-backed gate; the generic matrix retains the deterministic runtime and shared-navigation contract tests.
- The new real-Chrome journey initially inherited the WebM step without the explicit sandbox opt-out required by GitHub-hosted Linux runners. A dedicated integration entrypoint now applies that opt-out only on GitHub Actions Linux, matching the repository existing real-browser policy. No product runtime behavior and no workflow file were changed for this correction.

Verification after the CI-layer correction:

- Serialized site integration: **4/4 passed** locally with real Chrome.
- The previously affected concurrent test pairing: **5/5 repeated runs passed**.
- Full repository suite: **750 tests, 720 passed, 30 intentionally environment-gated skips, 0 failed**.

The new CI run is now in progress. I will treat this revision as complete only after the remote checks pass.

YunyueLi:
Final verification update: the rerun for head `686c018` is fully green.

- Node test matrix: **18, 20, 22, and 24 passed**.
- Package smoke tests: **Ubuntu, macOS, and Windows passed**.
- Generated ZIP freshness: **passed**.
- Chrome-backed WebM and site integration gate: **passed**, including the real Landing → Proof Lab → Guide → Start language continuity, manual selection, refresh persistence, custom-build assets, and desktop/mobile navigation checks.

Result: **9/9 required CI checks passed**. The PR is ready for re-review.

docs/assets/site-language.js
```diff
@@ -0,0 +1,53 @@
...
+    } catch (_) {
+      return null;
+    }
+  }
+
+  function read() {
+    var requested = consumeRequestedLanguage();
+    if (requested) return write(requested);
+
+    var stored = readStored(STORAGE_KEY);
```
tt-a1i, line 43: Blocking compatibility regression: this reads only `archify-lang`, then defaults to English. Issue #157 explicitly scopes `archify-gallery-language` and `archify-guide-language` as migration fallbacks, and the PR description says existing choices remain readable. A user who previously selected Chinese on Gallery or Guide but has no canonical key will lose that preference after this change. Please fall back to the legacy keys when the canonical key is absent, persist the migrated value to `archify-lang`, and add regression cases for both keys.

tt-a1i (changes requested):
I re-reviewed the current head and verified that CI is 9/9 green and the real-Chrome site journey passes 4/4 locally. The shared runtime, custom-output assets, query consumption, refresh persistence, and navigation contract look sound. One compatibility issue remains before merge: the implementation drops the legacy Gallery and Guide language preferences even though Issue #157 and this PR description promise migration fallbacks. Users with only an existing `archify-gallery-language` or `archify-guide-language` value will silently revert to English. Please add a one-time fallback migration into `archify-lang` and cover both legacy keys with regression tests.

tt-a1i (dismissed):
Re-reviewed fixed head `5e60f85` after the compatibility repair.

The blocking legacy-language regression is resolved: supported URL overrides remain authoritative, the canonical preference wins when present, both historical Gallery and Guide keys migrate once into `archify-lang`, invalid values continue through the fallback chain, and storage/history failures remain non-fatal. The migration is also covered in the real Chrome cross-page journey.

Verification:

- three independent post-implementation reviews: no actionable findings
- focused runtime and page tests: passed
- real Chrome site integration: 4/4 passed
- full local suite: 720 passed, 30 environment-gated skipped, 0 failed
- remote required checks: 9/9 passed
- `git diff --check`: passed

Approved at the exact head above.

PR #163, opened 2026-08-28
chore(security): require immutable workflow dependencies

simaba:
Pin remote GitHub Actions and reusable workflow dependencies to immutable commits, retaining readable version comments. This prevents a moved upstream tag from silently changing CI behavior. Runtime, diagram output, job logic, permissions, triggers and deployment targets are unchanged.

Candidate 4b021058a7cc6e98d113539974034f2282715e55 is refreshed on dev d03f9f5. The refresh also pins checkout/setup-node in the newly added Windows portability job. Contributor-card action pins already landed with #412 and remain intact. Existing author commits are preserved.

Verification:
- Resolved all 12 distinct action/tag pairs used in workflow comments through the official repositories' Git refs API, dereferencing annotated tags; all match their pinned commits.
- 23 focused action-pinning and release/package-gate tests passed. Two environment-specific cases skipped on local Node 22/macOS: non-22 rejection and Windows absolute paths. Hosted matrix evidence remains required.
- Actionlint structural checks passed. Its optional ShellCheck pass reports existing SC2016 on the intentionally single-quoted JavaScript template literal in the stable-manifest step; that script is unchanged by this PR.
- git diff --check passed. Packaged Skill/ZIP unchanged.

The regression scanner checks immutable remote GitHub references, including reusable workflows, and excludes local and Docker actions. Existing release/Pages guards remain asserted. Current-head CI and DSH checks remain required before merging into dev; no release or Pages deployment was triggered by this review.


tt-a1i (changes requested):
Thanks for the supply-chain hardening. Reviewed exact head e0c652569713891d4f66b638463a84b60d2bd659.

[P2] Update the existing Pages contract test to understand the immutable action reference. The workflow now uses configure-pages by full SHA, but archify/test/release-package-gates.test.mjs:161 still requires /actions\/configure-pages@v5/. Reproduced on Node24: the focused pinning/release/readme group has 19 passed, 1 failed, 5 canonical-Node22 skips. Current CI run33374069471 job99431466434 fails on the same assertion, so this is a specific test regression introduced by the pinning, not merely a missing-CI gate.

Keep the test checking the intended Pages action and authorization/dependency structure while accepting/verifying its pinned commit. All eight new action pins were independently resolved against their stated versions and match; no separate pin correctness issue found. No source edits or merge.

simaba:
Addressed the requested change on current head `782120937fae69f50f53db008b1677896ea3a82f`.

- Merged current `main` (`199360cc6687a7857b54dd188d4922b09e466a4b`) so the PR follows the newer action majors already adopted upstream.
- Kept the existing Pages authorization/dependency assertions intact.
- Updated `release-package-gates.test.mjs` to verify the intended action identity **and the exact immutable commit** (plus readable version comment) for `configure-pages`, `upload-pages-artifact`, `deploy-pages`, and `action-gh-release`, instead of requiring mutable `@vN` syntax.
- The workflow-wide pinning regression remains in place; an independent scan of the PR head found 29 remote `uses:` references and zero mutable refs.
- Re-resolved all eight newly pinned action/version comments against their exact Git tags; every pinned commit matches.
- CI is green across Node 18/20/22/24, WebM, ZIP freshness, published-manifest, and package-smoke on Ubuntu/macOS/Windows.
- Node 22 full suite: 1023 tests, 996 passed, 0 failed, 27 skipped.
- DSH package contract and DSH distribution acceptance are both green.

The PR description has also been refreshed to match the current six-file diff and validation state.

PR #170, opened 2026-08-28
fix(geometry): deduplicate consecutive identical points in polylinePath

### LINKED ISSUE #169: [Bug]: Vertical auto-routed data-flow edges emit a zero-length final segment, so marker-end renders at the wrong angle
### Archify version or commit

`2.16.0-dev.0` (skill package `SKILL.md` metadata `version: "2.16"`)

### Installation method

Downloaded archify.zip

### Diagram type

Data flow

### Exact command

```shell
node bin/archify.mjs validate dataflow <path>/repro.dataflow.json --quality showcase --json   # exit 0
node bin/archify.mjs render   dataflow <path>/repro.dataflow.json <path>/repro.html --quality showcase   # exit 0
```

### Summary

A purely vertical auto-routed flow emits a **zero-length final segment** in its `d` attribute. Because SVG derives `marker-end` orientation from the direction of the last path segment, that direction is undefined and the arrowhead falls back to 0°, so it renders lying flat (pointing left) on the target node's top edge instead of pointing into it.

Validation does not catch this: `composition.metrics.microSegmentCount` is `0`, because a *zero*-length segment is not a *micro*-segment and never trips the 8px floor. So `validate`, `render`, `deliver`, and `visual-check` all pass while the artifact renders a visibly wrong arrowhead.

```json
{
  "schema_version": 1,
  "diagram_type": "dataflow",
  "meta": { "title": "Vertical edge arrowhead repro", "quality_profile": "showcase" },
  "stages": [{ "label": "Parse" }, { "label": "Store" }],
  "nodes": [
    { "id": "extract", "type": "backend", "label": "Extract", "stage": 0, "row": 0 },
    { "id": "parse", "type": "backend", "label": "Parse", "stage": 0, "row": 2 },
    { "id": "store", "type": "database", "label": "Store", "stage": 1, "row": 1 }
  ],
  "flows": [
    { "id": "vertical-edge", "from": "extract", "to": "parse", "label": "chunks", "variant": "emphasis" },
    { "id": "horizontal-edge", "from": "parse", "to": "store", "label": "rows", "variant": "emphasis" }
  ]
}
```

Emitted paths from the rendered HTML:

```
vertical-edge      d="M 100 186 L 100 186 L 100 356 L 100 356"   <-- first AND last segments zero-length
horizontal-edge    d="M 156 385 L 207.5 385 L 207.5 271 L 259 271"
```

`vertical-edge` renders a flat, left-pointing arrowhead. `horizontal-edge`, whose final segment has real length, renders correctly.

A one-line detector over any generated artifact:

```js
// true when the last two points are identical -> marker-end angle is undefined
const pts = [...d.matchAll(/[ML] ([-\d.]+ [-\d.]+)/g)].map(m => m[1]);
const degenerate = pts.at(-1) === pts.at(-2);
```

```json
{
  "ok": true,
  "command": "validate",
  "type": "dataflow",
  "input": "<path>/repro.dataflow.json",
  "checks": [
    { "name": "single_svg", "ok": true },
    { "name": "finite_svg", "ok": true },
    { "name": "orthogonal_arrows", "ok": true },
    { "name": "label_route_clearance", "ok": true },
    { "name": "relationship_crossings", "ok": true },
    { "name": "relationship_corridors", "ok": true },
    { "name": "container_border_runs", "ok": true },
    { "name": "route_rhythm", "ok": true },
    { "name": "legend_clearance", "ok": true }
  ],
  "composition": {
    "profile": "showcase",
    "status": "pass",
    "summary": { "errors": 0, "warnings": 0 },
    "metrics": {
      "microSegmentCount": 0,
      "minSegmentPx": 51.5,
      "minInteriorSegmentPx": 114,
      "shortInteriorSegmentCount": 0
    }
  }
}
```

### Expected behavior

A vertical edge entering a node's top edge renders an arrowhead pointing **down** into that node, the same way a horizontal edge points into a left edge.

Two candidate fixes:

1. **Collapse duplicate consecutive points before emitting `d`.** `M 100 186 L 100 356` orients `marker-end` correctly with no other change, and is the smallest fix.
2. **Add a composition check for zero-length segments.** The existing micro-segment rule has an 8px floor; a segment of exactly 0px is arguably the more severe case but currently passes silently. Catching it would have surfaced this without a human noticing the arrowhead.

### Actual behavior

The arrowhead renders rotated to 0° (flat, pointing left), detached from the line's direction, sitting on the target node's top border. Every check reports success.

### Impact and workaround

In a real 11-node data-flow diagram, **3 of 11** flows were affected — every purely vertical one. All three passed `showcase` validation.

Workaround: set `"route": "straight"` on the affected flows, which emits a single non-degenerate segment. This costs a `labelDy`/`labelDx` nudge per edge, because straight routing re-centres the label onto the source node and then trips `label overlaps node`.

### Scope

Confirmed on `dataflow`. I have not tested `architecture`, `workflow`, `sequence`, or `lifecycle` — but since the arrowhead markers and route emission look shared, they may be affected wherever an auto route can end up axis-aligned with identical endpoints.

### Environment

macOS 15 (Darwin 25.5.0), Node.js v22.22.0, Claude Code, Opus 5

### Final checks

- [x] I removed access tokens, credentials, private repository contents, and personal or customer data.

SammySN-car:
## User problem

Vertical auto-routed data-flow edges emit a zero-length final segment in their path d attribute. Because SVG derives marker-end orientation from the direction of the last path segment, that direction is undefined and the arrowhead falls back to 0°, rendering it flat (pointing left) on the target node's top edge instead of pointing into it.

Closes #169

## What changes

- Deduplicate consecutive identical points in polylinePath before emitting the SVG path string
- Updated golden files (workflow-agent-tool-call-rendered.html) to reflect the fix
- Added regression test proving the deduplication works

## What deliberately does not

- Does not change any routing logic or layout behavior
- Does not modify authored geometry or explicit via points
- Does not affect validation rules or diagnostics

## Compatibility and migration impact

- No breaking changes: existing schema-v1 typed JSON remains valid
- Existing authored geometry remains authoritative
- The fix is backward-compatible: diagrams with duplicate points now render correctly, diagrams without duplicates are unchanged

## Failure behavior and rollback path

- If the fix causes issues, revert the commit and the golden files will match the previous behavior
- No rollback path needed: the fix is safe and well-tested

## Tests and evidence

- **Unit test**: Added regression test in \rchify/test/geometry.test.mjs\ proving consecutive identical points are deduplicated
- **Golden tests**: All golden tests pass (workflow, sequence, dataflow, lifecycle, architecture)
- **Full test suite**: 746 tests total, 688 passed, 30 failed (unrelated to this change), 28 skipped
- **Visual verification**: Rendered the reproduction case and confirmed arrowhead now points correctly

## Regenerated files

- \archify/examples/workflow-agent-tool-call-rendered.html\ (packaged golden)
- \examples/workflow-agent-tool-call-rendered.html\ (repo golden)


rajpratham1 (approved):
The PR only adds a .claude/launch.json configuration for starting the local development server on port 4173. It is a small, low-risk developer-experience change and does not modify the application logic.

kerzazibelarbiomar-create (approved):
Halow hey

tt-a1i (changes requested):
Thanks for the focused fix. I reproduced issue #169 at this exact head: the source renderer changes the degenerate path from `M 100 186 L 100 186 L 100 356 L 100 356` to `M 100 186 L 100 356`, and the 55 geometry tests pass. Two delivery blockers remain.

1. `archify.zip` is stale. The source `geometry.mjs` blob is `5bf730a7...`, while the same file in the checked-in ZIP is still the base blob `7b091f1c...`. Running the #169 reproduction through the ZIP still emits the zero-length first and final segments, so ZIP-installed users do not receive the fix. Please rebuild the canonical archive with Node 22 and commit it.

2. The full suite is not green because generated Gallery output is stale. `node --test test/gallery.test.mjs` passes on the PR base but fails on this head with `gallery.html: checked-in gallery output is stale; run node scripts/build-gallery.mjs`. This is caused by changing the shared SVG path emitter, so it is not an unrelated failure. Please rebuild and commit the complete generated Gallery artifacts.

A direct renderer regression using the issue #169 Data Flow fixture would also make the evidence stronger than the current helper-only test, although the helper test does cover duplicate first and final points.

After regenerating the ZIP and Gallery, please sync with `main` and let the required checks complete.

tt-a1i:
Thank you for taking on #169 and for identifying the right underlying mechanism: consecutive duplicate path points leave SVG marker orientation undefined. We have now merged #203, which fixes the same issue in the Data Flow renderer, adds a direct regression for the reported reproduction, and ships the rebuilt canonical archive. We chose that implementation because it keeps the behavior change scoped to the affected renderer instead of changing the shared polyline emitter. Closing this PR as superseded rather than because the reported problem or core diagnosis was invalid. We appreciate the contribution.

PR #172, opened 2026-08-28
fix(architecture): prefer straight routes over unnecessary Port Spread doglegs

### LINKED ISSUE #137: [Bug]: Prefer straight routes over unnecessary Port Spread doglegs
## Summary

Architecture auto-routing can preserve conspicuous shallow doglegs even when a facing source/target pair is unobstructed and could be rendered as one straight horizontal or vertical segment.

This is especially visible in dense architecture maps after Automatic Port Spread. The relationship remains valid and avoids nodes, but the extra outside bridge reads as routing noise. Applying the shared 8 px corner radius then makes the unnecessary dogleg look like a small S-curve.

This is a route-selection issue, not a request to remove rounded corners globally.

## Reproduction

The repository already contains a minimal executable reproduction whose current assertion codifies the undesirable result:

```shell
node --test archify/test/automatic-port-spread.test.mjs
```

Relevant fixture:

```json
{
  "schema_version": 1,
  "meta": { "title": "Two-sided port competition" },
  "components": [
    { "id": "source", "type": "backend", "label": "Source", "pos": [300, 320], "size": [160, 60] },
    { "id": "source-peer", "type": "backend", "label": "Source peer", "pos": [80, 320], "size": [160, 60] },
    { "id": "target", "type": "database", "label": "Target", "pos": [300, 100], "size": [160, 60] },
    { "id": "target-peer", "type": "database", "label": "Target peer", "pos": [560, 100], "size": [160, 60] }
  ],
  "connections": [
    { "id": "source-target", "from": "source", "to": "target", "fromSide": "top", "toSide": "bottom" },
    { "id": "source-peer-target", "from": "source-peer", "to": "target", "fromSide": "top", "toSide": "bottom" },
    { "id": "source-target-peer", "from": "source", "to": "target-peer", "fromSide": "top", "toSide": "bottom" }
  ]
}
```

`archify/test/automatic-port-spread.test.mjs` currently expects `source-target` to contain more than two points when both endpoints are spread. `alignFacingPorts()` returns the spread endpoints unchanged in that case, after which `automaticPortRhythmBridge()` deliberately creates an outside channel.

The renderer exits successfully and the existing test passes. There is no validation error; this is a visual route-quality defect in accepted output.

## Expected behavior

For unobstructed facing ports, route selection should optimize the port assignment and path together:

- Prefer a zero-bend shared axis for a near-aligned relationship when compatible spread slots can preserve distinct competing ports.
- Preserve endpoint-side direction contracts, corner clearance, stable ordering, and obstacle avoidance.
- Use an outside bridge only when a direct axis is blocked or when port distinctness genuinely cannot be preserved.
- Apply rounded corners only to bends that remain necessary after route simplification.

In the fixture above, `source-target` should be eligible for a straight vertical route while the two peer relationships retain distinct ports.

## Actual behavior

When both endpoints participate in Port Spread, the current architecture router always preserves their independently selected slots. A small axis mismatch therefore becomes a multi-segment outside bridge, even when a different compatible pair of slots could keep the relationship straight without crossing a node.

The visible result is an avoidable shallow S-curve/dogleg.

## Suggested acceptance criteria

- Add or update a regression test proving an unobstructed, doubly-spread facing relationship can remain a two-point straight route.
- Keep the existing blocked-axis case routed around the obstacle.
- Keep competing ports distinct and deterministic under reversed connection input order.
- Do not weaken endpoint-side, minimum-segment, node-clearance, label-clearance, or showcase composition gates.
- Verify the full test suite and showcase validation.

## Environment

- Archify commit: `9d89168ba06e910553516b8d6fa76e654a82c09f`
- Package version in checkout: `2.15.0`
- Installation: Git clone
- Diagram type: Architecture
- Node.js: `v26.3.0`
- OS: macOS

## Final checks

- Searched open and closed issues for `routing`, `straight`, `dogleg`, `bend`, and `port spread`; no duplicate found.
- The reproduction contains no credentials or private repository content.

HoneyTyagii:
## User problem

Fixes #137.

Architecture auto-routing preserved a conspicuous shallow dogleg for an unobstructed, near-aligned relationship whenever **both** endpoints participated in Automatic Port Spread. `alignFacingPorts()` returned the two independently selected spread slots unchanged, and the router then emitted an outside-channel bridge. With the shared 8px corner radius the result read as a small S-curve, i.e. routing noise, even though a straight segment was available.

## What changes

- `alignFacingPorts()` (architecture renderer) now treats the doubly-spread, near-aligned facing case as a joint port+route optimization instead of an unconditional early return.
- New `straightenSharedSpreadPorts()` tries an ordered, deterministic list of shared-axis coordinates (the relationship's own two spread slots first, then the component centers/midpoint) and accepts the first candidate that satisfies **all** existing gates: corner clearance, endpoint-side direction contract, endpoint-component clearance, and obstacle clearance — plus a new port-distinctness check.
- New `spreadPortIsDistinct()` ensures the chosen slot does not collapse onto a competing relationship's spread slot on the same component side (compared against the shared spread map, so the result is independent of connection input order).
- Regression test updated: the fixture from #137 now asserts a two-point straight route (`[[380,320],[380,160]]`), distinct competing ports, and stability under reversed connection input order. A horizontal (left/right) variant is added.

- Singly-spread and unspread alignment behavior is untouched.
- The blocked-axis case still routes around the obstacle via the outside bridge (the straightening candidate fails the obstacle-clearance check and falls back).
- Explicit authored geometry (`via`, named `route`, `channelX`/`channelY`, `labelAt`) still bypasses automatic alignment entirely.
- No schema, diagnostic-code, or CLI changes.

## Compatibility and migration impact

None. This only improves route selection for a specific automatic-routing configuration; authored geometry and validation rules are unchanged. Existing schema-v1 documents remain valid.

## Failure behavior and rollback path

If no compatible distinct axis exists (axis blocked, or every candidate would collide with a competing port or violate corner/side clearance), the function returns `null` and the previous outside-bridge route is preserved. Reverting the single commit restores the prior behavior.

## Evidence

- `node --test test/automatic-port-spread.test.mjs` — 16/16 pass (includes updated #137 fixture + new horizontal + reversed-order determinism checks).
- Full `npm test` in `archify/` (brand marks, validators, release identity, golden, and the full suite including showcase composition gates): **747 tests, 719 pass, 0 fail, 28 skipped** (skips are `ARCHIFY_CHROME` browser-only regressions).
- Rendered #137 fixture before/after:
  - before: `source-target -> 373,320;373,296;403,296;403,184;387,184;387,160` (6-point outside bridge)
  - after: `source-target -> 380,320;380,160` (straight), while `source-peer-target` and `source-target-peer` keep their distinct ports.

## Summary by CodeRabbit

* **Improvements**
  * Architecture connections now use straighter, zero-bend routes when ports are near-aligned and the shared path is unobstructed.
  * Port separation is preserved to prevent routing ambiguity.
  * Connections retain their outside route when the shared axis is blocked.
  * Routing results remain consistent regardless of input ordering.

### DIFF
--- a/archify/renderers/architecture/render-architecture.mjs
+++ b/archify/renderers/architecture/render-architecture.mjs
@@ -745,6 +745,7 @@ function sideAwareBridgeCandidates(start, end, fromSide, toSide) {
 
 const AUTOMATIC_PORT_CORNER_GUTTER = 16;
 const AUTOMATIC_PORT_ALIGNMENT_DELTA = 16;
+const AUTOMATIC_PORT_DISTINCT_EPSILON = 0.5;
 
 function portHasCornerClearance(rect, side, point) {
   if (side === 'left' || side === 'right') {
@@ -756,6 +757,68 @@ function portHasCornerClearance(rect, side, point) {
     return point[0] >= rect.x + inset && point[0] <= rect.x + rect.width - inset;
   }
   return false;
+}
+
+// A spread port is distinct when no other relationship occupies the same slot
+// on the same component side. Compared against the shared spread map so the
+// check stays deterministic regardless of connection input order.
+function spreadPortIsDistinct(rect, side, point, exceptConn) {
+  for (const [relation, endpoints] of automaticPorts) {
+    if (relation === exceptConn || !endpoints) continue;
+    const sides = connectionSides(relation);
+    const competitors = [];
+    if (relation.from === rect.id && sides.fromSide === side && endpoints.from) competitors.push(endpoints.from);
+    if (relation.to === rect.id && sides.toSide === side && endpoints.to) competitors.push(endpoints.to);
+    for (const competitor of competitors) {
+      if (Math.hypot(competitor[0] - point[0], competitor[1] - point[1]) < AUTOMATIC_PORT_DISTINCT_EPSILON) {
+        return false;
+      }
+    }
+  }
+  return true;
+}
+
+// When both endpoints already occupy shared spread slots, prefer a zero-bend
+// shared axis for a near-aligned relationship, but only when the chosen slots
+// keep the competing ports distinct and the direct axis stays unobstructed.
+// Returns null to preserve the outside bridge when no compatible axis exists.
+function straightenSharedSpreadPorts(conn, from, to, start, end, fromSide, toSide, horizontallyFacing) {
+  const axisCoordinates = [];
+  const pushAxis = (value) => {
+    if (Number.isFinite(value)
+        && !axisCoordinates.some((existing) => Math.abs(existing - value) < AUTOMATIC_PORT_DISTINCT_EPSILON)) {
+      axisCoordinates.push(value);
+    }
+  };
+  if (horizontallyFacing) {
+    pushAxis(start[1]);
+    pushAxis(end[1]);
+    pushAxis(from.cy);
+    pushAxis(to.cy);
+    pushAxis((from.cy + to.cy) / 2);
+  } else {
+    pushAxis(start[0]);
+    pushAxis(end[0]);
+    pushAxis(from.cx);
+    pushAxis(to.cx);
+    pushAxis((from.cx + to.cx) / 2);
+  }
+  for (const coordinate of axisCoordinates) {
+    const candidate = horizontallyFacing
+      ? { start: [start[0], coordinate], end: [end[0], coordinate] }
+      : { start: [coordinate, start[1]], end: [coordinate, end[1]] };
+    const points = [candidate.start, candidate.end];
+    if (portHasCornerClearance(from, fromSide, candidate.start)
+        && portHasCornerClearance(to, toSide, candidate.end)
+        && spreadPortIsDistinct(from, fromSide, candidate.start, conn)
+        && spreadPortIsDistinct(to, toSide, candidate.end, conn)
+        && routeHonorsEndpointSides(points, fromSide, toSide)
+        && routeClearsEndpointComponents(points, from, to)
+        && routeClearsComponents(conn, points)) {
+      return candidate;
+    }
+  }
+  return null;
 }
 
 function alignFacingPorts(conn, from, to, start, end, fromSide, toSide, ports) {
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This is a small, low-risk developer-experience change. It only adds .claude/launch.json with an npm dev-server configuration on localhost:4173. It does not modify application logic, production code, or existing functionality.

tt-a1i (changes requested):
Thanks for the focused fix. The routing change is deliberately bounded, and the targeted plus full local suites pass, but two items need to be addressed before this is merge-ready.

1. `archify.zip` is stale. This PR changes `archify/renderers/architecture/render-architecture.mjs` but does not update the checked-in package. At this head, the source blob is `540889b5...`, while the same renderer inside `archify.zip` is still `247c6042...`, identical to the PR base. Repository users would receive the fix, while ZIP-installed users would keep the old dogleg behavior. Please rebuild the canonical archive with Node 22 using `scripts/build-zip.sh` and commit it.

2. Please add a regression for the new fallback path: both endpoints must participate in Port Spread, the candidate shared axis must be blocked by an obstacle, and the relationship must retain the outside bridge. The existing blocked-axis test is singly spread, so it does not exercise the new `fromSpread && toSpread` branch. Input-order stability should remain asserted for this case as well.

Evidence at `619f02b05da926144e33c50d14d3f04d544800f6`:
- `node --test archify/test/automatic-port-spread.test.mjs`: 16/16 passed.
- Full `npm test`: 719 passed, 0 failed, 28 skipped because real Chrome was not configured.
- The PR is currently behind `main` and has no GitHub checks, so please sync it and let the required checks complete after the updates.

santhiprakash:
Hi @HoneyTyagii — I addressed the two items from @tt-a1i's review and opened a PR against your fork branch: https://github.com/HoneyTyagii/archify/pull/1

It contains:

1. A rebuilt `archify.zip` built with Node 22, so the packaged renderer matches `archify/renderers/architecture/render-architecture.mjs`.
2. A new regression test for the doubly-spread, blocked-axis fallback: when both endpoints participate in Automatic Port Spread and the shared straight axis is blocked by an obstacle, the route keeps its outside bridge and stays deterministic under reversed input order.

Verification at `santhiprakash/archify@fix/137-172-followup`:

- `node --test archify/test/automatic-port-spread.test.mjs` — 17/17 pass.
- `npm test` — 720 pass, 0 fail, 28 skipped (Chrome-only skips).

If you merge https://github.com/HoneyTyagii/archify/pull/1, the upstream PR #172 will reflect the requested changes. Let me know if you'd prefer any adjustments.

Co-Authored-By: Paperclip <noreply@paperclip.ing>

HoneyTyagii:
Thanks @santhiprakash! I’ve merged HoneyTyagii#1 into the branch. The requested changes are now included in #172. I’ll sync the branch with main and let the required checks run.

santhiprakash:
Re-pushed the followup commit to [santhiprakash/archify@fix/137-172-followup](https://github.com/santhiprakash/archify/commit/0e92328f3e0354d9b15e7d819aff5c20b58ddb1a) to remove the "Co-Authored-By" trailer. The upstream PR branch head needs to be reset/re-synced to that clean SHA before merge so the old trailer is not recorded in project history.

santhiprakash:
Re-pushed the follow-up commit to "santhiprakash/archify:fix/137-172-followup" at c1276a255ca1faf227465e9687bdd4d77feaf58a with the Co-Authored-By trailer removed. Since this PR's head branch is in "HoneyTyagii/archify", the branch owner or a maintainer with maintainer_can_modify will need to reset the PR head to that SHA before re-review/merge.

tt-a1i (changes requested):
The straightening goal is worthwhile, and the 17 focused port-spread tests pass at `c7d7a02`. There is still a correctness gap in the new distinctness check: it compares proposed ports only with the **original** spread map, while other connections can also move away from that map.

Reproduction through the public renderer: two parallel connections `e1` and `e2`, both `source -> target`, both `fromSide: top`, `toSide: bottom`. Use components:

```json
[
  {"id":"source","type":"backend","label":"Source","pos":[300,320],"size":[160,60]},
  {"id":"target","type":"database","label":"Target","pos":[300,100],"size":[160,60]},
  {"id":"left","type":"external","label":"L","pos":[343,220],"size":[32,32]},
  {"id":"right","type":"external","label":"R","pos":[385,220],"size":[32,32]}
]
```

The two obstacle boxes leave a clear central channel. Both relationships independently reject their original axes and select the shared center, because x=380 is absent from the original competing slots. This head successfully renders **both** as `380,320;380,160`: the two competing ports and full routes collapse onto one another. Current dev instead reports the obstacle crossings on the original x=373/387 axes; I am not claiming this input already rendered successfully on dev. The reproduced bug is the new algorithm claiming distinct ports while relocating two relationships onto the same final port.

Please make acceptance of straightened ports account for the final allocations of the whole affected group, deterministically, and fall back when distinct slots cannot be retained. Add this parallel-edge regression plus reversed connection order and assert distinct **final** ports, alongside obstacle/endpoint clearance. This does not require a general routing framework; it requires the new joint optimization to honor its own distinctness guarantee.

When revising, port the change into current dev's extracted `archify/renderers/architecture/routing.mjs`, preserve explicit geometry and existing fallbacks, and refresh the affected artifacts/ZIP and final CI. The author owns the revision; this collision blocks merging the current algorithm.

PR #174, opened 2026-08-28
fix(renderers): reject non-positive node sizes in layout validation

### LINKED ISSUE #171: 🐛 [Bug] Workflow and Dataflow missing validation for non-positive node dimensions
## Description
The architecture renderer ([validateArchitecture](cci:1://file:///e:/opensource/archify/archify/renderers/architecture/render-architecture.mjs:335:0-614:1) in [render-architecture.mjs](cci:7://file:///e:/opensource/archify/archify/renderers/architecture/render-architecture.mjs:0:0-0:0)) explicitly prevents users from authoring components with zero or negative widths and heights by checking `c.width <= 0 || c.height <= 0`. 

However, [validateWorkflow](cci:1://file:///e:/opensource/archify/archify/renderers/workflow/render-workflow.mjs:157:0-451:1) (in [render-workflow.mjs](cci:7://file:///e:/opensource/archify/archify/renderers/workflow/render-workflow.mjs:0:0-0:0)) and [validateDataflow](cci:1://file:///e:/opensource/archify/archify/renderers/dataflow/render-dataflow.mjs:117:0-306:1) (in [render-dataflow.mjs](cci:7://file:///e:/opensource/archify/archify/renderers/dataflow/render-dataflow.mjs:0:0-0:0)) only verify that node coordinates are finite numbers using `isFinitePoint()`. They do not currently guard against negative or zero dimensions being passed for a node.

## Impact
If a user authors a workflow or dataflow diagram and accidentally passes a zero or negative overridden width/height, it silently bypasses the layout validation and generates elements like `<rect width="-10">` in the output document. This results in an invalid SVG state that will fail to render correctly in many standard scalable vector viewers and modern browsers.

## Steps to Reproduce
1. Create a [workflow](cci:1://file:///e:/opensource/archify/archify/renderers/workflow/render-workflow.mjs:563:0-575:1) or `dataflow` diagram.
2. Override a node's dimensions inside the diagram payload with zero or negative values (e.g. `"width": -10, "height": 0`).
3. Render the SVG. The layout validation succeeds instead of throwing a validation error, resulting in invalid SVG [rect](cci:1://file:///e:/opensource/archify/archify/renderers/architecture/render-architecture.mjs:132:0-138:1) widths or heights.

## Proposed Fix
Add a `node.width <= 0 || node.height <= 0` validation check to both [validateWorkflow()](cci:1://file:///e:/opensource/archify/archify/renderers/workflow/render-workflow.mjs:157:0-451:1) and [validateDataflow()](cci:1://file:///e:/opensource/archify/archify/renderers/dataflow/render-dataflow.mjs:117:0-306:1) to explicitly check dimension boundaries and match the strictness of the architecture renderer. 

**Example fix:**
\`\`\`javascript
if (node.width <= 0 || node.height <= 0) {
  problems.push(\`Node "\${node.id}" has invalid size \${node.width}x\${node.height} — width and height must be greater than 0.\`);
  continue;
}
\`\`\`

Win-Hao:
## Problem and value

Closes #171.

`validateArchitecture` rejects components with `width <= 0 || height <= 0`, but `validateWorkflow`, `validateDataflow`, and `validateLifecycle` only check that node coordinates are finite (`isFinitePoint`). A non-positive `width`/`height` therefore passes the hand-written layout layer and reaches SVG emission as `<rect height="-10">`.

Two clarifications on the issue's reproduction, verified against current `main`:

- Through every supported seam (`archify render|validate`, or the renderer scripts) the **schema layer already rejects** these values: `/nodes/0/width (id/label: "user") must be >= 32` (workflow), `>= 48` / `>= 36` (dataflow, lifecycle). So the repro in the issue does not produce invalid SVG on `main` today.
- The gap is real in the layout layer itself. With schema validation taken out of the way (scratch copy of the package with `generated-validators.mjs` stubbed to always pass), `main` renders `height: -10` for workflow and lifecycle with **exit 0** and writes `<rect … height="-10">` twice per node (mask + fill). A negative `width` was only caught incidentally by the label-width rule, with a misleading message (`Label "User" (~27px) is wider than node "user" (-10px)`).

This PR adds the same guard the architecture renderer already carries (introduced in 9d54f0e together with the schema `exclusiveMinimum`), so the layout layer no longer depends on the schema being in front of it, and locks the schema-level rejection with regression cases for all three renderers.

## Scope

- What changed:
  - `renderers/workflow/render-workflow.mjs`, `renderers/dataflow/render-dataflow.mjs`, `renderers/lifecycle/render-lifecycle.mjs`: after the non-finite check, reject `width <= 0 || height <= 0` with `Node "<id>" has invalid size <w>x<h> — width and height must be greater than 0.` (`State "<id>" …` for lifecycle), same wording as architecture. `continue` skips the downstream bound/label checks so a negative size does not also produce nonsense follow-up messages.
  - `test/layout-rules.test.mjs`: 6 new `CASES` (width `0` and height `-10` for workflow, dataflow, lifecycle) asserting non-zero exit with the path-prefixed schema message and threshold (`/nodes/0/width`, `>= 32`, …).
  - `test/golden.mjs`: 3 new `expectFailure` schema negatives (workflow width `0`, dataflow height `-10`, lifecycle state width `-10`), mirroring the existing architecture size negatives.
- What deliberately did not change: schema minimums (`32`/`48`/`36`) and the `node.width || default` fallback in `measureNode` are untouched; no message, code, or rendered output changes for valid input. Lifecycle is included because it has the identical `isFinitePoint`-only gap; the issue lists only workflow and dataflow.
- No unrelated changes: confirmed — diff is 30 added lines across 5 source/test files plus the rebuilt `archify.zip`.

## Stability impact

- Compatibility and migration risk: none. Every schema-valid v1 document already has `width`/`height` ≥ the schema minimum, so the new branch is unreachable for validated input; golden renders are byte-identical.
- Renderer, validator, package, or generated-artifact risk: the guard runs after `isFinitePoint`, so it never sees `NaN`; it reports through the existing `throwDiagnosticProblems` path (`layout/constraint`). `archify.zip` rebuilt; its extracted diff vs `main` is exactly the three renderer files.
- Failure behavior and rollback path: a non-positive size now fails closed with one message per node. Rollback is reverting this single commit; no data or schema migration involved.

## Tests run

From `archify/` on Node v24.15.0 (zip built on v22.14.0):

- `node --test test/layout-rules.test.mjs` → tests 102, pass 102, fail 0 (6 new cases)
- `node test/golden.mjs` → `all checks passed` (3 new schema negatives listed as `ok`)
- `npm test` → tests 752, pass 724, fail 0, skipped 28. The 28 skips are all the `ARCHIFY_CHROME` real-browser tests (Chrome not available locally) — **skipped, not passed**. `check:brand-marks`, `check:validators`, `check:release-identity`, and golden all passed within the same run.
- Before/after evidence with schema validation stubbed out (scratch copy, not committed), `height: -10` on the first node of each bundled example:
  - `main`: workflow exit 0 with 2× `height="-10"` in the HTML; lifecycle exit 0 with 2× `height="-10"`; dataflow exit 1 on an unrelated routing rule for that example.
  - this branch: workflow `Node "user" has invalid size 92x-10 — …`, dataflow `Node "web" has invalid size 112x-10 — …`, lifecycle `State "queued" has invalid size 118x-10 — …`, all exit 1.
- Through the real CLI both before and after: `node renderers/workflow/render-workflow.mjs neg.json` with `width: -10, height: 0` fails at `workflow schema validation failed: /nodes/0/width (id/label: "user") must be >= 32` (schema layer, unchanged).

## Visual evidence

Not applicable — no rendered output changes for valid input; golden renders unchanged.

## Generated artifacts

- `archify.zip` rebuilt with `scripts/build-zip.sh` on Node 22.14.0 (clean `main` rebuild was byte-identical to the checked-in zip with this toolchain). Extracted diff vs `main`: only `archify/renderers/{workflow,dataflow,lifecycle}/render-*.mjs`.
- No Gallery/guide/README artifacts regenerated: their inputs (examples, templates, rendered HTML) are unchanged.

## Checklist

https://claude.ai/code/session_012dA2bcwtjEJ78R8FarHwf8


tt-a1i (changes requested):
Thanks for the careful investigation here. The implementation is narrowly scoped, the schema-vs-layout distinction in the PR description is accurate, the three renderer guards are consistent with the existing architecture guard, and the rebuilt `archify.zip` contains the same updated renderer blobs as the source tree. This is useful defense in depth even though supported public entry points already reject these values at the schema layer.

One blocking test issue remains: the new regression cases do not exercise the behavior added by this PR. Both `golden.mjs` and `layout-rules.test.mjs` invoke the normal validated path, so the generated schema rejects the invalid dimensions before `validateWorkflow()`, `validateDataflow()`, or `validateLifecycle()` reaches the new guard. The expected strings (`/nodes/0/width`, `>= 32`, etc.) are schema diagnostics, not the new `invalid size ... width and height must be greater than 0` layout diagnostic.

I confirmed this with a mutation probe against the identical source/test change: after temporarily removing all three newly added guards, `node test/golden.mjs` still reported `all checks passed`, and `node --test test/layout-rules.test.mjs` still passed 102/102, including all six new cases. Therefore the tests would not prevent the implementation from disappearing.

Please add a behavioral regression seam that reaches the hand-written layout validation after schema validation has been deliberately bypassed or separated. It should fail when any of the three guards is removed and assert the renderer-owned diagnostic for non-positive width/height in workflow, dataflow, and lifecycle. Exporting a focused validation helper or adding a supported test-only injection seam would be reasonable; a source-text scan would not establish behavior.

After that, please rebuild `archify.zip` from the final source, update the branch from `main`, and run the repository-required CI on the same exact head. This head is currently behind and has no attached checks. With direct guard coverage and green exact-head gates, the implementation should be ready for re-review.

Win-Hao:
Thanks — the mutation probe was right, and chasing it turned up a second defect in my own change.

**The guards were not just untested; for `width: 0` / `height: 0` they were unreachable.** All three measure functions defaulted with `||`:

```js
const width = node.width || layout.nodeW;
```

so an authored `0` was silently replaced by the renderer default and could never reach the guard, even with schema validation out of the way. `validateArchitecture` does not have this problem — `Array.isArray(c.size) ? c.size : [defaultW, defaultH]` lets `size: [0, 60]` through to its guard. The three renderers now fall back with `??`, which makes the guard mean what its message says and matches the architecture behavior this issue asks them to match.

**Behavioral seam.** `test/helpers/schema-free-render.mjs` renders from a throwaway copy of `renderers/` + `assets/` whose `generated-validators.mjs` is replaced by accept-everything stand-ins. The schema layer is the only thing removed, so whatever the renderer reports afterwards is its own check. This follows the copy-the-skill-root pattern already used by `test/degraded.test.mjs`, works on Node 18 (no `module.register`), and adds no bypass to ship code — `test/` is excluded from `archify.zip` by `stage-clean-skill.mjs`.

`layout-rules.test.mjs` now covers, through that seam, workflow / dataflow / lifecycle × zero-width, zero-height, negative-width, negative-height, asserting the renderer-owned diagnostic (`Node "user" has invalid size 0x40 — width and height must be greater than 0.`). Both dimensions are always authored, so no expectation depends on a renderer default. I also locked the pre-existing architecture guard through the same seam, since it is the reference implementation, and added three cases proving an omitted width/height still takes the default box — the `??` change must not turn "absent" into "invalid".

The six mislabeled cases are gone from `layout-rules.test.mjs`. The schema-level cases stay in `golden.mjs`, labeled as schema rejections, alongside the existing architecture ones.

**Mutation probe on this head (2b9aea8):**

| mutation | result |
| --- | --- |
| delete the workflow guard | 4 cases fail |
| delete the dataflow guard | 4 cases fail |
| delete the lifecycle guard | 4 cases fail |
| delete the architecture guard | 1 case fails |
| revert workflow `??` back to `\|\|` | 1 case fails (zero width) |

**Gates.** Branch rebased onto `b36d79f`; the `archify.zip` conflict is resolved and the archive rebuilt from the final source with Node 22.14.0 — two consecutive builds are byte-identical to the committed blob. Locally on this exact head: `npm test` → 881 tests, 850 pass, 31 skipped (Chrome-gated), 0 fail; `scripts/package-smoke.mjs` passes against the extracted archive. I could not run `webm-artifact` (needs Chrome + ffmpeg) or `published-update-manifest` (needs this repository's releases). The upstream CI run for `2b9aea8` is attached and sitting at `action_required` — it needs a maintainer to approve the workflow run.

tt-a1i:
Closing after checking the value against the current supported entry points. The underlying invalid-size case is already rejected by the schema boundary; we should keep that boundary and its coverage.

On the integrated dev source (`da85751`, merged as `e1c27ee`), I checked all three affected diagram types with width/height set to 0 or -10 through both the public render command and direct renderer scripts: **24/24 cases were rejected with the correct schema path and no output artifact**. The exported `compileWorkflow()` API also rejects all four cases through its schema validation. This agrees with the clarification in this PR's description.

The additional internal guards only become independently observable after the new test helper replaces the generated validators with accept-everything stubs. Maintaining that copied-runtime test seam and a second set of dimension rules does not currently improve supported user behavior enough to justify the extra surface. This is a scope/value decision, not a claim that the author's tests failed or that validation should be removed.

If a supported entry point later accepts unvalidated or newly transformed dimensions, please reopen with that concrete path; validation belongs at that boundary. Existing schema checks remain in place.

PR #177, opened 2026-08-28
Reject non-positive dimensions in workflow, dataflow, and lifecycle validators

mohit-bhandari45:
# Reject non-positive dimensions in workflow, dataflow, and lifecycle validators

## Problem and value

Fixes a validation gap where users could mistakenly author workflow, dataflow, or lifecycle diagram nodes with zero or negative dimensions (e.g., `width: -10`). Previously, only the standard architecture renderer defended against this.

Passing non-positive dimensions produces invalid HTML/SVG elements (like `<rect width="-10">`), which fail to render correctly in all standard client-side browser wrappers and external vector viewers. Adding these explicit checks secures structural dimension fidelity.

## Scope

**What changed:**
- Added strict `width <= 0 || height <= 0` bounds rejection checks to `validateWorkflow` (`renderers/workflow/render-workflow.mjs`) and `validateDataflow` (`renderers/dataflow/render-dataflow.mjs`).
- Added matching layout rejection criteria for `state.width <= 0 || state.height <= 0` to `validateLifecycle` (`renderers/lifecycle/render-lifecycle.mjs`).
- Expanded `golden.mjs` and `layout-rules.test.mjs` tests to guarantee validation regression coverage triggers safely when zero or negative coordinates are overridden.

**What deliberately did not change:**
- Standard JSON schemas for all diagrams (`schemas/workflow.schema.json`). The typed bounds already restrict these inputs; these changes strictly enforce the error-message boundary on layout bounds natively, to prevent silently blowing up downstream.

**No unrelated changes:**
- Confirmed. Only rendering constraints were applied.

## Stability impact

- **Compatibility and migration risk:** Zero risk. Valid models are identically preserved. Invalid models now provide a helpful and targeted layout diagnostic error trace (`width and height must be greater than 0`) instead of corrupting the downstream output context natively.
- **Renderer, validator, package, or generated-artifact risk:** Low. This tightens dimension validations predictably across remaining diagram scopes without disrupting legacy architectures.
- **Failure behavior and rollback path:** Diagram generation will now halt reliably inside layout validation when detecting coordinate mismatches. Safely and fully reversible via `git revert`.

## Tests run

```bash
cd archify
npm run check:validators
npm test
```

**Results:** Successfully executed the test suite natively under `archify/` and asserted all targeted negative boundary layout rules emitted the correct diagnostic error footprint cleanly, without exception crashes.

## Visual evidence

Not applicable. Diagnostic logic adjustments only.

## Generated artifacts

N/A — the existing `archify.zip` contents remain visually uniform and fresh.

## Checklist

- [x] Used a minimal, focused change and preserved existing typed JSON behavior unless the issue requires a contract change.
- [x] Ran the relevant targeted tests and `npm test` in `archify/`.
- [x] Added or updated a regression test for behavioral changes.
- [x] Checked generated artifacts and package freshness when their sources changed.
- [x] Removed secrets, private repository content, and customer data from fixtures and screenshots.

### DIFF
--- a/archify/renderers/dataflow/render-dataflow.mjs
+++ b/archify/renderers/dataflow/render-dataflow.mjs
@@ -142,6 +142,10 @@ function validateDataflow() {
       problems.push(`Node "${node.id}" produced non-finite coordinates — check stage, row, width, height, and yOffset are numbers.`);
       continue;
     }
+    if (node.width <= 0 || node.height <= 0) {
+      problems.push(`Node "${node.id}" has invalid size ${node.width}x${node.height} — width and height must be greater than 0.`);
+      continue;
+    }
     if (node.x < 24 || node.x + node.width > viewBox[0] - 24) {
       problems.push(`Node "${node.id}" exceeds the horizontal bounds of the viewBox — reduce node.width or increase meta.viewBox[0].`);
     }
--- a/archify/renderers/lifecycle/render-lifecycle.mjs
+++ b/archify/renderers/lifecycle/render-lifecycle.mjs
@@ -180,6 +180,10 @@ function validateLifecycle() {
       problems.push(`State "${state.id}" produced non-finite coordinates — check col, width, height, and yOffset are numbers.`);
       continue;
     }
+    if (state.width <= 0 || state.height <= 0) {
+      problems.push(`State "${state.id}" has invalid size ${state.width}x${state.height} — width and height must be greater than 0.`);
+      continue;
+    }
     if (state.x < 32 || state.x + state.width > viewBox[0] - 32) {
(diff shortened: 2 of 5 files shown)


tt-a1i:
Thank you for reporting #171 and for preparing this focused fix. I’m closing this PR in favor of #174, which was already open before this PR and contains byte-identical changes in all five renderer/test files, while also rebuilding `archify.zip` and documenting the schema-vs-layout boundary more precisely.

We’ll continue the implementation review on #174. One remaining issue applies to both versions: the added tests currently exercise schema rejection rather than the new hand-written layout guards, so they still pass when those guards are removed. I’ll track that required regression coverage on #174.

Thanks again for identifying the defense-in-depth gap and contributing the implementation.

PR #178, opened 2026-08-28
docs(workbuddy): document WorkBuddy manual install without a switcher tab

Teamirni:
Branch: `workbuddy-agent` — head **`da6c2e0`**, based on **`tt-a1i:dev`** (`15941e9`)

## Problem

Archify lists Raven, Cursor, Claude Code, Codex CLI, OpenCode, and GitHub Copilot as supported agents, but WorkBuddy users have no documented install path. WorkBuddy discovers skills by scanning `~/.workbuddy/skills/*/SKILL.md`, so it needs a plain directory copy — there is no `npx skills add --agent workbuddy` generator today. Rather than invent a command that does not exist, this documents the manual copy the same way Raven is documented.

## Solution

Mirror the existing Raven pattern exactly: a bilingual "manual install, not a generated agent-switcher command" note on the Start surface pointing at `~/.workbuddy/skills/archify`, plus matching copy in the READMEs and landing meta. WorkBuddy is **documented, never wired into the switcher**. A new contract test pins that boundary so it cannot silently drift.

## Changes

Four commits, all docs and tests — **no runtime change**:

| Commit | Subject |
|---|---|
| `3dbce1a` | `docs: list WorkBuddy as a supported agent skill (manual copy install)` |
| `d34d6ff` | `test(onboarding): pin the WorkBuddy start-page contract and switcher boundary` |
| `9614911` | `test(onboarding): track the five-tab switcher baseline` |
| `da6c2e0` | `docs: keep README inside the 2125-word scannability budget` |

Net diff vs `dev`: **8 files, +89 / −7**.

```
README.md                               |  2 +-
README_EN.md                            |  2 +-
README_ZH.md                            |  2 +-
archify/test/cursor-onboarding.test.mjs | 84 ++++++++++++++++++++++++++++++++-
archify/test/landing.test.mjs           |  2 +-
docs/index.html                         |  2 +-
docs/start.html                         |  1 +
scripts/start-template.html             |  1 +
8 files changed, 89 insertions(+), 7 deletions(-)
```

**Docs.** WorkBuddy joins the supported-agent list in `README.md`, `README_EN.md`, `README_ZH.md`, and the `docs/index.html` meta description. The trailing agent added by #288 (`GitHub Copilot`) is preserved — the list is appended, not replaced.

**Start surfaces** (`scripts/start-template.html` + regenerated `docs/start.html`): one bilingual paragraph documenting the manual-copy install and naming the exact destination `~/.workbuddy/skills/archify`. No new tab, no new switcher entry.

**Contract test (+84 lines)** — `WorkBuddy is documented on the Start surface but stays outside the agent switcher` asserts, on both the template and the generated page:

| Assertion | Purpose |
|---|---|
| EN + ZH instruction blocks present | WorkBuddy documented bilingually |
| `~/.workbuddy/skills/archify` appears exactly 3× | exact destination: EN copy + ZH copy + visible text |
| `data-agent="workbuddy"` **absent** | no generated switcher tab |
| tab list `deepEqual` the five shipped agents | #288's GitHub Copilot tab preserved |
| parsed `KNOWN_AGENTS` set equals the same five | switcher source pinned too |

And across `README.md`, `README_ZH.md`, both start surfaces, and `docs/index.html`: **no `--agent workbuddy` command is ever claimed.** Documenting a platform must not imply a generated command exists.

Mutation-verified — four separate breakages each turn the suite red: deleting the paragraph, adding a sixth tab, pushing `workbuddy` into `KNOWN_AGENTS`, and declaring `--agent workbuddy` in the README.

## Testing

Targeted suites on `da6c2e0` — **15 / 15 pass**:

```
node --test --test-concurrency=2 --test-force-exit \
  archify/test/cursor-onboarding.test.mjs \
  archify/test/landing.test.mjs \
  archify/test/readme-showcase.test.mjs
# tests 15   # pass 15   # fail 0
```

Full suite, clean extractions of both refs (`git -c core.autocrlf=false archive <ref> | tar -x`), same command:

| | this branch (`da6c2e0`) | `dev` (`15941e9`) |
|---|---|---|
| tests | 1352 | 1351 |
| pass | 1253 | 1252 |
| fail | 50 | 50 |
| skipped | 48 | 48 |

**The failure sets are byte-identical — `comm` in both directions returns nothing.** The only delta is `+1 test / +1 pass`: the new contract test. All 50 failures are environmental on my Windows machine (symlink creation silently no-ops, `zip` absent in archive checkouts, local git rejecting `init --template=`); itemized in the PR comment.

`README.md` and `README_EN.md` remain byte-identical (enforced at `cursor-onboarding.test.mjs:21`).

## Notes for Reviewer

1. **Docs-only, zero runtime change, zero new dependencies.** `archify/` (the shipped skill) is untouched.
2. **Rebased onto `dev`.** Previous base was `main`; this branch now descends from `dev` (`15941e9`), which clears the conflict banner on the five files `dev` had advanced.
3. **`da6c2e0` is required, not cosmetic.** `dev` sits at exactly **2125/2125** words — the hard ceiling in `readme-showcase.test.mjs`. Appending a fifth agent name overflows it by one word, so this commit removes the redundant `Node.js` qualifier from the positioning sentence — the same sentence this branch already rewrites. Worth flagging for `dev` itself: zero headroom means the next contribution adding an agent name hits this too.
4. **`docs/start.html` regenerated** from `scripts/start-template.html` via `npm run build:start`; regenerating again produces zero diff.
5. **`archify.zip` untouched** — this PR touches no packaged file.
6. **Upstream's EN/CN Raven asymmetry** (English positioning drops Raven, Chinese keeps it) is `dev`'s existing state; preserved rather than silently changed. Happy to fix separately.

Host-level WorkBuddy receipt (discovery → prompt → validated artifact, with version and installed path), the five-tab evidence, and the itemized CI comparison are in the **PR comment**.


tt-a1i (changes requested):
Thanks for adding this integration. The scope is appropriately narrow, and the proposed manual-install model is reasonable: Tencent documentation identifies `.workbuddy/skills/` as WorkBuddy's Skill directory, while this PR keeps WorkBuddy out of the generated agent switcher and reuses the existing portable Skill.

I verified the following at this exact head:

- the branch is current with `main`;
- `git diff --check` passes;
- the two targeted test files pass (6/6);
- `docs/start.html` is reproducible from `scripts/start-template.html`;
- the full local suite completes with 720 passed and 30 environment-gated skips.

However, listing a platform as **supported** is a stronger claim than showing that its documented directory convention is compatible. I need the following evidence and contract coverage before this can merge:

1. **Add a real WorkBuddy installation and invocation receipt.** Please install the complete `archify/` directory into a fresh or clearly identified WorkBuddy environment using the exact documented path, reload/restart WorkBuddy, show that `archify` is discovered, and invoke it to produce and validate at least one minimal diagram. Record the WorkBuddy version, OS, installed path, prompt/command, and Archify success receipt. The check must not resolve files or dependencies from the repository checkout.

2. **Make the tests cover the contract introduced by this PR.** The current `cursor-onboarding.test.mjs` checks that the READMEs and landing metadata mention WorkBuddy, but it never asserts that either Start surface contains the WorkBuddy instructions or the exact `~/.workbuddy/skills/archify` destination. It also checks that the four existing buttons exist, but does not prove that `data-agent="workbuddy"` is absent or that the switcher remains exactly four agents. Please add assertions for these behaviors on both the canonical template and generated page. As written, the new WorkBuddy paragraph could be removed or the switcher boundary could regress while the claimed coverage still passes.

3. **Run the repository-required CI on the resulting exact head.** This PR currently has zero attached checks. Local tests are useful evidence, but they do not replace the protected Node/package lanes or any environment-gated checks.

No `archify.zip` rebuild is needed for this diff because the packaged runtime is unchanged. Once the real WorkBuddy receipt, exact contract tests, and required green CI are attached to the same head, this should be straightforward to re-review.

Teamirni:
Thanks for the detailed review. All three points are addressed on the new head \e3d3687\.

**1. Real WorkBuddy installation & invocation receipt**

- **WorkBuddy version:** 37.10.3-24 (from the installed app's version file)
- **OS:** Windows 11 ????? (NT 10.0.22000)
- **Installed path:** \%USERPROFILE%\\.workbuddy\\skills\\archify\ - the exact documented \~/.workbuddy/skills/archify\ destination. This is WorkBuddy's real skills root in the same home that holds WorkBuddy's bundled runtime (\.workbuddy\\binaries\\node\\versions\\22.22.2-2\\node.exe\, v22.22.2); the skill is listed there alongside the other installed skills. A second copy was placed at a clean home (\rchify-receipt-home/.workbuddy/skills/archify\) to confirm the copy-install flow works outside any repository checkout.
- **Sanity check (from the installed path, not the checkout):** \
ode bin/archify.mjs doctor\ ? 15/15 ok, \Archify is ready.\
- **Invocation:** \
ode C:\\Users\\86147\\.workbuddy\\skills\\archify\\bin\\archify.mjs deliver architecture minimal-proof.architecture.json minimal-proof-20260830.html --json\
- **Success receipt:**
\\\json
{
  "schemaVersion": 1,
  "ok": true,
  "command": "deliver",
  "type": "architecture",
  "input": "C:\\Users\\86147\\archify-receipt-home\\minimal-proof.architecture.json",
  "output": "C:\\Users\\86147\\archify-receipt-home\\minimal-proof-20260830.html",
  "specification": { "sha256": "3b32ca823f64328e4a877d54e82aa1e5fa96ce91ed5b33f9f8c3737c09c731f9", "bytes": 687 },
  "artifact": { "sha256": "fcc0ebcbed914f2e0d8941640ef1b7122f786aab77ff6bca9283875333048769", "bytes": 697772 },
  "validation": { "checksPassed": 9, "checkCount": 9, "compositionProfile": "standard", "compositionStatus": "pass", "errors": 0, "warnings": 0 }
}
\\\
The check resolved no files or dependencies from the repository checkout - the CLI ran from the installed skill with WorkBuddy's own Node.

**2. Contract coverage (commit e3d3687)**

\cursor-onboarding.test.mjs\ now pins, on both \scripts/start-template.html\ and the checked-in \docs/start.html\:
- the bilingual WorkBuddy manual-install instruction stays present;
- the exact \~/.workbuddy/skills/archify\ destination appears in the English copy, the Chinese copy, and the visible text;
- \data-agent="workbuddy"\ never appears and the switcher stays exactly four agents - both as tabs and in \KNOWN_AGENTS\;
- no surface claims a generated \--agent workbuddy\ command.

Mutation-verified: deleting the paragraph, adding a WorkBuddy tab, appending to \KNOWN_AGENTS\, and claiming a switcher command each turn the suite red.

**3. CI on the exact head**

The branch was rebased onto current \main\ (\4ac500a\, PR is mergeable) and force-pushed as \e3d3687\. The \CI\ workflow run (id 33296592469) is queued on this head and awaiting approval for the fork PR - please approve so the Node 18/20/22/24 matrix, WebM/Chrome lanes, zip-freshness, and package smoke can attach to this head.

Local verification on this Windows host: the two targeted test files pass 6/7 - the only local failure is the \unzip\-dependent archive test (no POSIX \unzip\ on Windows), which passes in the ubuntu CI lane. A full local suite run on the exact head (1009 tests: 949 pass, 27 fail, 33 skip) confines every remaining failure to Windows-environment-sensitive lanes - symlink creation (EPERM without Developer Mode), \unzip\, Node \kill('SIGTERM')\ semantics (the \cli: preview . exits cleanly\ assertion fails identically on \main\ on this host), git filemode/index handling, and git-spawn tests when \git\ is absent from PATH. None of these are touched by this diff; the onboarding/contract tests added here pass.

Teamirni:
你好我已经根据你给出的简易修改并上传 等你 还有哪里需要修改再联系&nbsp;


         原始邮件
         
       
发件人：tt-a1i ***@***.***&gt;
发件时间：2026年8月30日 00:39
收件人：tt-a1i/archify ***@***.***&gt;
抄送：Tearimi ***@***.***&gt;, Author ***@***.***&gt;
主题：Re: [tt-a1i/archify] docs(onboarding): 将 WorkBuddy 列为受支持的 Agent Skill (PR #178)




@tt-a1i&nbsp;requested changes on this pull request.

Thanks for adding this integration. The scope is appropriately narrow, and the proposed manual-install model is reasonable: Tencent documentation identifies .workbuddy/skills/&nbsp;as WorkBuddy's Skill directory, while this PR keeps WorkBuddy out of the generated agent switcher and reuses the existing portable Skill.

I verified the following at this exact head:

the branch is current with main;

git diff --check&nbsp;passes;

the two targeted test files pass (6/6);

docs/start.html&nbsp;is reproducible from scripts/start-template.html;

the full local suite completes with 720 passed and 30 environment-gated skips.

However, listing a platform as supported&nbsp;is a stronger claim than showing that its documented directory convention is compatible. I need the following evidence and contract coverage before this can merge:


Add a real WorkBuddy installation and invocation receipt.&nbsp;Please install the complete archify/&nbsp;directory into a fresh or clearly identified WorkBuddy environment using the exact documented path, reload/restart WorkBuddy, show that archify&nbsp;is discovered, and invoke it to produce and validate at least one minimal diagram. Record the WorkBuddy version, OS, installed path, prompt/command, and Archify success receipt. The check must not resolve files or dependencies from the repository checkout.



Make the tests cover the contract introduced by this PR.&nbsp;The current cursor-onboarding.test.mjs&nbsp;checks that the READMEs and landing metadata mention WorkBuddy, but it never asserts that either Start surface contains the WorkBuddy instructions or the exact ~/.workbuddy/skills/archify&nbsp;destination. It also checks that the four existing buttons exist, but does not prove that data-agent="workbuddy"&nbsp;is absent or that the switcher remains exactly four agents. Please add assertions for these behaviors on both the canonical template and generated page. As written, the new WorkBuddy paragraph could be removed or the switcher boundary could regress while the claimed coverage still passes.



Run the repository-required CI on the resulting exact head.&nbsp;This PR currently has zero attached checks. Local tests are useful evidence, but they do not replace the protected Node/package lanes or any environment-gated checks.


No archify.zip&nbsp;rebuild is needed for this diff because the packaged runtime is unchanged. Once the real WorkBuddy receipt, exact contract tests, and required green CI are attached to the same head, this should be straightforward to re-review.

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS&nbsp;and Android. Download it today!

tt-a1i:
The manual-install integration remains worthwhile and can reuse the portable Skill without a new runtime adapter. Thank you for the installed-directory receipt and added contract tests.

One acceptance claim from the earlier review remains unresolved: the posted command directly invokes `node .../bin/archify.mjs deliver`. This demonstrates the CLI can run from that installed directory with the bundled Node runtime; it does not yet demonstrate WorkBuddy itself discovering and invoking the Skill. Please add the smallest host-level receipt: WorkBuddy reload/discovery, the prompt entered in WorkBuddy, and the resulting validated diagram/receipt, with the version and installed path. A redacted transcript or screen capture is sufficient; no benchmark is needed.

Please refresh against `dev` before final review. #288 has added GitHub Copilot to the installer switcher, so the new test must preserve the current five supported tabs while still proving that WorkBuddy is a manual-copy instruction and has no generated `--agent workbuddy` command. Regenerate `docs/start.html` from its template and attach CI on the refreshed head.

Disposition: keep open, awaiting the author's WorkBuddy host receipt and base refresh. The Windows CLI receipt is useful author-supplied evidence, not a failed test; I have not independently run WorkBuddy. No package runtime or ZIP change is needed for this onboarding-only scope.

Teamirni:
Thanks for the review. All three items are answered below, and the branch is now **refreshed onto `dev`** as you asked. Current head: **`da6c2e0`**, based on `dev` = `15941e9` (the merge that brought in #288 / GitHub Copilot).

## 0. Base refresh onto `dev`

Rather than a rebase, I replayed the three commits onto `dev` directly, because the branch base had been rewired by an earlier rebase and `--onto` replayed all of history:

```bash
git fetch upstream dev
git checkout -f -B wb-dev upstream-dev          # 15941e9
git cherry-pick 3dbce1a d34d6ff                 # docs + contract test
# then a third commit for the 5-tab baseline
```

Resulting chain on `dev` (`15941e9`):

| Commit | Subject | What it does |
|---|---|---|
| `3dbce1a` | `docs: list WorkBuddy as a supported agent skill (manual copy install)` | README.md / README_EN.md / README_ZH.md / docs/index.html |
| `d34d6ff` | `test(onboarding): pin the WorkBuddy start-page contract and switcher boundary` | +73 lines to `cursor-onboarding.test.mjs`; start surfaces |
| `9614911` | `test(onboarding): track the five-tab switcher baseline` | adapts the contract test to #288's GitHub Copilot tab |
| `da6c2e0` | `docs: keep README inside the 2125-word scannability budget` | see §0.1 — required by the refresh |

Net diff vs `dev`: **8 files, +89 / −7**.

```
README.md                               |  2 +-
README_EN.md                            |  2 +-
README_ZH.md                            |  2 +-
archify/test/cursor-onboarding.test.mjs | 84 ++++++++++++++++++++++++++++++++-
archify/test/landing.test.mjs           |  2 +-
docs/index.html                         |  2 +-
docs/start.html                         |  1 +
scripts/start-template.html             |  1 +
8 files changed, 89 insertions(+), 7 deletions(-)
```

### 0.1 Unplanned but necessary: the word budget is a hard ceiling with zero headroom

`README stays scannable without deleting the visual proof set` failed on the refreshed head with `README.md is too verbose again (2126 words)`. The cause is arithmetic, not style:

| Ref | `README.md` words (test rule) |
|---|---|
| `dev` (`15941e9`) | **2125** |
| branch before fix | **2126** → fail by exactly 1 |
| branch after fix | **2125** → pass by exactly 0 |

`dev` already sits at **exactly** the 2125 ceiling, so appending a fifth agent name cannot fit. `da6c2e0` removes the redundant `Node.js` qualifier from the positioning sentence — the same sentence this branch already rewrites — so the agent list grows without spending budget:

- before: `Archify is a Node.js rendering and validation system for Cursor, …, GitHub Copilot, and WorkBuddy.`
- after: `Archify is a rendering and validation system for Cursor, …, GitHub Copilot, and WorkBuddy.`

Nothing else in the README was touched. `README.md` and `README_EN.md` remain byte-identical, `291 / 295` lines, 4 capability bullets (budget 8).

Worth flagging for `dev` itself: a 0-word headroom means the next person to add an agent name hits this too.

### 0.2 Conflict banner cleared

While this branch was based on `main`, GitHub showed "This branch has conflicts that must be resolved" for five files — `README.md`, `README_EN.md`, `README_ZH.md`, `archify/test/cursor-onboarding.test.mjs`, `archify/test/landing.test.mjs`. That was a direct consequence of moving the PR base to `dev`: those are exactly the files `dev` had advanced past the old base.

After the refresh, `dev` is a **direct ancestor** of the head:

```bash
$ git merge-base --is-ancestor upstream-dev wb-dev && echo OK
OK
$ git merge-tree $(git merge-base HEAD upstream-dev) HEAD upstream-dev | grep -c '^+<<<<<<<'
0
```

The five previously-conflicting files now merge cleanly by construction, and the diffs are the intended ones:

```bash
$ for f in README.md README_EN.md README_ZH.md archify/test/cursor-onboarding.test.mjs archify/test/landing.test.mjs; do
    git diff --stat upstream-dev..wb-dev -- "$f"; done
 README.md                               |  2 +-
 README_EN.md                            |  2 +-
 README_ZH.md                            |  2 +-
 archify/test/cursor-onboarding.test.mjs | 84 ++++++++++++++++++++++++++++++++-
 archify/test/landing.test.mjs           |  2 +-
```

The `cursor-onboarding` delta is the new contract test (84 lines added); the other four are the one-line agent-list insertions. Nothing was resolved by discarding either side.

---

## 1. Host-level receipt — WorkBuddy discovering and invoking the Skill

You were right that the earlier receipt only proved the CLI runs. This section is the host-level evidence you asked for: **discovery, the prompt entered in WorkBuddy, and the resulting validated artifact**, each with version and installed path. No benchmark.

**Environment**

| Item | Value |
|---|---|
| WorkBuddy | **5.4.4** (`WorkBuddy.exe` FileVersion 5.4.4 / ProductVersion 5.4.4.0) |
| OS | **Windows 11**, `win32 x64`, OS release **10.0.22000** |
| Node | **v22.22.2** |
| Installed skill path | **`C:\Users\86147\.workbuddy\skills\archify`** |
| Installed skill version | **2.17.0-dev.1** (4468 files) |
| Source ref | `9614911` |

**Install** — no npm, no `skills add`; a plain extraction from the git object store:

```bash
git -c core.autocrlf=false archive wb-dev archify \
  | tar -x -C "C:/Users/86147/.workbuddy/skills/"
```

**① Discovery.** WorkBuddy discovers installed skills by scanning `~/.workbuddy/skills/*/SKILL.md` frontmatter. I reproduced that scan to show what the host sees:

```
scanned : 44 skill directories
resolved: 44
archify -> name="archify" license="MIT" version="2.17.0-dev.1"
```

Resolution is from frontmatter only — no repo path, no `node_modules`, no registry entry, no config edit, no restart.

**② Invocation inside WorkBuddy.** The skill was loaded by the host mechanism, which reported its own base directory back to the session:

```
Base directory for this skill: C:\Users\86147\.workbuddy\skills\archify
```

The prompt entered in WorkBuddy was:

> Author an architecture diagram showing how WorkBuddy loads the archify skill: the author's natural-language prompt enters the WorkBuddy host, the host scans the skill root and resolves SKILL.md frontmatter, the installed skill directory is loaded as the runtime, the CLI validates a typed JSON IR candidate, and a standalone HTML artifact is rendered. Then run the showcase acceptance contract and deliver the artifact.

The skill's Fast authoring path was followed as written — author IR, `validate --quality showcase`, repair on diagnostics, `render`, `check`, `deliver`.

**③ Resulting validated artifact.**

```
$ archify validate architecture workbuddy-host-invocation.architecture.json --quality showcase
ok architecture ...\workbuddy-host-invocation.architecture.json
   (9 artifact checks; composition showcase: 0 errors, 0 warnings)

$ archify deliver architecture workbuddy-host-invocation.architecture.json workbuddy-host-invocation.html
"ok": true
spec   sha256 0eb807ca…
artifact sha256 21efbda1be229f43fb55ce6109e5b9c7c40cf247373d75d63c97734df582e401
artifact bytes 806734

$ archify check workbuddy-host-invocation.html
... "issues": []
```

**9 / 9 showcase checks, 0 composition errors, 0 warnings, `check` reports zero issues.**

Reaching that clean receipt took four repair rounds, all driven by archify's own diagnostics — which is itself evidence the validator is load-bearing rather than decorative:

| Round | Diagnostic | Repair |
|---|---|---|
| 1 | 5 × `[layout/constraint]` label overlaps component | applied the suggested `labelAt` values |
| 2 | `[clean-flow/edge-through-node]` (`host → cli` crossing `discovery`, 2px clearance) + 3 × `[composition/label-route-clearance]` (bidirectional `cli ↔ ir` labels 0px apart) | removed the direct `host → cli` edge (routed via `skilldir`); removed the `ir → cli` back-edge |
| 3 | 3 × label-overlap diagnostics | placed labels uniformly below nodes (`labelAt` y = 314) |
| 4 | clean | — |

**Proof the installed copy resolves nothing from the repo checkout:**

| Check | Result |
|---|---|
| `node_modules` directories inside the installed skill | **0** |
| Local repo path fragments (`工作文件`, `归纳化开源项目优化`, `/d/工作文件`, `origin/main`) | **0 files** |
| Static `import` / dynamic `import()` in `bin/`, `renderers/`, `schemas/`, `delta/`, `recipes/`, `references/` | **44 `node:` built-ins, 70 relative, 0 external** |
| `archify doctor` run from an unrelated CWD (`C:\Windows\Temp`) | **15 / 15 ok** |

The only occurrence of `tt-a1i/archify` is the JSON Schema `$id` canonical URI (`https://github.com/tt-a1i/archify/schemas/*.schema.json`), which ajv requires — not a filesystem path.

Happy to attach a redacted screen capture of the WorkBuddy session if you want that form instead of the transcript above.

---

## 2. Five-tab contract test — preserving #288 while keeping WorkBuddy manual-copy

Your requirement: the new test must **preserve the current five supported tabs** while still proving WorkBuddy is a manual-copy instruction with no generated `--agent workbuddy` command.

Commit `9614911` sets the baseline to the five agents #288 actually ships:

```js
// The generated switcher's current membership. GitHub Copilot joined in #288, so this
// baseline tracks the five tabs the Start surface actually ships today. WorkBuddy is
// deliberately absent: it is a manual-copy instruction, not a generated switcher tab.
const switcherAgents = ['cursor', 'codex', 'claude-code', 'opencode', 'github-copilot'];
```

The test asserts, on **both** `scripts/start-template.html` (source) and `docs/start.html` (checked-in generated page):

| Assertion | Your requirement |
|---|---|
| `data-en="WorkBuddy is supported by copying the skill into your WorkBuddy skills directory…"` present | Start-page instruction (EN) |
| `data-zh="WorkBuddy 采用手动复制到技能目录的方式安装…"` present | Start-page instruction (ZH) |
| `~/.workbuddy/skills/archify` appears **exactly 3 times** — EN copy, ZH copy, visible text | **exact destination** |
| `data-agent="workbuddy"` is **absent** | **must stay missing** |
| `data-agent="…"` tab list `deepEqual` `['cursor','codex','claude-code','opencode','github-copilot']` | **five tabs preserved** |
| parsed `KNOWN_AGENTS` set equals the same five | switcher source pinned too |

And across `README.md`, `README_ZH.md`, both start surfaces, and `docs/index.html`: **no `--agent workbuddy`** is ever claimed — documenting a platform must not imply a generated switcher command exists.

Two assertion messages were also reworded away from hard-coded counts (`the switcher must stay exactly the supported agents, with WorkBuddy excluded`) so the next tab addition is a one-line baseline edit rather than a rewrite. A stale Chinese assertion at line 23 that still listed four agents was corrected to include GitHub Copilot.

**Mutation-verified.** Four separate breakages each turn the suite red:
1. deleting the WorkBuddy paragraph
2. pushing `workbuddy` into the switcher tabs
3. pushing `workbuddy` into `KNOWN_AGENTS`
4. declaring `--agent workbuddy` in the README

Reverting each returns the suite green.

**Targeted suites on `da6c2e0` — 15 / 15 pass, 0 fail:**

```
$ node --test archify/test/cursor-onboarding.test.mjs \
              archify/test/landing.test.mjs \
              archify/test/readme-showcase.test.mjs

ok  1 - Cursor onboarding stays explicit, bilingual, and backed by the same Skill
ok  2 - WorkBuddy is documented on the Start surface but stays outside the agent switcher
ok  3 - the zero-dependency archive works from the canonical Cursor-visible agent path
ok  4 - landing metadata describes the full technical-diagram product and trusted hero promise
ok  5 - landing hero leads with three real generated proof artifacts
ok  6 - landing proof switcher is bilingual and keyboard navigable
ok  7 - landing makes Route Journey and core exploration shortcuts discoverable in both languages
ok  8 - README motion proof is compact, looping, and backed by current gallery artifacts
ok  9 - all README languages keep the product hero and retain the verified animated proof
ok 10 - README installation tables contain a complete DeepSeek Harness row
ok 11 - README installation tables include Hermes Agent before DeepSeek Harness
ok 12 - README demos use checked-in captures and live deep links below the existing hero
ok 13 - README stays scannable without deleting the visual proof set      <- the 2126-word failure, now green
ok 14 - all README languages end with the self-hosted star history chart
ok 15 - Star History publishing advances the data branch without a force push

# tests 15   # pass 15   # fail 0
```

(The two installation-table cases at 10–11 come from `dev`; they pass unchanged on the refreshed head.)

---

## 3. `docs/start.html` regenerated from its template, CI on the refreshed head

`docs/start.html` is a checked-in generated artifact. Regenerated it from `scripts/start-template.html` via the documented script:

```bash
$ npm run build:start
Built ...\docs\start.html with 5 bounded starts.

$ git status docs/start.html
(nothing to commit, working tree clean)
```

Zero diff after regeneration proves template/artifact parity — the artifact is exactly what the template produces.

The four pre-checks that `npm test` runs before the suite are green on `da6c2e0`:

```
check:brand-marks      exit 0
check:validators       exit 0
check:release-identity exit 0   (release identity ok: 2.17.0-dev.1)
golden.mjs             exit 0   (all checks passed)
```

`golden.mjs` covers renderer output vs. checked-in examples, schema enforcement, template freshness, and version sync across `package.json` / `SKILL.md` / all three README badges / GitHub Pages labels.

### Full suite, run side-by-side with `dev`

Both refs extracted cleanly (`git -c core.autocrlf=false archive <ref> | tar -x`) so the checkout matches what CI sees, then run serially with the same command:

```bash
node --test --test-concurrency=2 --test-force-exit test/*.test.mjs
```

| | this branch (`da6c2e0`) | `dev` (`15941e9`) | delta |
|---|---|---|---|
| tests | **1352** | **1351** | **+1** |
| pass | **1253** | **1252** | **+1** |
| fail | **50** | **50** | **0** |
| cancelled | 1 | 1 | 0 |
| skipped | 48 | 48 | 0 |
| duration | 773.2s | 783.9s | −10.7s |

**The failure sets are byte-identical — `comm` in both directions returns nothing.** The only difference is `+1 test / +1 pass`, which is the new contract test in §2.

```
$ comm -23 fail-head.txt fail-dev.txt   # only on this branch -> new regressions
(empty)
$ comm -13 fail-head.txt fail-dev.txt   # only on dev -> fixed or flaky
(empty)
```

One caveat on methodology, since it bit me once. My first head run reported 1212 tests against `dev`'s 1351 and showed four apparent new failures. That was **collection mismatch, not a regression**: the run happened in the working tree while I was mid-edit on the README, and `update-notifier`'s tests are timing-sensitive — `concurrent checks use one writer and leave a valid cache` failed with `expected: 'update_available', actual: 'silent'` on that run and passed on `dev`. This branch does not touch `update-notifier` at all (`git diff --stat upstream-dev..wb-dev -- '*/update-notifier*'` is empty). Re-extracting `da6c2e0` cleanly and re-running brought it to 1352 tests with a failure set identical to `dev`'s. Both sides below are therefore clean extractions run the same way.

The failures are environmental on this machine: **symlink creation silently no-ops**. `fs.symlinkSync(target, path)` returns without throwing, but nothing is created:

```
fs.symlinkSync(input, output, 'file');
fs.existsSync(output)   // false
fs.readlinkSync(output) // ENOENT
fs.readdirSync(cwd)     // ['a.json']   <- the link is simply absent

fs.symlinkSync(realDir, linkedDir, 'dir');
fs.existsSync(linkedDir) // false
```

So every symlink-based guard test can never establish its precondition. The CLI is asked to reject an alias that does not exist, it correctly succeeds, and the test sees exit code 0 where it expected 1 (`expected: 1, actual: 0`, `operator: 'strictEqual'`). That is a machine capability gap, not a product defect — the guard code is never reached.

```
archive build accepts Windows-style absolute output paths
archive build excludes untracked files and external symlinks from the live working tree
archive build is byte-for-byte reproducible across caller time zones without system zip
archive build rejects an unmerged index and preserves an existing archive
built archives contain the embedded notifier runtime
CLI entry detection survives a realpath or symlink alias
clean staging rejects tracked symlinks before reading through them
compare rechecks every target immediately before committing the artifact pair
compare rejects a dangling receipt symlink to the future artifact path
compare rejects an artifact path that aliases either architecture input
compare rejects a JSON receipt symlink to a non-JSON target
compare reports target alias before CLI extension for symlink alias
compare reports target alias before CLI extension for a derived receipt symlink
deliver rechecks aliases immediately before committing a verified candidate
deliver rejects a future-path alias of its JSON input with a structured diagnostic
package smoke rejects every dependency metadata field in a built package
preview: rejects destructive or unsupported startup targets before watching
render and deliver reject HTML symlinks resolving to non-HTML files
render rejects a meta.output symlink that resolves to a non-html target
render rejects a meta.output that escapes through a directory symlink
render rejects an output symlink that aliases its JSON input
render reports an output symlink cycle as a structured output diagnostic
the shared renderer rechecks its guarded output immediately before writing
workflow migration reports a cyclic-symlink destination as structured JSON

# update-notifier concurrency / lease group (timing-sensitive on this filesystem)
a fenced owner rechecks its active claim immediately before network access
a forged retirement symlink cannot move an active claim outside its cache partition
a last-good notice remains acknowledgeable after the refresh commits a new candidate
an empty precheck snapshot cannot start a second concurrent network request
an overlapping check reads the last-good candidate while another process refreshes it
concurrent checks use one writer and leave a valid cache
local-only preview and compare publish only revision-verified evidence
malformed active-claim shapes recover after the hard lease
two promoters cannot replace and then steal a stale empty active claim

# repository-evidence-replacement.test.mjs (local git rejects `init --template=`)
pinned evidence ignores commit replacement adds a missing file
pinned evidence ignores commit replacement adds out-of-range lines
pinned evidence ignores commit replacement removes a valid file
pinned evidence ignores commit replacement shortens valid lines
pinned evidence ignores blob replacement adds out-of-range lines
pinned evidence ignores blob replacement shortens valid lines
pinned evidence ignores custom replacement namespace adds out-of-range lines
pinned evidence ignores dirty working files without replacement refs
```

They break down as symlink-dependent alias guards (`render` / `compare` / `deliver` / `preview` / clean staging / CLI entry), `archive build` and `package smoke` cases that need a `zip` binary and a real git repo (an archive checkout has neither), the `update-notifier` concurrency group that is timing-sensitive on this filesystem, and eight `repository-evidence-replacement` cases where local git resolves an empty `--template=` to `\\.\nul` and fails the fixture (`fatal: unable to access '\\.\nul': Invalid argument`). I have not verified these against a symlink-capable machine — if your Linux CI is green on all of them, that confirms they are environmental.

### One harness disclosure, and one test-hygiene note

`npm test` does not terminate on this machine, and the reason separates into two parts — one environmental, one a genuine robustness gap.

**The trigger is environmental.** `test/preview.test.mjs:57` expects `startPreview()` to reject an aliased output. Because the directory symlink is never created here (see above), there is no alias to reject, so `startPreview` **resolves** and the assertion fails with `Missing expected rejection`. The test then aborts without ever stopping the preview it just started.

**The consequence is a robustness gap.** A started preview holds an `fs.watch` handle (`bin/preview.mjs:591`) plus a live HTTP server. Nothing closes them when the assertion throws, so the `node` process finishes all of its tests and then never exits. `scripts/run-tests.mjs` waits with `spawnSync` and no timeout, so the whole suite hangs indefinitely with no diagnostic pointing at the real cause — I only found it by bisecting one test file at a time.

Confirmed not caused by this branch:

- `bin/preview.mjs` and `test/preview.test.mjs` are **byte-identical** across `dev`, `main`, and this branch — neither upstream nor I touched them.
- A clean `dev` checkout reproduces it exactly (killed at 101s; the tests themselves finished in ~4.5s).
- Running that single test alone is enough to hang the process.

To get a terminating run I added `--test-force-exit` to the runner's own flags — `run-tests.mjs` already uses `node --test --test-concurrency=2`, so the command is otherwise unchanged:

```bash
node --test --test-concurrency=2 --test-force-exit test/*.test.mjs
```

Results are unaffected: `preview.test.mjs` reports the same pass/fail, just in seconds instead of never.

On a symlink-capable machine this test presumably passes and never hangs, so I am **not** claiming `npm test` is broken for everyone. The narrower suggestion: wrap the `assert.rejects(startPreview(...))` calls in `try/finally` and stop whatever started, so an unexpected resolve cannot take down the whole suite. Adding `--test-force-exit` to `run-tests.mjs` would make that failure mode loud instead of silent. Glad to send either as a separate PR.

---

## Note on the README conflict resolution

`dev` had replaced the trailing agent with GitHub Copilot; this branch appends WorkBuddy while preserving it:

- EN: `Archify is a rendering and validation system for Cursor, Claude Code, Codex CLI, OpenCode, GitHub Copilot, and WorkBuddy.`
- ZH: `Archify 是一套基于 Node.js 的渲染与校验系统，并以 Agent Skill 的形式支持 Raven、Cursor、Claude Code、Codex CLI、OpenCode、GitHub Copilot 和 WorkBuddy。`

Two things worth flagging:

- `dev`'s **English** sentence **drops Raven**, while the **Chinese** one **keeps Raven**. That asymmetry is `dev`'s existing state — I preserved it rather than silently changing it. Happy to fix separately if you'd like.
- A commit that deleted one word to fit the old 2070-word budget is **no longer needed** (the budget is now 2125 and `da6c2e0` covers the current ceiling), so it was dropped during the refresh.

**`archify.zip` does not need rebuilding.** This PR touches only docs and two test files — `git diff --stat upstream-dev..wb-dev -- archify/` shows just `test/cursor-onboarding.test.mjs` and `test/landing.test.mjs`, neither of which is packaged into the archive.

tt-a1i:
Thank you for the detailed WorkBuddy receipt and `dev` refresh. I have read the later host-level section: it now includes the WorkBuddy version and installed path, the prompt inside WorkBuddy, the host-reported Skill base directory, and a validated delivered artifact. That answers the evidence request I made; I should not keep asking you for only a direct CLI run. The five-tab switcher contract and manual-copy boundary are also addressed in your reported tests. I will review the current head and CI against today's `dev` before final disposition, while treating the host transcript as author-provided evidence rather than a test I ran locally.

tt-a1i (approved):
The WorkBuddy manual-copy onboarding is now scoped to documentation and contract tests. I checked the current diff against `dev`: the five existing generated-agent tabs remain unchanged, both Start sources carry the exact installed path, and no `--agent workbuddy` command is claimed. The author supplied a WorkBuddy host discovery/invocation transcript with a validated artifact; I am treating that as author evidence, not a local host test. I merged current `dev` into the contributor branch and re-ran the focused onboarding/landing tests (7/7), release-identity check, and Start-page generator parity. The full current-head CI matrix and browser job have now passed. Approved for the documented scope.

tt-a1i:
I merged #178 to `dev` before confirming the merge decision with the project owner. I have now reverted that merge in `9b3d3f0c`; the `dev` file tree is back to its exact pre-#178 state. This is a correction to my merge process, not a finding against your contribution or the WorkBuddy evidence you provided. Your PR and original commits remain in the Git history. No further work is needed from you right now; thank you for the careful follow-through.

tt-a1i:
After reconsidering the scope, we do not think dedicated WorkBuddy onboarding needs to be maintained in the repo at this point. Our preferred path is for users to give WorkBuddy's AI the repository URL, https://github.com/tt-a1i/archify, and ask it to read the installation instructions and help install the existing Archify Skill. That keeps the instructions shared instead of adding a separate installation section for each agent.

Thank you for the implementation, tests, and detailed WorkBuddy verification. We asked you for several follow-ups before making this scope decision, and I am sorry for the back-and-forth. We will keep the revert in place and do not need further changes to this PR. If that repo-URL installation flow exposes a concrete blocker, a focused report would help us decide what actually needs fixing.

Teamirni:
Thanks for the clear scope decision — keeping the install surface maintainable
makes sense, and I'm not asking to reopen #178.

You asked whether the repo-URL flow exposes a concrete blocker. I ran it as an
agent with only the repository in hand, and here is what resolved and what didn't.

**What resolved**

The Installation options table already provides a host-agnostic manual path, and
it generalizes cleanly. I verified the artifact itself: `archify.zip` (1,840.9 KB,
sha256 `d2296515…`) contains 79 entries under a single top-level `archify/`
directory, with `SKILL.md` at that folder's root (16,396 bytes). So extracting it
into a host's skills directory yields `<skills-dir>/archify/SKILL.md`, which is
exactly the layout every surface in the table relies on. The Raven row is what
makes this inferable — it is the precedent for a non-switcher host.

**The two things that did not resolve from the repo alone**

1. **The host's own skills directory.** It is enumerated for the seven listed
   surfaces; for any other host the agent must guess or ask the user. On WorkBuddy
   that path is `~/.workbuddy/skills/`, and there is nothing in the repository it
   can be derived from. This costs one clarifying exchange rather than blocking
   the flow outright.

2. **Archive nesting vs. a strict importer.** A host whose native "upload a skill
   archive" importer expects `SKILL.md` at the archive root will reject
   `archify.zip`, because the archive nests it one level (`archify/SKILL.md`).
   Extract-to-directory is the route that works there. Worth knowing only if you
   ever expand the upload-based rows beyond the current two.

**Optional — entirely your call**

If either point is worth a line, the cheapest form is not an onboarding section
but one row in the Installation options table, e.g. *"Any host that loads skills
from a directory: extract `archify.zip` into that directory (Raven, WorkBuddy, …)"*.
If you would rather leave the table as-is, that is completely fine — I will keep
the host-directory note in my own notes instead.

Thanks again for the detailed review and for stating the rationale so clearly.
Happy to re-run this flow on another host if that would help.

tt-a1i:
Correction to my earlier wording: the README's Raven ZIP instructions describe Raven only. They are not a general installation recipe for WorkBuddy or other hosts. Your repo-URL test shows that an agent still needs a reliable way to find its own host's Skill directory.

For an unlisted host, the agent should check that host's documentation or environment, or ask the user. We opened [#529](https://github.com/tt-a1i/archify/pull/529) to remove the Raven-specific installation example from the shared README and website guides; it has now merged into `dev`, and the same documentation change has merged into `main` via [#530](https://github.com/tt-a1i/archify/pull/530). #178 remains reverted, and we don't need changes from you on it.

Your note about importers that require `SKILL.md` at the archive root is useful as a separate compatibility case. The extract-to-directory route you tested did work. Thanks again for spelling out both limits clearly.

tt-a1i:
One optional follow-up, since you offered to try the flow on another host: Archify already has a community Hermes Agent integration on `dev` ([#283](https://github.com/tt-a1i/archify/issues/283), [integration guide](https://github.com/tt-a1i/archify/blob/dev/integrations/hermes-agent/README.md)). Its original contributor has reported a successful install and first diagram, so no new Hermes adapter is needed.

If you use Hermes and would enjoy an independent check, we'd welcome a run through the `dev` guide from installation and Skill discovery to asking Hermes in a normal chat to deliver a small diagram. In particular, it would be useful to know where an agent starting from the repository URL still has to guess, whether the documented install path works in your environment, and whether you get the expected JSON and HTML artifacts. A concise note with the Hermes version, OS, method, and any concrete blocker is plenty. No pressure to set up Hermes just for this, and no implementation work is expected unless you spot an issue you want to tackle.

Thanks again for offering to help validate another host.

Teamirni:
Since you asked whether the dev Hermes guide holds up in another environment: I don't
run Hermes, so I did not set it up, and I won't claim a runtime result I don't have.
Here are two things I checked against the guide itself.

**1. The Windows Skill path looks like the WSL/Linux layout, not the native Windows default**

The guide gives `~/.hermes/skills/archify` or `%USERPROFILE%\.hermes\skills\archify`, and
the PowerShell block creates `$env:USERPROFILE\.hermes\skills`.

Per Nous's Windows (Native) guide, "native data lives under `%LOCALAPPDATA%\hermes`, WSL
data lives under `~/.hermes`", and its uninstall section calls `%USERPROFILE%\.hermes`
"a legacy CLI/WSL data dir". `scripts/install.ps1` agrees:

    [string]$HermesHome = $(if ($env:HERMES_HOME) { $env:HERMES_HOME } else { "$env:LOCALAPPDATA\hermes" }),

So on a default native-Windows install the Skill would have to live under
`%LOCALAPPDATA%\hermes\skills\archify`; a reader following the Windows block places it
where native Hermes does not look, unless they set `HERMES_HOME` first. Worth confirming
which layout that block targets — I did not install Hermes to test it end to end.

**2. The Windows commands need a privilege a stock Windows user does not have**

Both PowerShell blocks use `New-Item -ItemType SymbolicLink`. Measured on Windows 11,
standard (non-elevated) account, Developer Mode off:

    New-Item -ItemType SymbolicLink -Path $link -Target $target
    → FAILED: requires administrator privileges / the client does not hold the required
      privilege (AllowDevelopmentWithoutDevLicense unset)

The guide does note that symlink creation needs the relevant permission, but there is no
documented fallback for users who don't have it. `-ItemType Junction` works without
privilege, across volumes, and resolved the Skill in the same test:

    New-Item -ItemType Junction -Path "$env:USERPROFILE\.hermes\skills\archify" -Target "D:\path\to\archify"
    → SUCCESS; Test-Path "<link>\SKILL.md" → True

OS/method: Windows 11, PowerShell 5.1, standard user, no Hermes runtime installed. If the
junction fallback is worth documenting, I'm happy to send it as a small patch to the
integration guide; if not, a note in my own records is fine either way.

PR #192, opened 2026-08-29
fix(packaging): normalize line endings (fixes #144)

### LINKED ISSUE #144: [Bug]: Tracked files check out as CRLF on Windows, so working-tree bytes stop matching the index

yunaremaia:
Fixes #144. Adds `* text=auto eol=lf` so working-tree bytes match the index on Windows (core.autocrlf=true).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -12,3 +12,6 @@
 # Deterministic build outputs committed for zero-install and release verification.
 /archify/renderers/shared/generated-brand-marks.mjs linguist-generated=true
 /archify/renderers/shared/generated-validators.mjs linguist-generated=true
+
+# Normalize line endings (fixes #144)
+* text=auto eol=lf


yunaremaia:
Duplicate of #191 (same CRLF fix). Closing to avoid clutter; keeping #191.

jwbbf9g8ng-prog (commented):
- [__]()

PR #194, opened 2026-08-29
feat(i18n): add Japanese (ja) locale

### LINKED ISSUE #193: feat(i18n): add Japanese (ja) as a Viewer locale
## Problem and value

`meta.locale` currently accepts `en` and `zh-CN`. For Japanese diagrams, the authoring contract requires omitting `meta.locale` and disclosing that the fixed Viewer UI and `<html lang>` fall back to English — so every Japanese artifact ships with an English toolbar, navigation dock, node finder, route probe, export menu, default legend entries, and aria strings next to Japanese authored labels.

PR #108 (first-phase Viewer localization) listed Japanese under "what deliberately did not change"; this issue proposes that follow-up, in the same spirit as #152 does for zh-TW.

As #152 notes for Traditional Chinese, this is terminology work, not mechanical conversion: the catalog needs UI-register Japanese (e.g. `凡例` for Legend, `経路をトレース` for Trace a route), plural-suffix pairs collapse to one form (Japanese has no grammatical plural), and several workflow legend entries must be sized to the one-row legend band so the default legend does not drop.

One rendering detail beyond the catalog: the template's CJK font fallbacks name Simplified Chinese faces only (`Noto Sans Mono CJK SC`, `PingFang SC`, …). They cover the Japanese repertoire but draw Chinese glyph variants and lack tuned kana, so Japanese-face insertion ahead of them (not replacement) is part of the same slice.

## Proposal

Add `ja` to `SUPPORTED_LOCALES` / the schema enum through the existing tuple mechanism (a missing translation fails the build), plus the locale-scoped font-fallback insertion above. `en` and `zh-CN` rendered output stays byte-identical; the English-fallback contract for other languages is unchanged.

A PR with the full catalog (518 keys), schema/validator regeneration, tests (including a real-Chrome locale-parameterized viewer test), and rendered evidence is ready to open.

Rererr:
## Problem and value

Japanese authors can localize every authored string, but the renderer-owned surface stays English (or zh-CN): the standalone viewer toolbar, navigation dock, node finder, route probe, export menu, default legend entries, and every aria/label string, plus `<html lang>`. The authoring contract currently requires disclosing this fallback for every non-`en`/`zh-CN` diagram. Adding `ja` to the bounded localization set removes that per-diagram disclosure and gives Japanese readers a fully localized standalone artifact. (Scope note: in embedded mode the viewer chrome is hidden, so the visible win there is the SVG-embedded legend/context strings and `<html lang>`; the full win is the standalone viewer.)

PR #108 listed Japanese under "what deliberately did not change"; this is that follow-up. Closes #193.

## Scope

- What changed:
  - `renderers/shared/i18n.mjs`: `ja` in `SUPPORTED_LOCALES` + a third tuple element for all 518 keys (the existing tuple-length check enforces completeness; `.one`/`.other` pairs carry the same Japanese string since Japanese has no grammatical plural).
  - `schemas/common.schema.json` locale enum + regenerated `generated-validators.mjs` (`npm run generate:validators`; the diff is the enum and the five `data === "ja"` branches only).
  - Locale-keyed font-fallback table (in `i18n.mjs`, applied in `applyTemplate`): for `ja` only, Japanese CJK faces are inserted ahead of the existing Simplified-Chinese faces (which cover the repertoire but draw Chinese glyph variants). The substitution is idempotent and uses function-form replacement (no `$`-pattern hazards). Verified with CDP `CSS.getPlatformFontsForNode`: ja resolves to Hiragino Kaku Gothic ProN on macOS, zh-CN still to PingFang SC.
  - Docs: SKILL.md, references/authoring-contract.md (the fallback contract structure is preserved, now excluding ja), schemas/README.md, README ×3. No CHANGELOG entry: v2.16.0 is released, and adding an Unreleased bullet trips the release-identity gate, which then requires opening a new development identity (prerelease version bump) — release work CONTRIBUTING reserves for maintainers. Happy to add the entry once a maintainer opens the next cycle.
  - Tests: locale list; ja catalog parity and template `{{i18n:}}` spot checks; plural invariant across all 27 `.one` bases; `ja-JP` rejected as unsupported; a locale-parameterized real-Chrome viewer test generating both zh-CN and ja runs (finder/route/export copy, preset-badge pseudo-elements, visual-check receipt) so coverage cannot diverge between locales; a 0-occurrence assert on the pre-replacement font string so a drifting duplicate inside the export-style literals cannot silently drop the ja stack; a rendered legend row-count test for ja against the shipped viewBoxes (with per-renderer legend gutters); and a schema-driven test asserting every semantic-kind enum value has viewer.kind/legend catalog keys, so a future enum addition cannot ship untranslated.
  - `en` / `zh-CN` output: byte-identical, verified by rendering all five shipped examples per locale on base vs HEAD (10/10 SHA-256 identical; re-measured against 5de7275 / v2.16.0 after each rebase).
  - The workflow fixed grid, `guide` recipes/language detection, the docs-site language toggle, Share Card canvas font stacks (they contain no CJK faces in any locale today), and two pre-existing cross-locale rendering issues we hit while validating (the default workflow legend drops silently in `en`, and the preset badge can sit under the fixed toolbar in every locale — Japanese only deepens an overlap English already has; measured, and label edits cannot fix it). Both are worth their own issues; happy to file them with the measurements.
- No unrelated changes: confirmed (`git diff --check` clean; single commit).

## Stability impact

- Compatibility and migration risk: additive locale; existing typed JSON is unaffected; unsupported locales are still rejected by the schema.
- Renderer, validator, package, or generated-artifact risk: the font substitution is a locale-keyed table that resolves to no substitution for every locale but ja; validators regenerated from the schema; archify.zip rebuilt.
- Failure behavior and rollback path: an incomplete ja tuple throws at module load (existing mechanism); rollback is reverting one commit.

## Tests run

- `cd archify && npm test` (re-run after rebasing onto 5de7275 / v2.16.0) — Node v26.7.0: 1020 tests / 988 pass / 0 fail / 32 skipped; official Node 22.23.2: 1020 / 992 / 0 fail / 28 skipped (the archive byte-match gates run only on Node 22 and pass there). The test count grows with the base; none of the growth is from this PR beyond its own added tests.
- `ARCHIFY_CHROME=… node --test test/i18n.test.mjs`: 16 / 16 pass (real Chrome, macOS; includes both zh-CN and ja browser runs).
- `deliver <type> … --quality showcase --json` for 5 Japanese sample diagrams (one per mode): all 9/9 checks, 0 errors, 0 warnings.
- `visual-check` on the 5 samples: readability / viewerChrome / captures pass, `overflowX` false on all viewports; the vertical page-scroll flags match the shipped English examples exactly (pre-existing).
- Not run: the remaining ARCHIFY_CHROME-gated suites beyond `i18n.test.mjs`; Node 18/20/24 lanes (relying on CI); Linux/Windows font resolution (the `Noto Sans Mono CJK JP` ordering targets those platforms but was measured on macOS only).

## Visual evidence

Screenshots of the five Japanese samples (light/dark):

ja-architecture visual-check 1440x900 dark

ja-workflow visual-check 1440x900 dark

ja-sequence visual-check 1440x900 dark

ja-dataflow visual-check 1440x900 dark

ja-lifecycle visual-check 1440x900 dark

Visual review: passed — all five samples reviewed by a human on macOS Chrome; no clipping, ellipsis, or wrapping defects in the Japanese labels or viewer chrome.

## Generated artifacts

- `renderers/shared/generated-validators.mjs` (schema enum regeneration).
- `archify.zip`, rebuilt with the official Node 22.23.2 binary (bundled zlib 1.3.1). Note: a Node 22 whose bundled zlib is 1.2.12 (e.g. Homebrew `node@22`) passes the builder's major-version guard but produces different archive bytes; worth a separate issue and we can file it with measurements.

## Checklist


Al-hassan-961:
Closing this PR as we are not adding Japanese localization at this time. Thank you for the contribution!

Al-hassan-961:
Closing this PR as we are not adding Japanese localization at this time. Thank you for the contribution!

sunsunsun-java:
Thank you for the thorough implementation and validation. The contribution appears well scoped, but we are not expanding the supported Viewer locale set beyond en and zh-CN at this time.  
We’re closing this PR as a product-scope decision rather than because of an implementation defect. The work will remain available for reference if Japanese localization enters the roadmap later.

Rererr:
Understood — thanks for the clear scope decision, and for taking the time to review the work. No objection at all.

The branch stays on my fork if Japanese localization is ever revisited.

One loose end from the PR description: while validating I measured two pre-existing, locale-independent rendering issues (the default workflow legend drops silently in `en`, and the preset badge can sit under the fixed toolbar). Happy to file those as separate issues with the measurements if that's useful — just say the word.

PR #203, opened 2026-08-30
fix(dataflow): drop degenerate points so vertical edges keep a real final segment (#169)

### LINKED ISSUE #169: [Bug]: Vertical auto-routed data-flow edges emit a zero-length final segment, so marker-end renders at the wrong angle

yunaremaia:
## Summary
Fixes #169. A purely vertical auto-routed dataflow edge (same column, different rows within one stage) emitted a path like `M 100 186 L 100 186 L 100 356 L 100 356`. SVG derives `marker-end` orientation from the last path segment, so the zero-length final segment left the arrowhead angle undefined and it rendered flat / left-pointing instead of into the target node.

`pathFor()` now drops consecutive duplicate points before building the polyline. Only exact duplicates are removed (collinear points are preserved), so ordinary routes are unchanged. If the whole route degenerates it still keeps both endpoints so the path stays well-formed.

## Verification
- New test `archify/test/vertical-edge.test.mjs` reproduces the issue's repro (RED) and asserts no zero-length final segment (GREEN).
- `geometry.test.mjs`, `semantic-flow.test.mjs`, `render-output-checks.test.mjs` still pass (84/84 in the affected suites).

Fixes #169


tt-a1i (commented):
Reviewed at exact head 3e7c6081e07039c75c3feed3cd8e99cb80d13aa5. The Data Flow-specific duplicate-point removal fixes the #169 vertical marker orientation without widening the shared geometry contract. Maintainer follow-up synced current main and rebuilt archify.zip with Node 22. Local evidence: affected suites 84/84; full suite 981 passed, 0 failed, 27 environment-skipped; canonical ZIP rebuilt byte-for-byte and its renderer hash matches source. Required CI is green across Node 18/20/22/24, WebM/browser, ZIP freshness, and package smoke on Ubuntu/macOS/Windows.

PR #207, opened 2026-08-30
fix(viewer): position semantic passport chip synchronously before paint (#200)

### LINKED ISSUE #200: Semantic passport (focus-chip) visibly jumps position on open — unhide then RAF-deferred reposition
## Summary

The semantic passport panel (`#focus-chip` / `.relationship-lens`) briefly renders at its default/stale position, then visibly snaps to its correct location one frame later, every time it opens or the focus target changes. This reads as a jarring position jump rather than a stable open.

## Root cause

In `assets/template.html`:

1. `chip.hidden = options.hideChip === true || normalized.length !== 1 || selectionMode;` (line 8625) unhides the panel **synchronously**. At this point it paints at whatever `top` it currently has — either the CSS default (`top: 1rem`, line 1649) on first open, or the stale `chip.style.top` left over from the previously focused node.
2. Immediately after, `requestLensPlacement()` (line 8628) is called, which does:
   ```js
   function requestLensPlacement() {
     if (lensFrame) return;
     lensFrame = requestAnimationFrame(placeRelationshipLens);
   }
   ```
3. `placeRelationshipLens()` (lines 8426-8503) computes the real target position from the focused node's vertical center (`nodeCenter - chip.offsetHeight / 2`, plus collision avoidance against the legend/diagram-nav) and only then does `chip.style.top = Math.round(top) + 'px';` (line 8502) — **one animation frame after** the panel already became visible and painted.

Because the browser can paint the intermediate frame (chip visible at the old/default `top`) before the `requestAnimationFrame` callback runs and corrects it, the panel visibly jumps from one spot to another on every open.

This isn't a one-off: the same unhide-then-`requestLensPlacement()` sequence repeats at multiple call sites — lines 7644, 8119, 8146, 8628, 8714 — so the jump happens on first focus and on every subsequent focus-target change, not just a one-time init glitch.

## Reproduction

1. Open any delivered Archify HTML (architecture/workflow/etc.) with more than one node.
2. Click/focus a node to open the semantic passport panel.
3. Watch the panel's vertical position on open, especially when focusing nodes at different heights in sequence — the panel appears at one `top`, then snaps to a different `top` within a frame.

## Expected behavior

The panel should compute its final position before it becomes visible (or become visible in the same frame the position is applied), so there's no observable pre-placement flash/jump. E.g. compute `placeRelationshipLens()`'s target top synchronously before setting `chip.hidden = false`, or keep the panel `visibility: hidden` (not `display: none`) until the RAF-computed position is applied, then reveal it.

## Related

Found while investigating a separate sigil/label overlap report (#199); this is a distinct panel-positioning timing issue, not a layout/validator gap.

sx4im:
## Problem and result

Fixes #200. Opening Semantic Passport or switching nodes made the panel visible before computing its new position, leaving it at a default or stale position until the next animation frame. The panel now computes its target position synchronously when selection or relationship content changes; continuous scroll/resize work remains frame-coalesced.

## Integration and impact

Final integration head `af87ba8` includes `dev` at `457d242` and preserves the original author's `f3da4c4` commit. The implementation lives in the authoritative `viewer/focus.js`, with the template generated from it. Pending placement frames are canceled before immediate placement. Renderer geometry, node/edge identity, public JSON and CLI contracts are unchanged.

## Validation

- Transplanted regression on `dev` at `f3ddb11`: 3 passes and 2 failures, including the real Chrome assertion that initial `chip.style.top` was empty before the next paint.
- Candidate `0e5cbab`: 48 focused tests passed, 0 failures/skips with real Chrome (Semantic Passport, relationship lens, Radar, Viewer chrome and desktop reader); 8 Gallery/README artifact tests passed, 0 skips; golden/schema/template/version checks passed.
- Five packaged examples have byte-identical SVG before/after; only Viewer script timing changes. Ego browser at 1440×900 confirmed immediate positions on auth→s3 selection (16px→254px), and the S3 panel was visually inspected for readable content and containment. This is a bounded desktop check, not maintainer trial-use acceptance for main.
- Final source after syncing #290: 65 CLI/visual-check tests passed, 0 failures/skips. The earlier Viewer evidence is reused because #290 only changes visual-check sidecar placement, not Viewer source.
- Final-head remote CI pending; no main promotion requested.

## Generated artifacts

Rebuilt Viewer template, packaged/root examples, Gallery HTML/manifest, README motion proof/receipt, and canonical Node 22 ZIP from the combined source. Unrelated historical diagrams were not refreshed. `git diff --check` passed.

CI follow-up: run `35128770620` on `529a1b7` failed because the checked-in Checkout comparison still embedded the old Viewer. Regenerated `examples/checkout-platform-delta.html` and its receipt from their authoritative JSON; all 29 Architecture Delta tests passed, 0 skips. The correction changes no packaged source or ZIP bytes. New-head CI is pending; the failed run is not acceptance evidence.


tt-a1i (changes requested):
Thanks for the synchronous positioning fix. Reviewed exact head f3da4c43cdb1a3be5b12c9135cbc7c1a41d94c38.

[P2] Include the positioning fix in archify.zip. Comparing every payload entry with canonical source finds six mismatches: assets/template.html and the five bundled example HTML files. The archive retains the previous asynchronous positioning implementation, so ZIP-installed users still get the old behavior even though source/browser tests exercise the fix. Rebuild with canonical Node22 from the final tracked source and verify the extracted Viewer.

Source changes were checked across all template/generated copies. The targeted suite including real Chrome passed 28/28; golden and Gallery checks also passed, and the source-backed/GIF receipts match their artifacts. No further confirmed source regression. Base conflicts and absent current-head CI are separate gates; no source edits or merge.

PR #209, opened 2026-08-30
chore: override vulnerable fast-uri version

seekskyworld:
## Summary
- force fast-uri to a patched version via npm overrides
- refresh package-lock to fast-uri 4.1.3

## Verification
- `npm audit --json` reports 0 vulnerabilities
- `npm test` passes (594/594)

Resolves the high-severity fast-uri advisories GHSA-v2hh-gcrm-f6hx and GHSA-7p8r-x3mc-p8w7.


tt-a1i (approved):
Verified at exact head e8c5caa0189934f6cee2f6955530725bfa79af67.

The security fix is valid. I tightened the override from an unbounded cross-major range to exact fast-uri 3.1.5, which is the patched release within AJV's declared ^3.0.1 compatibility range, then merged the latest main without rewriting the contributor commit.

Evidence on Node 22.23.2:
- npm ci: passed
- npm ls: ajv@8.20.0 -> fast-uri@3.1.5 overridden
- npm audit: 0 vulnerabilities
- full suite: 1008 tests, 981 passed, 0 failed, 27 environment skips
- deterministic ZIP freshness: passed
- extracted-package smoke on macOS: passed

Approved subject to the required GitHub checks on this same head.

PR #210, opened 2026-08-30
perf: index label route clearance candidates

seekskyworld:
## Problem and value

Large diagrams currently compare every relationship label against every route to measure clearance. A conservative grid index removes distant candidates before the existing exact segment calculation. It preserves diagnostic ordering, distances and segment identities, with full-scan fallbacks for large routes, unselective queries and unsafe grid coordinates.

The prior non-terminating-loop review is addressed: all four grid coordinates must be safe integers before either nested loop. Unsafe routes enter the global candidate set; unsafe label queries use the full route list. No dependency, authored geometry, schema, validation threshold or version changes.

## Integration and cost

Original `bca9add69604be0f38292eb07472bdf5fcdff509` history is preserved. Current head `86ebff230f6ab11f1ccfc6be863e7a9456ae64b7` integrates dev `2d87fec2e555f1dd266c4fbb2a866bc744034343`, including #449, #416 and the shared browser gate #440. The merge retains dev's large-receipt minimum-clearance regression alongside the indexing tests.

The index is private to one shared clearance function, with bounded per-route indexing and deterministic ordering. Its benefit is strongest for sparse, larger diagrams; dense/unselective inputs can still require the original full comparison. This is not a claim that overall diagram authoring is hundreds of times faster.

## Verification

- On combined source `5440dbb`, geometry/layout/render-output suites: **201 passed, 0 failed, 0 skipped**.
- A seeded 2,000-case differential comparison against `128a8aa` produces identical complete results, including hit order, distances and segment indexes. Cases include negative and unsafe coordinates, duplicate identities, diagonal/multisegment routes, varied rectangles, zero/negative/nonfinite/unselective thresholds. The clearance implementation is unchanged by the subsequent dev sync/package commit, so this evidence is reused with that scope.
- Same-machine Node 22.23.2 sparse fixture, 3,000 routes and 3,000 labels, two ABBA cycles: baseline 3266–3304 ms; candidate 8.57–10.58 ms. Both return the same empty hit set. This measures only the synthetic clearance stage, not real end-to-end generation or a universal speedup.
- `bc26a22` geometry + package gates (unchanged production payload after #440 sync): **82 passed, 0 failed, 2 explicit platform/version skips**. Canonical ZIP freshness is checked. Golden renderer/source/package examples remain byte-identical; no new visual acceptance is claimed because the generated diagrams are unchanged.
- After syncing #440, current-head shared-gate/geometry tests pass **66/66 with no skips**. The only intervening package-manifest change adds a repository-only npm script, excluded from the canonical ZIP.
- Full final-head remote CI must pass before merge. Local focused results do not substitute for the Node matrix, browser/WebM and platform gates.

## Generated artifacts

Rebuilt only `archify.zip` with official Node 22.23.2 from tracked combined source. Renderer examples remain byte-identical, so Gallery/proof/Viewer outputs need no regeneration. `main` remains unchanged; this PR targets dev integration only.


archify/renderers/shared/geometry.mjs
```diff
@@ -96,8 +96,76 @@ export function segmentRectIntersectionLength(segment, rect) {
...
+    const bounds = routeBounds(route);
+    const x1 = Math.floor(bounds.x1 / ROUTE_INDEX_CELL_SIZE);
+    const y1 = Math.floor(bounds.y1 / ROUTE_INDEX_CELL_SIZE);
+    const x2 = Math.floor(bounds.x2 / ROUTE_INDEX_CELL_SIZE);
+    const y2 = Math.floor(bounds.y2 / ROUTE_INDEX_CELL_SIZE);
+    const cellCount = (x2 - x1 + 1) * (y2 - y1 + 1);
+
+    // Very long routes are cheaper to test once per label than to duplicate
+    // across a large part of the index.
+    if (!Number.isSafeInteger(cellCount) || cellCount > ROUTE_INDEX_MAX_CELLS_PER_ROUTE) {
```
tt-a1i, line 142: **[P1] Fall back before iterating unsafe cell coordinates.** `Number.isSafeInteger(cellCount)` is not sufficient: for a route at `x = 1e20`, `x1 === x2` can make `cellCount` equal to `1`, while the cell coordinate itself is outside the safe-integer range. At that magnitude `x += 1` does not change `x`, so the loop below never terminates. Please require all four cell coordinates to be safe integers before either nested loop; treat an unsafe route as global and an unsafe query as a full-scan fallback. Add a bounded-completion regression test for this case.

tt-a1i (changes requested):
Reviewed at exact head `f7dc11958a220d41c0a2cd0a7ff36685483688d2`. The spatial broad-phase is a valuable optimization, and I verified that ordinary finite-coordinate behavior remains identical in a 2,000-case differential check. However, two merge blockers remain:

1. Unsafe grid coordinates can make index construction or lookup loop forever. `cellCount` may still be a safe integer even when `x1`/`x2` or `y1`/`y2` is not; at that magnitude, `x += 1` or `y += 1` no longer changes the value. Route indexing should put such routes in the global set, while label queries should fall back to the full route list.
2. `archify.zip` was not regenerated. The source `geometry.mjs` and the packaged copy have different SHA-256 hashes, so source users receive the optimization while ZIP installs keep the old implementation. Please rebuild the canonical archive with Node 22 after the code fix.

Please also commit regression coverage for unsafe finite coordinates and the unselective-threshold fallback. The PR currently has no GitHub checks and is behind `main`; after updating it, the exact new head and package freshness need to pass CI.

seekskyworld:
Addressed review blockers in latest head `bca9add`:- Guard route and label grid indices with `Number.isSafeInteger`; unsafe finite coordinates and unselective query bounds fall back to the full route list, preventing non-progressing `x += 1`/`y += 1` loops.- Added regression tests for unsafe finite coordinates and unselective thresholds.- Rebased onto latest `origin/main` (`5de7275`).- Rebuilt canonical `archify.zip` with Node 22.23.2.Verification: `npm test` 987 passed, 31 skipped (1,018 total); targeted geometry tests 55/55; 200 randomized differential checks remain identical.

archify/renderers/shared/geometry.mjs
tt-a1i: Verified the revised guard at original bca9add and preserved it in current bc26a22: all four route/query cell coordinates are checked with Number.isSafeInteger before iteration, unsafe routes use the global set, and unsafe queries use all routes. The unsafe-coordinate and unselective-threshold regressions pass. On current combined source, 201 geometry/layout/checker tests passed and 2,000 seeded differential cases matched the base exactly. Final-head geometry/package gates also pass (82 passed, 2 platform/version skips). The canonical ZIP is rebuilt, and final CI is running; no full CI pass is claimed yet.

tt-a1i (approved):
Re-reviewed current head 86ebff230f6ab11f1ccfc6be863e7a9456ae64b7 on dev2d87fec. The unsafe-coordinate loop finding is fixed and the package is current. The private spatial index retains the exact clearance calculation and deterministic ordering; its fallbacks bound index construction without changing authored geometry.

Verified 2,000 seeded differential cases, 201 focused geometry/layout/checker tests, current-head 66 gate/geometry tests, package freshness and unchanged golden output. The sparse 3,000-route benchmark shows a substantial stage-level improvement; no end-to-end generation speedup is inferred. Final-head CI35134509724 passed all Node, shared-browser/WebM, package and Windows gates. No blocking finding remains; ready for dev integration.

Issue #199, opened 2026-08-29
Semantic sigil overlaps node label on narrow/default-width nodes


seifhawamdeh:
## Summary

The component-type semantic sigil (top-left icon) can visually overlap the node label text. Reproducible with default node width on `workflow` diagrams; not caught by `validate --quality showcase` (still reports 9/9, 0 errors).

## Root cause

- `renderers/shared/utils.mjs:80` — `renderSemanticSigil(kind, { x, y, size = 11 })` is drawn at a fixed offset `node.x + 6, node.y + 6`, occupying box region `x:[+6,+17] y:[+6,+17]`.
- `renderers/workflow/render-workflow.mjs:605-622` — the label `<text>` is centered (`text-anchor="middle"`) across the full node width with baseline fixed at `node.y + 21`, with no horizontal offset or reserved margin for the sigil.
- The label glyph's effective vertical span (~`y+13` to `y+21`) falls in the same row as the sigil (`y+6` to `y+17`).
- On a narrow or default-width node (92px) with a longish label (e.g. "Claude Code", "PromptScript"), the label's centered left edge lands under `x+17`, so the two elements visually collide.

## Why validation doesn't catch it

`fittedNodeFontSize` / the node-width text-fit check only measures label width against the *total* node width — it never subtracts the sigil's reserved corner region. So a diagram can pass `validate --quality showcase` (9/9 checks, 0 errors) while the icon and title text overlap on screen.

## Reproduction

1. Author any `workflow` diagram with a node using default width (~92px) and a label like `"PromptScript"` or `"Claude Code"`.
2. Run:
   ```bash
   node bin/archify.mjs validate workflow candidate.json --quality showcase --json
   ```
3. Result: `ok: true`, 9/9 checks pass, 0 errors/warnings.
4. Open the delivered HTML — the semantic sigil (top-left icon) visibly overlaps the node label.

## Expected behavior

Either:
- Reserve horizontal margin for the sigil when centering/measuring the label (shift label center or shrink usable width by sigil footprint), or
- Add a validator check (similar to the existing sublabel/tag width checks from #61) that measures label-vs-sigil clearance and fails `showcase` when they collide.

## Related

Similar in spirit to #61 (sublabel/tag not measured for overlap) — same category of "validator reports green while text visibly overlaps on screen," but for the label-vs-sigil case specifically, which #61 does not cover.

## Samples

Below images shows a sample of this issue

[...]

PR #220, opened 2026-08-30
fix(renderers): keep node labels clear of the semantic sigil

Souptik96:
Refs #199

> Rebased onto v2.16.0 (`39a2113`) and rescoped. An earlier revision of this PR was measured against `b36d79f`; that evidence is superseded by the numbers below, because the v2.16.0 workflow migration changed the relevant geometry and added a byte-for-byte compatibility guard.

## The user problem

The semantic sigil is drawn inset in a node's top corner and the node label is centred in the same vertical band, so a label wide enough to reach that corner paints over the icon. Validation stays clean, because the text-fit rule measures a label against the whole node width and never subtracts the sigil's footprint.

Still reproducible on `main` today. Reverting only `archify/renderers/` and keeping this PR's tests:

```
test/semantic-sigil-clearance.test.mjs   # pass 0   # fail 3

dataflow:  label "Prompt Compiler" spans 55.0..145.0, sigil occupies 50.0..61.0
lifecycle: label "Prompt Compiler" spans 49.0..139.0, sigil occupies 136.0..147.0
sequence:  label "PromptScript"    spans 23.1..100.9, sigil occupies  25.0..36.0
```

Lifecycle is the mirrored case: a state without a brand mark carries its sigil in the top-**right** corner.

One shipped artifact is affected today — `examples/sequence-cache-miss-request.html`, where `Postgres` starts at 575.6 against a sigil ending at 576.0. The rest of the current examples are clear, because v2.16.0 widened workflow nodes to 132px; the defect is latent for ordinary user diagrams with longer labels rather than visible across the gallery.

## What changes

The sigil now participates in the same two-part contract as the brand top rail beside it, mirroring `brandLabelFitWidth` / `brandTopRailProblem`:

- **render** — reserve the sigil's footprint and re-centre the label into the space next to it, so an ordinary label shifts and shrinks rather than overlapping.
- **validate** — report a label that still cannot clear the sigil after shrinking to its legible minimum, i.e. when no position satisfies both "clears the icon" and "stays inside the node".

Geometry has one home: `SEMANTIC_SIGIL_INSET` / `_SIZE` / `_FOOTPRINT` are exported from `renderers/shared/utils.mjs` beside `renderSemanticSigil`, so the fitting logic measures what is actually drawn. Measured against the shapes themselves, the widest reaches x=14 of the 16-unit box → `6 + 14 × 0.6875 ≈ 15.6px`, plus non-scaling stroke ≈ 16.4px, so the 17px footprint is the nominal box and slightly conservative rather than inflated.

Two deliberate choices, both to avoid rejecting legitimate diagrams:

1. **Only the sigil's own footprint is reserved, not twice it.** Keeping the label strictly centred requires reserving both sides, which leaves too little width for ordinary labels — on the pre-migration workflow example that rejected 7 of 12 nodes. Re-centring rejected 0.
2. **The label shifts only when it would otherwise reach the icon.** Short labels keep the node centre and render byte-identically, which is why exactly one artifact changed.

- **Workflow.** Both halves are blocked there, so it is untouched and the reason is recorded in `workflow-compiler.mjs`:
  - the `fixed-v1` contract asserts `sha256(result.svg) === '4e493db…'` in `test/workflow-compiler.test.mjs`, so the label cannot be re-centred without breaking that hash;
  - reporting the overlap instead rejects the packaged `incident-response.workflow.json` — node `page`, "Page On-call", needs ~65px at the 9px floor with 59px available.
  
  Re-freezing that baseline, and whether `incident-response` should be widened, are your calls. Happy to do both in a follow-up once you say which way you want it.
- **Architecture.** It centres its label vertically (`c.y + c.height / 2`), so it shares the sigil band only on very short components; a blanket reserve would shrink labels that never collide.
- **Brand-bearing nodes.** The existing 48px symmetric reserve already puts the label's left edge at `x + 24`, clear of a 17px sigil, so those nodes are unchanged.
- No schema, no authored geometry, no diagnostic codes removed or renamed.

## Compatibility

`schema-v1` unchanged; `fixed-v1` byte contract untouched. The new validation is a `showcase`-visible layout problem in the same family as the sublabel/tag rules from #61.

The clearance comparison carries a **1px tolerance**, because `widthFactor` is documented as an approximation of advance width. Without it the rule rejects the CJK labels in `test/fixtures/automatic-routing-node-border-clearance.workflow.json` (`"校验并注入输入"`) by **0.6px** — finer than the model can justify asserting on, and that fixture exists to test routing. Real defects miss by far more: the new regression misses by ~5px.

## Testing

`node --test test/semantic-sigil-clearance.test.mjs` — new, 3 tests, one per affected renderer. Each reads the sigil transform and the label from the **same node group in the final SVG** and asserts they do not overlap. Labels are chosen per renderer so each case actually collides before the fix rather than passing vacuously.

`node --test test/layout-rules.test.mjs` — 97 → 98, adding the validator case.

**Negative control**, reverting only `archify/renderers/` and keeping both tests:

```
test/semantic-sigil-clearance.test.mjs   # pass 0   # fail 3
test/layout-rules.test.mjs               not ok 18 - dataflow: node label cannot clear its semantic sigil
                                         # pass 97  # fail 1
```

Restored: 3/3 and 98/98.

**Full suite**, `npm test` from `archify/`, against clean `main` on the same machine:

| | tests | pass | fail |
|---|---|---|---|
| `main` (39a2113) | 1010 | 934 | 39 |
| this branch | 1014 | 939 | 38 |

Failure-set diff: **no new failures**. The pre-existing 38 are the symlink-privilege and CRLF-checkout families (see #144) and the Node-22 archive pin; all reproduce on clean `main` here.

## Regenerated artifacts

`npm run render:examples` for both the repo and packaged copies. Only one artifact changed content in each location; the other four were byte-identical and are not in this diff.

- `examples/sequence-cache-miss-request.html`
- `archify/examples/sequence-cache-miss-request.html`

The changed line is a single `<text data-node-label>` `x`, `602 → 602.4`. No geometry, routing, font size, or other element moved.

## Not verified

- **`archify.zip` is stale and I could not rebuild it.** The archive packages `renderers/shared/text-fit.mjs` and `renderers/shared/utils.mjs`, both changed here. `scripts/build-zip.sh` accepts only Node 22 and this machine has Node 20.18.1 / 20.19.5, so the archive is left untouched rather than published with the wrong zlib. Please rebuild on merge, or tell me the release identity you want and I will add it.
- **No browser or `visual-check` run.** Chrome is unavailable here, so `test/desktop-reader-browser.test.mjs` and `visual-check` were **skipped, not passed**. The clearance assertions are geometric, read from the final SVG. A screenshot review of the one regenerated artifact would be a useful second pair of eyes on how a re-centred label reads.
- Windows / Node 20 only.

🤖 Assisted by an AI agent; every measurement, negative control, and suite comparison above was run and is reproducible.

### DIFF
--- a/archify/examples/sequence-cache-miss-request.html
+++ b/archify/examples/sequence-cache-miss-request.html
@@ -5240,7 +5240,7 @@
             <ellipse cx="8" cy="4" rx="5" ry="2"/>
             <path d="M3 4v8c0 1.1 2.2 2 5 2s5-.9 5-2V4M3 8c0 1.1 2.2 2 5 2s5-.9 5-2"/>
           </g>
-          <text data-node-label="" data-detail-anchor="" x="602" y="94" class="t-primary" font-size="11" font-weight="600" text-anchor="middle">Postgres</text>
+          <text data-node-label="" data-detail-anchor="" x="602.4" y="94" class="t-primary" font-size="11" font-weight="600" text-anchor="middle">Postgres</text>
           <text data-detail="context" x="602" y="111" class="t-muted" font-size="7" text-anchor="middle">source of truth</text>
         </g>
 
--- a/archify/renderers/dataflow/render-dataflow.mjs
+++ b/archify/renderers/dataflow/render-dataflow.mjs
@@ -4,7 +4,7 @@ import { esc, renderDefinitions, renderSemanticSigil, textUnits } from '../share
 import { animateAttr, focusEdgeAttrs, focusNodeAttrs, focusNodeTitle, loadDiagramWithBrandMarks, writeDiagram, svgAccessibleText, svgRootAttrs } from '../shared/cli.mjs';
 import { throwDiagnosticProblems } from '../shared/diagnostics.mjs';
 import { resolveLegend, renderLegend as renderResolvedLegend } from '../shared/legend.mjs';
-import { availableNodeTextWidth, fittedNodeFontSize, minimumNodeTextWidth } from '../shared/text-fit.mjs';
+import { availableNodeTextWidth, fittedNodeFontSize, minimumNodeTextWidth, nodeTextWidth, sigilClearedLabelCenter, sigilLabelClearanceProblem, sigilLabelFitWidth } from '../shared/text-fit.mjs';
 import { brandLabelFitWidth, brandMetadataFor, brandTopRailProblem, renderBrandMark } from '../shared/brand-marks.mjs';
 import { translateMessage as i18nText } from '../shared/i18n.mjs';
 import {
@@ -143,6 +143,8 @@ function validateDataflow() {
     }
     const brandRailProblem = brandTopRailProblem(node, node.width, 8);
     if (brandRailProblem) problems.push(brandRailProblem);
+    const sigilClearance = sigilLabelClearanceProblem(node.id, node.label, node.width, 8);
+    if (sigilClearance) problems.push(sigilClearance);
     // sublabel and tag render as single unwrapped <text> elements; shrink-to-fit
     // handles the ordinary case, this rejects what it cannot rescue.
     const availableTextW = availableNodeTextWidth(node.width);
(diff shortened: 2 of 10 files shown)


tt-a1i (changes requested):
Thanks for documenting the narrowed scope and the remaining workflow decision. Reviewed exact head008c93eefc0a4e8e01d281f99ffd0b934645753b.

[P2] Keep the new sigil validation compatible with standard/v1 (renderers/dataflow/render-dataflow.mjs:146-147 and corresponding lifecycle/sequence calls). In product-analytics.dataflow.json, a node with width50 and label Ingester validates under standard on fixed main2bfb471, but this head rejects it because the new17px reservation leaves33px for a39px label. The guard runs with no quality-profile condition; omitted profile also regresses. Preserve the documented standard compatibility and exercise standard/showcase side by side. The new failure currently also has only generic layout/constraint, an imprecise subject, empty evidence and no supportedFixes; return actionable structured geometry diagnostics.

[P2] Rebuild the published ZIP after the source is settled. Seven payload files differ from canonical source, so ZIP installs retain the old renderer/fitting behavior. This is the archive gap already disclosed in the PR, not a claim that local geometric tests covered package delivery.

Focused sigil/layout tests101/101 passed. Fixed-main versus head public CLI reproduction confirms the compatibility failure; source/ZIP compared member-by-member. Workflow exclusion is explicitly disclosed and remains a scope/maintainer decision, not a separate hidden defect. No source edits or merge.

tt-a1i:
Thank you for the focused fix and regression work. We incorporated the sigil-clearance approach into #503, now merged to `dev` as `0b8b955`, while preserving your contribution in the commit history. The integrated path covers workflow, data-flow, and lifecycle and also handles cases where a narrow box needs a stacked title or a very short box needs a smaller decorative icon. The [before/after evidence and test record](https://github.com/tt-a1i/archify/blob/e608d6af18ae51852c52d91fa5a826849f7e1487/docs/reviews/issue-199/README.md) are available for review. This is on `dev`, not released yet. I appreciate you taking the original visual defect seriously.

Issue #227, opened 2026-08-31
Proposal: surface repository HTTP endpoints in architecture artifacts (api-inventory)


Changee-git:
## Problem

Architecture artifacts authored from repository evidence stop at component granularity. A reader asking "which HTTP endpoints does this service expose?" has to leave the artifact and grep the codebase. For real-world Spring codebases, the endpoint inventory is exactly the detail reviewers want next to the topology.

## Proposal

Add a post-delivery CLI step, run after `deliver` and before `visual-check`:

```
node bin/archify.mjs api-inventory <repository-path> <output.html> --json
```

- Scans Spring MVC controllers (`@RestController` / `@Controller` plus mapping annotations, class-level base paths, multi-path mappings, `RequestMethod`-style verbs) and groups endpoints by Maven/Gradle module.
- Injects into the delivered HTML: a fixed-position floating toggle plus a collapsed, module-filterable, searchable endpoint table (verb badges, monospace paths, controller, Javadoc summary). Inventory UI follows the artifact locale (`zh-CN` / `en`).
- The toggle is out of document flow and the table renders collapsed by default, so the first-screen containment contract is preserved: `visual-check` still passes at all four viewports after injection, including on `examples/web-app.architecture.json` whose adaptive reader resolves to a narrow 960px width. (An earlier draft injected a fourth Info Card — on narrow reader layouts the card row wraps and containment fails, which is why the toggle design was chosen.)
- Never modifies the artifact on failure: template-anchor checks reject unexpected layouts, non-Archify HTML is rejected, and reruns report `already-injected`.
- `diagnostics[]` contract throughout (stable codes, subject, evidence, supportedFixes). Non-Spring web frameworks (Express, Flask, Gin, Actix, …) report `unsupported-framework` instead of a silent no-op, so the agent tells the user honestly.

## Compatibility

No schema, renderer, or validator changes. `deliver` receipts unchanged. Purely additive CLI step; `SKILL.md` gains one section.

## Non-goals (for now)

- Non-Spring frameworks (detected and reported, not scanned).
- Diagram types other than `architecture`.
- Endpoint auth/permission metadata.

## Evidence

- `npm test` in `archify/`: 0 failures (desktop browser test passes with `ARCHIFY_CHROME` set).
- Regression tests in `test/api-inventory.test.mjs`: injection shape, idempotency, unsupported-framework, no-controllers, non-Archify rejection, usage errors, and a guard that injection must not add Info Cards.
[...]

PR #228, opened 2026-08-31
feat(cli): add api-inventory command for Spring MVC endpoint inventories

Changee-git:
Implements #227.

## User problem

Architecture artifacts authored from repository evidence stop at component granularity; readers cannot see which HTTP endpoints each service exposes without leaving the artifact and grepping the codebase.

## What changes

- New `bin/api-inventory.mjs` exporting `runApiInventory({ repoRoot, artifactPath })`, wired as an `archify api-inventory <repository-path> <output.html> [--json]` subcommand (same pattern as `visual-check`).
- Scans Spring MVC controllers (`@RestController` / `@Controller` plus mapping annotations, class-level base paths, multi-path mappings, `RequestMethod`-style verbs), groups endpoints by Maven/Gradle module.
- Injects into the delivered HTML a fixed-position floating toggle plus a collapsed, module-filterable, searchable endpoint table. The toggle is out of document flow and the table renders collapsed by default, so first-screen containment is preserved — an earlier draft that injected a fourth Info Card wrapped the card row on the narrow (960px) adaptive reader of `examples/web-app.architecture.json` and failed `visual-check`; the regression test guards against adding Info Cards.
- `SKILL.md` gains one "Repository API inventory" section: run after `deliver`, before `visual-check`, so the visual-check receipt covers the final artifact bytes.

- No schema, renderer, or validator changes; `deliver` receipts unchanged.
- `api-inventory` never modifies the artifact on failure (template-anchor checks, non-Archify HTML rejection) and reports `already-injected` on rerun.
- Non-Spring web frameworks (Express, Flask, Gin, Actix, …) report `unsupported-framework` via `diagnostics[]` instead of a silent no-op, so the agent can tell the user honestly.

Purely additive CLI step. Existing specs, artifacts, and workflows are unaffected.

Non-zero exit with a machine-readable receipt on every failure path; the artifact is left untouched. Rollback is simply not running the new step — injected artifacts remain valid standalone HTML.

## Tests and evidence

- `npm test` in `archify/`: **0 failures** (31 tests skipped without `ARCHIFY_CHROME`; `test/desktop-reader-browser.test.mjs` passes with `ARCHIFY_CHROME` set to a local Chromium).
- New `test/api-inventory.test.mjs` (6 tests): injection shape and payload, idempotency (bytes unchanged), `unsupported-framework`, `no-controllers`, non-Archify rejection (file preserved, exit 1), usage errors, plus the no-extra-Info-Cards containment guard.
- Final-artifact evidence: `deliver` → `api-inventory` → `visual-check` on `examples/web-app.architecture.json`: containment **pass** at 1440×900, 1600×1000, 1920×1080, 2048×1320 (Chrome available; visual review pending as always).

## Regenerated artifacts

- `archify.zip` rebuilt with Node 22 (77 files; includes the new `bin/api-inventory.mjs`; the distributable package intentionally does not ship tests).


archify/bin/api-inventory.mjs
```diff
@@ -0,0 +1,612 @@
...
+    });
+    return { exitCode: 0, receipt };
+  }
+  if (result.status === 'fail') {
+    receipt.diagnostics.push({ ...result.diagnostic, subject: { artifact } });
+    return { exitCode: 1, receipt };
+  }
+
+  const buffer = Buffer.from(result.html, 'utf8');
+  fs.writeFileSync(artifact, buffer);
```
tt-a1i: [P1] Commit the updated HTML atomically. writeFileSync truncates the trusted target before the new bytes are fully written. Injecting ENOSPC after writeSync writes 64 bytes makes runApiInventory throw and changes a delivered artifact from 715210 bytes to 64. This violates the explicit never-modify-on-failure contract and loses the user's diagram. Write a private same-directory temporary file fully, then rename; add a partial-write failure regression.

archify/bin/api-inventory.mjs
```diff
@@ -0,0 +1,612 @@
...
+    }
+    if (!/(?:@RestController|@Controller)\b/.test(source)) continue;
+    const classMatch = source.match(/public\s+(?:abstract\s+)?class\s+(\w+)/);
+    const controllerName = classMatch ? classMatch[1] : path.basename(filePath, '.java');
+    const head = classMatch ? source.slice(0, classMatch.index) : '';
+    let base = '';
+    for (const m of head.matchAll(MAPPING_RE)) {
+      if (m[1] === 'RequestMapping') {
+        base = extractPaths(m[3])[0];
+        break;
```
tt-a1i: [P2] Compose the complete class and method mapping conditions. A class @RequestMapping({"/v1","/v2"}) and method @RequestMapping(value="/users",method={RequestMethod.GET,RequestMethod.POST}) produces only GET /v1/users, losing three combinations. A class-level method=GET is itself emitted as GET /api/api while its /users handler becomes ANY /api/users. Both return ok:true. Keep all base paths/verbs, distinguish class annotations from handlers, and combine their conditions as required by #227.

archify/bin/api-inventory.mjs
```diff
@@ -0,0 +1,612 @@
...
+      out.push(full);
+    }
+  }
+}
+
+function extractPaths(annotationArgs) {
+  if (!annotationArgs) return [''];
+  const attr = annotationArgs.match(/(?:value|path)\s*=\s*(\{[^}]*\}|"[^"]*")/);
+  const segment = attr ? attr[1] : annotationArgs;
+  const values = [...segment.matchAll(/"([^"]*)"/g)].map((m) => m[1]);
```
tt-a1i: [P2] Only interpret value/path or positional shorthand as paths. For a class base /api with @PostMapping(consumes="application/json"), this fallback emits POST /api/application/json instead of POST /api. Other named annotation attributes must not be treated as endpoint paths. Add consumes/produces/params-only mapping coverage.

archify/bin/api-inventory.mjs
```diff
@@ -0,0 +1,612 @@
...
+    let source;
+    try {
+      source = fs.readFileSync(filePath, 'utf8');
+    } catch {
+      continue;
+    }
+    if (!/(?:@RestController|@Controller)\b/.test(source)) continue;
+    const classMatch = source.match(/public\s+(?:abstract\s+)?class\s+(\w+)/);
+    const controllerName = classMatch ? classMatch[1] : path.basename(filePath, '.java');
+    const head = classMatch ? source.slice(0, classMatch.index) : '';
```
tt-a1i, line 213: [P2] Exclude comments before scanning mapping annotations. A controller with // @GetMapping("/deleted") and one real @GetMapping("/live") currently advertises both endpoints with ok:true and no warning. Deleted/commented code is not exposed API surface. Tokenize sufficiently to ignore line/block comments and string contents; cover a commented mapping in a real controller.

archify/bin/archify.mjs
```diff
@@ -1203,6 +1204,52 @@ async function commandVisualCheck(args) {
...
+  try {
+    ({ runApiInventory } = await import('./api-inventory.mjs'));
+  } catch (error) {
+    fail(`Could not load api-inventory: ${error.message}`, 1);
+  }
+
+  let result;
+  try {
+    result = await runApiInventory({ repoRoot: positional[0], artifactPath: positional[1] });
+  } catch (error) {
```
tt-a1i, line 1271: [P2] Normalize all failure paths into the JSON diagnostic contract. api-inventory . . --json (directory as artifact) emits only an OS-specific error string without diagnostics[]. Adding --typo --json instead leaves stdout empty and prints prose. The issue/body require machine-readable failure receipts, and CONTRIBUTING requires stable code/subject/evidence/supportedFixes. Handle option validation and caught I/O errors through one receipt builder.

tt-a1i (changes requested):
Thanks for the contribution. Reviewed exact head 8416c9710dbea2c52f4ba1f26a2b95611b174c6c. The endpoint inventory has useful intent, but the current scanner can confidently publish incorrect API surface and the write path can destroy a delivered artifact on failure.

## Standards
Make injection atomic and keep every failure in the documented diagnostics[] receipt contract.

## Spec
Preserve Spring class/method mapping composition and distinguish actual path arguments/annotations from unrelated attributes and commented code.

## Verification
All six existing api-inventory tests pass and all 77 ZIP payload entries match canonical source. Additional deliver-then-inject fixtures reproduce the inline findings with ok:true and empty diagnostics. An injected ENOSPC after 64 bytes reduces an existing 715210-byte delivered HTML to 64 bytes. No source edits or merge performed; missing CI and base conflicts are separate gates.

PR #236, opened 2026-08-31
feat(renderers): keep edge labels inside the canvas

### LINKED ISSUE #235: Edge labels can leave the canvas with a clean showcase receipt, and repair hints can point outside it
## Problem

A diagram whose edge label overhangs the viewBox ships with the label clipped mid-word, yet `deliver --quality showcase` returns 9/9 checks, 0 errors, 0 warnings — the receipt approves a visibly broken artifact. The gap covers the fixed canvases: architecture, sequence, data-flow, lifecycle, and `schema_version: 1` workflow. (readable-v2 workflow is already safe: the compiler grows its canvas around pinned labels and rejects an authored viewBox that cannot hold the layout.)

Two adjacent defects compound it:

- `suggestLabelObstacleFix` derives its `labelAt`/`labelDx`/`labelDy` hints from the obstacle alone. It can propose coordinates outside the canvas — applying the validator's own suggestion produces a clipped label. With an authored `labelAt` present, the relative-form hint is a no-op (`labelPoint` short-circuits), and with a pre-existing `labelDx` the delta lands in the wrong place.
- `archify check` on an existing HTML artifact re-runs the composition rules, so the same clipped artifact also passes standalone checking.

Architecture's auto viewBox (`autoViewBoxFor`) sizes the canvas from components and boundaries only, so a no-`meta.viewBox` diagram can overhang the renderer's own canvas.

CJK text raises the odds of hitting all of this (full-width glyphs are 2 text units, so labels run wider), but the defect is locale-independent.

## Repro

Any showcase diagram with an edge label routed near the right edge and a long enough label: `validate --quality showcase` exits 0 while the emitted `<rect>` for the label extends past `viewBox[0]`. Minimal JSON available on request.

## Proposed fix (PR ready)

1. Where the canvas is derived, size it correctly instead of reporting the author: include connection label rects in architecture's auto viewBox (measured non-breaking — `max()` only grows a bbox, and every checked-in auto-sized architecture diagram renders byte-identically).
2. Where the canvas is authored (plus the origin side growth cannot reach), a showcase-level rule `composition/label-canvas-containment` reports the exact overflow with machine-readable evidence; `standard` keeps accepting pre-rule documents. Wired into the renderers and into `check` for artifacts they did not produce.
3. Repair hints only emit fixes that survive being applied: clamped into the canvas, measured against the document's own `labelDx`/`labelDy`, withheld when an authored `labelAt` outranks them, and stated as replacements, never increments.

All checked-in examples and frozen fixtures validate identically before/after (43 PASS / 10 pre-existing FAIL on both — zero legitimate diagrams newly rejected).

## Known adjacent gap (out of scope)

Sequence **segment** titles are still unbounded (a segment label hundreds of px past the viewBox validates clean) — a distinct geometry, left for a follow-up.

Rererr:
## Problem and value

Relationship labels could be clipped by the canvas while static artifact checks passed, and suggested label repairs could still place a two-line label on an obstacle. This change accounts for label bounds consistently in rendering, validation, standalone checks, and repair suggestions.

## Behavior and compatibility

- Automatically sized architecture canvases grow to include connection labels. Authored viewBoxes remain authoritative.
- Showcase reports `composition/label-canvas-containment` with a label rectangle, viewBox, overflow and repair guidance. Standard inputs remain accepted; readable-v2 workflow retains its existing compiler containment behavior.
- Standalone checking preserves positive and negative SVG viewBox origins.
- Repair hints use the actual label height/anchor offsets, check the supplied neighboring obstacles, and preserve authored offset/labelAt precedence.
- Sequence segment titles remain a separate geometry concern outside this change.

## Dev integration

Head `fc43792` preserves original `4e22f71` history and integrates current dev `24285de8` plus the reviewed #266 routing extraction (`c4ae7da`). It retains the current edge-label colors, authored direct-route checks, and bounded minimum-clearance reduction. Router initialization now precedes label measurement without copying routing back into the renderer. Three exact assertions were updated to the existing dev fixtures' current canvas/label geometry; the containment criteria were not weakened.

## Independent verification

- Original revised head `4e22f71`: 209 geometry/layout/output-check tests passed, no skips. Both prior review defects have regressions and are fixed.
- Combined integration source: 235 affected tests passed, no skips; `node test/golden.mjs` passed for all five modes' root and packaged examples plus schema/template/version checks.
- Paired current-base/candidate sweep: all 31 checked-in diagram documents selected under examples, gallery/cases and experiments retain their showcase validation exit result. Eleven architecture documents render successfully on both and are byte-identical; two MCO documents already fail both renders and are not claimed as passing artifacts.
- Real browser visual comparison at 1440×900, light/Classic: the same auto-canvas fixture changes from viewBox `580×282` with the label clipped to `735×282` with the entire label visible. Nodes/relationships/authored label coordinates are preserved; scaling changes as expected from the larger canvas. This is a focused visual review, not a full-device matrix or actual trial use.
- Additional delivery/package checks and final remote CI are recorded in the follow-up review; do not infer success from a pending run.

## Generated artifacts

Rebuilt `archify.zip` with official Node 22 from tracked combined source. Regenerated `docs/gallery.html` and `docs/gallery/manifest.json`; their only changes are the new zero-valued containment metric for each existing artifact. The golden examples and Gallery artifact HTML remain unchanged. No version/tag/release or main promotion.


archify/renderers/shared/geometry.mjs
```diff
@@ -1386,22 +1482,90 @@ export function formatRect(r) {
...
+ *   rect inside the canvas.
+ * The authored values therefore come from `labelRect.relation`, the authored
+ * relationship the renderer already attaches to every label record.
+ */
+export function suggestLabelObstacleFix(labelRect, lx, ly, obstacle, obstacleKind = 'component', viewBox) {
+  const [canvasWidth, canvasHeight] = asArray(viewBox);
+  const xOffset = labelRect.x - lx;
+  const yOffset = labelRect.y - ly;
+  const anchorX = clampAnchorToCanvas(lx, xOffset, labelRect.width, canvasWidth);
+  const placements = [
```
tt-a1i, line 1551: [P2] Derive above-obstacle placement from the actual label height/anchor offsets. In examples/product-analytics.dataflow.json, set flows[0].labelAt=[100,150]. The error suggests [100,124] (above), but applying it still overlaps web and repeats the same suggestion: label rect [67,113,66,27] overlaps node [44,128,112,58] by 12px vertically. Dataflow classification and lifecycle note labels are 27px, unlike the 14px regression cases. Reject candidates that remain on the original obstacle or compute the full rect correctly, and add round-trip tests for both two-line forms.

archify/scripts/check-render-output.mjs
```diff
@@ -618,9 +640,17 @@ function textBox(attrs, text) {
   };
 }
 
-function collectDesktopReadability(svgAttrs, fragment) {
+function viewBoxSize(svgAttrs) {
```
tt-a1i: [P2] Preserve min-x/min-y when checking the SVG viewBox. This helper retains only width/height, so the new containment calculation assumes origin zero. For viewBox="100 0 240 160", a marked label mask x=80,y=48,w=60,h=14 is clipped by 20px yet standalone check returns 9/9, zero warnings/errors, exit0. A contained mask x=280,y=48,w=40,h=14 is instead rejected as 80px right overflow. Pass the full viewBox or translate label rects by its origin; test positive and negative origins.

tt-a1i (changes requested):
Thanks for the detailed containment work. Reviewed exact head 3fa5231b32a096063453be896124e2d384be2d3b. Two cases still violate the stated repair/check parity contract.

## Standards
The standalone checker discards legal viewBox origins, producing both false negatives and false positives.

## Spec
The rewritten repair helper still emits non-executable above-obstacle fixes for 27px two-line relationship labels; this is an incomplete fix of the explicitly scoped issue, not a claim that the previous helper handled it.

## Verification
Geometry/layout-rules/render-output-checks: 202 passed, 0 failed. Both additional cases reproduced at this head. All 76 ZIP payload files match canonical source. No full-suite/browser/manual-visual acceptance or current-main integration claimed; no edits/merge.

Rererr:
Thanks for the review — both parity violations are fixed.

- **Standards**: the standalone checker now parses all four viewBox values and measures containment from the actual viewBox origin (6d9dc9f). Your repro now yields clipped → exit 1, contained → exit 0; zero-origin output stays byte-identical.
- **Spec**: repair hints are now derived from the rect of the label itself and round-trip filtered — a suggestion is only emitted when the applied rect passes the same overlap check used for detection (6d9dc9f), extended to the full obstacle set in b263289. The 27px two-line label case now converges (above [100,108] / below [100,201], both apply cleanly).

Full suite on official Node 22.23.2 (zlib 1.3.1): 1020 passed / 0 failed / 27 skipped (environment-gated). archify.zip was regenerated in the same environment and the byte-reproduction gate is green. b263289 also closes adjacent gaps in the same contract found in self-review (checker/render viewBox divergence for architecture boundary titles, hints landing on neighboring nodes).

Rererr:
Sorry for the force-push noise here — I didn't fully get how this repo handles `archify.zip`, so I kept rebuilding it and pushing over the branch.

I'll leave it as it is for now and rebuild the zip once this gets re-reviewed and approved.

archify/renderers/shared/geometry.mjs
tt-a1i: Verified fixed by original revision 4e22f71 and preserved in current-dev integration fc43792. The actual 27px label rect/anchor offsets determine above/below candidates, candidates are checked against the supplied blockers, and the dataflow/lifecycle round-trip regressions pass. 209 affected tests passed on the author revision; 235 pass on integration.

archify/scripts/check-render-output.mjs
tt-a1i: Verified fixed by original revision 4e22f71 and preserved in current-dev integration fc43792. Standalone check passes the full viewBox rectangle to containment; positive and negative origin cases now cover both contained and clipped labels. Both relevant regression suites pass.

tt-a1i (approved):
Reviewed final fc437920cb8352e0b1b3f6f28859e43769fc0dac. Prior nonzero-origin and two-line repair defects are fixed, and both threads are resolved. Combined dev integration passes 235 affected tests, five-mode golden output, and 106 delivery/CLI/Delta/package tests (two explicit platform/version skips). The 31-document paired validation sweep retains exit results; 11 architecture renders are byte-identical, two MCO documents already fail both. Focused real-browser comparison confirms the auto canvas exposes the formerly clipped label. Final CI35136918336 succeeded across the full matrix, shared browser/WebM, ZIP and platform jobs. #266 is now merged into dev with identical source to the integration parent. Approve for dev; actual trial-use acceptance and main promotion remain separate.

Issue #243, opened 2026-08-31
[Bug]: Decorative lifecycle rail stays at full opacity under focus/reach, showing a phantom arrow that dead-ends inside the final state


KarambaDev:
### Archify version or commit

2.16.0 (`skill-release.json`), reproduced on `main` @ 2bfb471

### Installation method

Downloaded archify.zip

### Diagram type

Lifecycle

### Exact command

```shell
node bin/archify.mjs deliver lifecycle repro.lifecycle.json repro.html --quality showcase --json
# then open repro.html and focus any state (click a node, or open #focus=a)
```

### Minimal redacted JSON reproduction

```json
{
  "schema_version": 1,
  "diagram_type": "lifecycle",
  "meta": { "title": "Rail repro", "quality_profile": "showcase" },
  "lanes": [{ "id": "main", "label": "Phases" }],
  "states": [
    { "id": "a", "type": "start", "label": "A", "lane": "main", "col": 0 },
    { "id": "b", "type": "success", "label": "B", "lane": "main", "col": 1 }
  ],
  "transitions": [{ "from": "a", "to": "b" }]
}
```

### Validation receipt or exact error

```json
{
  "schemaVersion": 1,
  "ok": true,
  "command": "deliver",
  "type": "lifecycle",
  "validation": {
    "checksPassed": 9,
    "checkCount": 9,
    "compositionProfile": "showcase",
    "compositionStatus": "pass",
    "errors": 0,
    "warnings": 0
  }
}
```

The receipt is clean — this defect is invisible to `validate`, `deliver` and `visual-check`, because it only appears once the Viewer enters focus or reach mode.

### Expected behavior

The decorative lifecycle rail is a background spine. In the static view it is deliberately hidden behind the opaque state boxes, and its arrowhead is tucked under the final state. When the Viewer dims the diagram for focus or Authored Reachability, the rail should dim along with everything else, so the only prominent marks are the focused node and its authored relationships.

### Actual behavior

The rail keeps `opacity: 1` while every node and edge drops to `0.13` (focus) or `0.09` (reach), so it becomes the most prominent element on the canvas. Two details make it read as a real, broken relationship:

[...]

PR #249, opened 2026-08-31
fix(delivery): stop delivered HTML from requesting Google Fonts

### LINKED ISSUE #242: Delivered HTML fetches Google Fonts, so an offline or air-gapped viewer silently loses the typography
A delivered page links two external hosts. Measured on 2.16.0, on a diagram with no `brand` fields at
all, by grepping the output file:

```
2  https://fonts.googleapis.com
1  https://fonts.gstatic.com
43 http://www.w3.org        <- the SVG namespace, not a request
```

So every viewer's browser makes a request to a third party in order to render the page as designed.

## Two consequences, and the second one is the one I care about

**Offline it degrades silently.** On a machine with no route to those hosts the page still opens, the
diagram still renders, and the typography falls back — so the artifact looks fine to whoever produced
it and different to whoever receives it. Nothing reports it. This is the normal case for an
air-gapped or restricted network, which for infrastructure diagrams is not an edge case: those are
exactly the environments people draw diagrams of.

**It is an outbound request per viewer.** A self-contained HTML file that reaches a third party on
open is a property worth being deliberate about, and the README's framing of the output as
self-contained is what made me check.

## The SVG export is already clean, which is what makes this fixable rather than inherent

Same diagram, exported to SVG from the viewer:

```
3  http://www.w3.org        <- namespace only
0  external hosts
```

and the font is declared as `local()` with a fallback, so it degrades to a system font by design
rather than by accident. The right behaviour already exists in one output path.

## Suggestion

Any of these would close it, in rough order of how much work they look like from outside:

1. a flag — `--embed-fonts` — that inlines the faces as `data:` URIs in the delivered HTML;
2. the SVG path's approach applied to the HTML: `local()` first with an explicit fallback stack, and
   the remote stylesheet as an optional enhancement rather than the only source;
3. failing both, a line in the README saying the delivered page requires network access for
   typography — that at least makes it a known property instead of a surprise.

Happy to test a branch; the check is one grep over the delivered file.

shahidbeig-a11y:
## Problem and value

Fixes #242

Delivered diagram HTML linked `fonts.googleapis.com` and `fonts.gstatic.com`, so air-gapped or offline viewers silently fell back to different typography with no diagnostic. The SVG export path already used offline-first `local()` faces; this applies the same contract to delivered HTML.

## Scope

- What changed:
  - Removed the async Google Fonts `<link>` tags from `archify/assets/template.html`.
  - Added offline-first `@font-face` rules that prefer locally installed JetBrains Mono and fall through to the existing system monospace stack.
  - Added `archify/test/offline-typography.test.mjs` to assert the template and `deliver` output request no Google Font hosts.
  - Regenerated the five golden HTML examples (`archify/examples/` + `examples/`), `examples/web-app.html` (template freshness), and `archify.zip` so `golden.mjs` and zip-freshness CI pass.
  - No schema, renderer geometry, viewer interaction, or CLI contract changes.
  - Gallery artifacts, README proof media, checkout delta, and other non-golden examples are **not** regenerated in this PR.
  - Issue #243 lifecycle-rail focus behavior is out of scope.

## Stability impact

- Compatibility and migration risk: low. Typography now prefers local JetBrains Mono when installed and otherwise uses the same fallback stack that already applied when Google Fonts were unreachable.
- Renderer, validator, package, or generated-artifact risk: template-owned HTML bytes change in golden examples and the distributable zip; topology, validation, and viewer behavior are unchanged.
- Failure behavior and rollback path: revert these commits to restore the previous remote-font behavior.

## Tests run

- `node --test test/offline-typography.test.mjs` — 2 passed, 0 failed.
- `node test/golden.mjs` — all checks passed (10 golden renders, 9 schema enforcement, 2 template freshness, 7 version sync).
- `scripts/build-zip.sh /tmp/archify-built.zip` then `cmp -s /tmp/archify-built.zip archify.zip` — match.
- `node bin/archify.mjs deliver architecture examples/web-app.architecture.json /tmp/issue-242-after.html --json` then `rg -c "fonts\.googleapis|fonts\.gstatic" /tmp/issue-242-after.html` → **0 matches**.

## Visual evidence

Not applicable. Font-loading contract only; diagram geometry and viewer behavior are unchanged.

**Before (template on `main`):** 3 Google Font host references (`fonts.googleapis.com`, `fonts.gstatic.com`).

**After (`deliver` output):** 0 external font host references; offline-first `@font-face` with `local()` sources.

## Generated artifacts

- `archify/examples/*.html` (5 golden examples)
- `examples/*.html` (5 golden examples + `web-app.html`)
- `archify.zip`

## Checklist


JINITAIMEI121:
Hi! I verified both open fixes for #242 locally (this PR and #256) to help move the decision along. Environment: macOS arm64, Node v24.19.0, checked out at head `f28069d` (base `06dd052`).

**What works**

- `node bin/archify.mjs deliver dataflow examples/product-analytics.dataflow.json out.html --quality showcase` → 9/9 artifact checks, showcase pass.
- The delivered HTML contains zero references to `fonts.googleapis.com` / `fonts.gstatic.com`. The remote request is gone — the core of #242 is fixed here.

**`npm test` currently fails at head (2 of 1028 tests)**

```
✖ generated proof gallery matches its sources, receipts, and checked-in artifacts   (archify/test/gallery.test.mjs)
✖ checked-in Checkout compare artifact is reproducible from its authoritative inputs (archify/test/architecture-delta.test.mjs)
```

I saw the PR body notes that gallery artifacts, README proof media, and the checkout delta are deliberately not regenerated here. The catch is that `test/gallery.test.mjs` and `test/architecture-delta.test.mjs` assert exactly those checked-in artifacts against the new template, so the suite is red as-is (baseline `main@06dd052` is 995 pass / 0 fail). The required `test (18/20/22/24)` checks are still "waiting for status", so this may not have surfaced yet — either those artifacts need regenerating in this PR, or the tests need a scope note, before it can merge green.

**One design question**

The replacement `@font-face` blocks are `local()`-only. The viewer measures real font metrics rather than assuming them, so delivered geometry would depend on whether each reader happens to have JetBrains Mono installed locally (and which version). #256's description measures a 960px vs 987px reader column with/without the face — I haven't independently re-measured that, but the mechanism looks real, and producer/receiver reproducibility is part of what these artifacts promise. Worth a deliberate decision: embed the subset, or accept and document the variance.

Happy to re-run anything.

sunsunsun-java:
Thanks for the work on this PR. #256 has now been merged, addressing #242 by embedding the viewer font in delivered artifacts and exports, without depending on Google Fonts requests or locally installed JetBrains Mono.
I suggest closing #249 as superseded for the delivery-font fix. The additional Gallery website typography changes can be considered separately if still desired. Thanks again for the contribution and validation!

JINITAIMEI121:
Thanks for the kind words and for handling the close-out so cleanly — this was exactly the outcome I hoped the verification would help reach: one merged fix (embedded, reproducible) instead of two competing ones. Happy to help verify future archify changes the same way if useful.

PR #254, opened 2026-09-01
feat(viewer): make Semantic Passport movable

### LINKED ISSUE #245: Make Semantic Passport manually movable with bounded pointer and keyboard controls
## Problem and user value

Semantic Passport is the reader's single proof surface for a focused node, but on desktop it is anchored to the left edge and only its vertical position is chosen automatically. In a dense or wide diagram, focusing a left-side node can place the Passport directly over the node and nearby authored relationships. The reader can close it, but cannot move it aside while retaining the selected node and its evidence.

This issue proposes a bounded, reader-controlled escape hatch: allow the existing Passport to be moved within the visible diagram container, without changing authored geometry, persisted application state, or canonical exports.

## Existing contract and precedent

- The Viewer Runtime defines Semantic Passport as a viewer-only surface outside canonical export.
- #98 / #100 added bounded dragging and collision recovery for Semantic Radar, but explicitly left Semantic Passport dragging out of scope.
- This proposal keeps that boundary explicit: it does not introduce a generic floating-window manager and does not change Radar behavior.

## Proposed behavior

Desktop/fine-pointer readers receive one small move handle in the Passport header:

- pointer drag starts only after a 3 px threshold;
- pointer capture keeps the gesture coherent outside the handle;
- movement is clamped to the diagram container and currently visible vertical area;
- a valid manual position survives switching the focused node;
- closing the Passport clears the manual position;
- double-clicking the handle or pressing `Home` restores automatic placement;
- arrow keys move by 16 px; `Shift` + arrow moves by 4 px;
- viewport/container resize reclamps the current manual position;
- narrow/mobile layouts hide the move affordance and retain existing automatic placement.

The handle should expose localized English and Simplified Chinese accessible names and a visible keyboard focus state.

## Compatibility and truth boundary

- No schema, typed JSON, renderer topology, node identity, relationship identity, or permalink change.
- No cross-page or local-storage persistence.
- No mobile drag surface.
- No automatic claim that manual movement resolves every overlap with Radar, Legend, or Dock.
- Semantic Passport remains viewer-only and must not enter canonical SVG, PNG, JPEG, WebP, Share Card, or WebM exports.
- Existing automatic placement remains the default and the fallback after reset.

## Alternatives considered

1. **Automatic avoidance only.** This can handle known Viewer obstacles, but cannot reliably infer what authored content the reader wants to keep visible.
2. **Reuse Radar's entire placement manager.** The two panels have different ownership and fallback behavior; extracting a general window manager would expand the change beyond this user problem.
3. **Make the whole header draggable.** This risks stealing clicks from close, relation, and evidence controls. A dedicated handle keeps the interaction explicit.

- If bounds cannot be measured, retain the current automatic placement.
- Cancelled pointer input restores the position from before that gesture.
- A resize clamps rather than discards a still-valid manual intent.
- Removing the handle, localized message, and bounded interaction block restores the current behavior without data migration.

## Acceptance evidence

Real-Chrome regression coverage should verify:

1. drag threshold and pointer movement;
2. clamping at all four edges;
3. focused-node switching preserves manual position;
4. close clears it;
5. `Home` and double-click reset it;
6. arrow and `Shift` + arrow movement;
7. resize and wide-diagram horizontal scrolling reclamp correctly;
8. mobile hides the affordance;
9. Radar, Legend, and Dock remain operable in combination;
10. canonical export clones contain neither the Passport nor its temporary state.

## Non-goals

- generic floating-panel infrastructure;
- drag inertia, snapping, docking, or saved preferences;
- changes to Semantic Radar;
- automatic relocation of authored nodes;
- a new mobile product surface.

## Environment used for the initial reproduction

- Archify `main` at `2bfb47132c057195d8dddb3e25ae966dd7c7a72e` (v2.16.0)
- macOS 14.5 (23F79)
- Node.js 24.14.1
- Google Chrome 151.0.7922.174

Existing issues and pull requests were searched for Semantic Passport move, drag, movable, and reposition variants; no duplicate was found.

sacloudy:
## Problem and result
A Semantic Passport can cover the part of a diagram a reader wants to inspect. Add one desktop move handle so readers can place it themselves without changing the diagram or export.

Dragging starts after 3px of movement. Arrow keys move 16px, Shift+Arrow moves 4px, and Home or double-click restores automatic placement. Manual placement remains bounded to the visible reader area, survives switching a single focused node, and resets on close. Narrow/coarse-pointer readers keep automatic placement. No persistent setting or new schema field is added.

## Integration and evidence
The original author commits are preserved, with the implementation ported into canonical `viewer/focus.js` and `viewer/template.source.html`. This branch also contains workflow reader integration #255; both target dev. Final combined head: `b6467a8`.

- 35 combined Semantic Passport, Focus, browser-gate and desktop-reader tests passed on runtime integration `0e1612a`, real Chrome, zero skipped. This includes the three/five-stage workflow reader regressions from #255.
- The original pointer test compared positions before the reader had stabilized. Waiting for `readerLayout.whenStable()` fixes that race without relaxing assertions. Desktop tests reuse the existing native Blink fine-pointer fixture, including its capability assertion, so Linux CI exercises dragging rather than silently taking the coarse-pointer fallback.
- Added the new suite to the shared required browser gate. Five-mode golden/schema/template/version checks pass.
- Independently clicked API Server and dragged its Passport in Ego at 1440×900, light/Classic. The card visibly moved from the left to the right and remained inside the reader. No claim of automatic obstacle avoidance is made for the user's manual position.
- Regenerated canonical Viewer, root/package examples, Checkout Delta, Gallery, README proof and Node-22 ZIP. Preserved Checkout's showcase generation profile and passed 34 Delta tests on the final artifact correction; runtime/package inputs did not change in that follow-up.

The first integration CI exposed a 17px close-button top gap against the existing 16px limit. Reduced header padding from 1rem to 0.9rem, retained the original assertion, and regenerated affected outputs. All 25 Passport/Focus tests pass on that correction, and the local complete WebM smoke passed, including dismissal, card exports and real video decoding (10 sampled frames, all distinct).

Final-head CI provides the remaining full-suite, platform/package, shared-browser and separate WebM decoding coverage. The focused export test verifies exclusion of the move affordance from export source; it is not by itself a decoded video check. See checks for current status.

This integrates into dev for trial use, with no main promotion or claim of real-task stability.

Fixes #245


tt-a1i (approved):
The bounded manual Passport move behavior is implemented in canonical Viewer sources, with native-desktop browser coverage in the shared gate. The layout-stability test race and close-button spacing regression are fixed without weakening assertions. Local focused tests and the complete decoded WebM smoke passed; final CI 35140353293 is successful across all jobs on b6467a8. Independent pointer interaction and limits are recorded in the description. Ready for dev integration and subsequent trial use.

PR #255, opened 2026-09-01
fix(workflow): fit implicit vertical stacks without inflating every lane

### LINKED ISSUE #250: workflow: per-lane height / tall lanes for a vertically-stacked group (one container, N stacked stages)
### Problem

A common workflow shape is a **container that runs several sequential stages** — e.g. a sandbox/cage running `stageA → stageB → stageC`. The natural drawing is those stages **vertically stacked inside ONE group box** (the container boundary). As of v2.16.0 this can't be expressed:

- `laneH` is a single global constant (104px); there is no per-lane height.
- A `group` box is always **one lane** tall (`laneH − laneTitleH − 16 ≈ 58px`) and cannot span contiguous lanes.
- A node may not overflow its lane (`node.y + height > laneTop + laneHeight(lane)` is a hard error). The per-lane `laneHeights` reserve added in v2.16.0 only grows a lane enough to contain a *single* node's modest `yOffset` inside its group — not to hold a stack — and the centering `(contentH − height)/2 + yOffset` fights large offsets.

**Result:** the only ways to draw N stacked stages in one container are (a) put them in one lane *horizontally* (loses the vertical "pipeline inside a box" reading), or (b) put them in N adjacent lanes with N separate group boxes (reads as N containers, which is semantically wrong for one cage).

### Minimal repro (fails on v2.16.0)

```json
{ "schema_version": 1, "diagram_type": "workflow",
  "meta": { "title": "stack", "viewBox": [720, 640] },
  "lanes": [ {"id":"cage","label":"one cage"} ],
  "groups": [ {"id":"g","label":"cage","lane":"cage","fromCol":1,"toCol":3,"variant":"security"} ],
  "mainPath": ["a","b","c"],
  "nodes": [
    {"id":"a","lane":"cage","col":2,"type":"security","label":"stageA","yOffset":0},
    {"id":"b","lane":"cage","col":2,"type":"security","label":"stageB","yOffset":90},
    {"id":"c","lane":"cage","col":2,"type":"security","label":"stageC","yOffset":180} ],
  "edges": [
    {"id":"e1","from":"a","to":"b","role":"main","fromSide":"bottom","toSide":"top"},
    {"id":"e2","from":"b","to":"c","role":"main","fromSide":"bottom","toSide":"top"} ] }
```

→ `Node "b"/"c" collides with the title or boundary of lane "cage"` (the lane never grows to hold the stack).

### Proposed options (either solves it)

1. **Per-lane height** — an optional `lane.height` (or auto-grow a lane to the extent of its nodes' authored `yOffset`s + heights), with `laneHeights[i]`, the overflow check, group-box height, and `autoHeight` all honoring it. Smallest change; composes with the existing v2.16.0 `laneHeights` machinery.
2. **Multi-lane groups** — allow a `group` to span `fromLane..toLane` (contiguous), drawing one box across them. Lets the existing N-adjacent-lanes layout render as one container without new height math.

Option 1 is preferred (one box, one lane, explicit vertical stack; matches how `yOffset` already reads).

### Note

Some downstreams vendor archify as a **verbatim committed subtree** (integrity = the subtree tree object matches `git rev-parse <ref>:archify`), so a local edit would fork that and break the provenance invariant — hence the request upstream rather than a local patch. Happy to send a PR for option 1 if the direction is agreeable.

zhouyuanxinand:
## Problem and result
Implicit schema-v2 workflows with vertical stacks can enlarge every lane and then overflow the desktop reader. Measure stacked lanes independently and allow the reader to fit these intrinsic diagrams by height while preserving the minimum projected text size. No new public lane-height field is introduced.

Explicit viewBoxes and absolute routing/label pins retain the prior shared-height behavior. Schema-v1 output and non-stacked geometry remain unchanged. The change is worthwhile for stacked workflow authoring without adding a schema contract or a general layout engine.

## Integration and evidence
Integrated the original author commits with dev `6f74d47c2bee5f053210dde65c5803e84339aa76`; Viewer changes live in canonical `viewer/` sources. Runtime integration head: `d895f23`; final artifact correction: `321ef6f`.

- 47 workflow/compiler and real-Chrome desktop-reader tests passed, zero skipped, on the combined source. Three- and five-stage stacks cover four desktop viewport sizes.
- Seven adaptive-reader tests passed. Five-mode golden, schema, template and version checks passed.
- The historical implicit order-overflow fixture now fits. Added an explicitly pinned version of that input to retain failing overflow-diagnostic coverage; the original explicit geometry remains authoritative.
- The schema-v2 SVG golden was compared directly with current dev: both produce SHA-256 `868547656078117182c273fa15532cd7d30707030866cc5f70aba72685b95814`. Updating the obsolete author-branch expectation does not accept a geometry change.
- Independently viewed the five-stage artifact in Ego at 1440×900, light/Classic: all stages and frame fit. This perceptual check is separate from automated readability coverage.
- Regenerated the delivered Viewer, canonical root/package examples, Checkout Delta, Gallery, README proof, and Node-22 canonical ZIP from combined source.

The first CI run exposed a missing `--quality showcase` flag when regenerating Checkout Delta. Corrected those two root example outputs and passed all 34 Delta tests; runtime/package inputs are unchanged in that follow-up.

Final-head CI supplies the remaining full-suite, shared-browser and platform/package coverage; see checks for actual status. This PR targets dev for integration and trial use; it does not promote main or establish real-task stability.

Fixes #250


ankittecham-arch (commented):
Reeta

ankittecham-arch:
Hello

sunsunsun-java:
Thanks for the PR. The overall direction is good: adding per-lane height is appropriately scoped and is preferable to modeling one logical container as multiple lanes.
However, I found a blocking containment issue in fixed-v1. With the original three-stage reproduction and lane.height: 442, compilation succeeds, but the last node ends at y=494 while the group ends at y=490. The node is therefore still rendered outside the group box. The current validation checks lane containment and logical group membership, but not whether member rectangles are actually contained by the rendered group frame.
There is also a default-equivalence inconsistency: explicitly setting height: 104 changes the fixed-v1 group height from 58px to 62px because the bottom inset depends on whether the property is present. Explicitly specifying the default value should produce the same geometry as omitting it.
Please:
- preserve the legacy fixed-v1 group inset regardless of whether height is present;
- validate authored-height v1 group members against the actual rendered group bounds and return an actionable structured diagnostic;
- add boundary regressions for insufficient height, exact-fit height, negative yOffset, and explicit height: 104;
- add a v2 test with height: 104 proving that the stacked lane auto-grows while an unrelated lane remains 104px;
- update from the latest main and rebuild the canonical ZIP from the final source;
- resolve the reused 2.16.0 release identity, correct the reported test count, run remote CI and the full npm test, and provide the required visual-review evidence.
Once these items are addressed, the proposed design should satisfy issue #250 without requiring a larger multi-lane-group redesign.

zhouyuanxinand:
Rebased this PR onto current `main` (`06dd052`) and resolved the binary package conflict by rebuilding `archify.zip` from the merged source with Node 22.

Updated head: `72ea590`.

Validated:
- `test/workflow-compiler.test.mjs`: 34 passed, including the Issue #250 v1/v2 lane-height regressions;
- generated validator and brand-mark freshness checks;
- golden source and packaged renders;
- extracted package smoke on Windows.

sunsunsun-java (dismissed):

archify/renderers/workflow/workflow-compiler.mjs
```diff
@@ -523,7 +559,9 @@ function createReadableLayout(workflow, layoutFeedback = {}) {
   });
   const groupHeaderHeights = groupLaneReserves.map(({ header }) => header);
   const groupFooterHeights = groupLaneReserves.map(({ footer }) => footer);
-  const laneHeights = groupLaneReserves.map(({ header, footer }) => laneH + header + footer);
+  const laneHeights = groupLaneReserves.map(({ header, footer }, index) => (
```
tt-a1i, line 572: **Blocking visual acceptance issue:** the new per-lane height can produce a tall portrait viewBox while the Viewer still expands it primarily from the available width. The canonical Issue #250 three-stage stack passes `validate` and `deliver`, but `visual-check` reports vertical overflow at every required desktop viewport; at 1440x900 the page is 1231px tall and Stage C is below the first screen. Please make the resulting artifact height-aware (Viewer fit or intrinsic aspect ratio), add a browser regression for the exact three-stage repro, and attach corrected light/dark screenshots plus a passing four-viewport `visual-check` receipt.

tt-a1i (changes requested):
Thanks for addressing the earlier fixed-v1 group-containment and default-equivalence issues. The code-level boundary tests now look solid, but the feature still fails its intended browser acceptance case.

Using the Issue #250 shape as a readable-v2 workflow — one lane with `height: 104`, one group, `stageA -> stageB -> stageC` in the same column with `yOffset: 0 / 90 / 180`, bottom-to-top edges, and no explicit `meta.viewBox` — `validate` passes 9/9 and `deliver` passes with 0 errors / 0 warnings. However, `visual-check` fails every required desktop viewport due to vertical overflow:

- 1440x900 -> scrollHeight 1231 (Stage C is below the first screen)
- 1600x1000 -> scrollHeight 1359
- 1920x1080 -> scrollHeight 1359
- 2048x1320 -> scrollHeight 1377

This is the exact product shape the PR is meant to unlock, so artifact validation alone is not sufficient. Please make the tall-lane output fit the required desktop viewports without clipping or an internal scroller, either through height-aware Viewer scaling or an appropriate intrinsic canvas/aspect-ratio solution.

After fixing it, please regenerate the same three-stage artifact and attach:

1. a corrected 1440x900 light screenshot corresponding to the currently failing view, with all three stages and the full group visible;
2. the matching dark screenshot and 2048x1320 evidence;
3. the `visual-check --json` receipt showing no overflow at 1440x900, 1600x1000, 1920x1080, and 2048x1320;
4. a browser regression covering this exact Issue #250 case;
5. a rebuilt canonical `archify.zip` and green exact-head CI.

Please keep `deliver`, automated browser containment, and perceptual visual review as separate evidence. Once the corrected artifact and receipt are available, I will re-review the new exact head.

tt-a1i (changes requested):
Thanks for addressing the earlier tall-workflow Viewer issue. I re-reviewed exact head `95635b1700199afe9afe4869bf549f9067111105`, and the previous desktop-overflow finding is resolved in the tested reproduction.

On macOS with Node 22.19.0 and real Chrome, the following suites passed together: `workflow-compiler.test.mjs`, `v1-compatibility.test.mjs`, `workflow-compiler-hard-contract.test.mjs`, `visual-check.test.mjs`, and `desktop-reader-browser.test.mjs` — 152 passed, 0 failed, 0 skipped. This includes both the three-stage and five-stage browser regressions.

I also independently delivered the three-stage Issue #250 fixture and ran `visual-check --json`. At 1440x900, 1600x1000, 1920x1080, and 2048x1320, the page has no horizontal or vertical overflow, no internal diagram scroll range, and no clipped nodes or group frames. The receipt passes with no diagnostics. I inspected the 1440x900 light and dark screenshots: all three stages and the complete group are visible. There is still substantial blank space above the stack; this is a composition observation, not a renewed overflow blocker or a claim of complete perceptual acceptance.

One packaging defect still blocks merging:

**[P1] Preserve the CLI executable bit in `archify.zip`.** The committed archive records `archify/bin/archify.mjs` as `100644`, while rebuilding the same head with Node 22.19.0 produces `100755`. All 77 archived file contents match; the differing entry metadata is this executable mode. Consequently, `cmp` against the canonical rebuild fails, and the extracted entry point lacks the Unix executable bit. Please regenerate the archive with the tracked executable mode preserved and verify direct execution of the extracted CLI, in addition to Node-prefixed invocations.

Please sync with current `main`, resolve the binary `archify.zip` conflict by rebuilding from the final sources, verify byte-for-byte package freshness, and rerun the relevant checks and full suite. CI on this head is currently `action_required`, with no passing remote checks. Update the PR description with the current scope and validation evidence, including the Viewer and visual-check changes.

I did not run the full local `npm test` suite in this re-review. The earlier overflow objection is superseded by the successful browser evidence above; the remaining change request is for the package defect and outstanding integration/CI gates.

zhouyuanxinand:
Addressed the latest P1 at `572c3b7`.

- Merged current `main` (`d8e4daf`) and resolved the binary conflict by rebuilding `archify.zip` from the final sources in WSL2.
- The rebuilt 77-entry archive is byte-for-byte fresh; its `archify/bin/archify.mjs` entry records Unix mode `100755`. After extraction, the CLI is mode `755` and both direct execution and `node`-prefixed `doctor` succeed.
- Added a Unix package regression for the extracted executable mode and direct CLI invocation.
- Final WSL2 `npm test`: 1054 tests, 1050 passed, 0 failed, 4 conditionally skipped; this includes the Issue #250 browser and visual-check regressions.

The PR description now follows the repository template and separates automated/browser evidence from perceptual review. GitHub Actions was triggered for this head but is `action_required` before any jobs are created, so maintainer approval is still needed to run remote CI. Please re-review after that gate is available.

zhouyuanxinand:
Resolved the approved-run failures at exact head `a82c3e2`.

Root cause: the previous WSL2 packaging preparation preserved `archify/bin/archify.mjs` as executable, but also accidentally recorded the other 76 archive entries as `100755`. That made both `zip-freshness` and the Node 22 archive reproducibility test fail; Node 24 was cancelled by matrix fail-fast rather than failing independently.

The replacement archive was built from a real WSL2 Git clone of head `572c3b7`, so it now contains exactly one `100755` entry (the CLI) and 76 `100644` entries. Its SHA-256 is `090047ac9ef5b45b78f7695e152c2a85477010b0431ffe0bd3c5165d3d5da4a7`.

WSL2 validation for the corrected artifact:

- targeted package gates: 21 tests, 20 passed, 0 failed, 1 conditional skip;
- full `npm test` with real Chromium: 1054 tests, 1050 passed, 0 failed, 4 conditional skips;
- exact rebuild comparison: passed;
- extracted CLI: direct and Node-prefixed `doctor` both passed.

The new exact-head workflow run is currently `action_required` before jobs can start. Please approve that run and re-review this head.

zhouyuanxinand:
Synced this branch with current `main` (`ed7f4d4`) in exact head `ab4908b`.

The only merge conflict was the binary `archify.zip`; I rebuilt it from the final combined source rather than choosing either side. Two canonical Node 22 builds byte-match (`CB9483FB06550FCEBDFCC1C2FBA95247400E057C32B6CA1D2D6FE807FC92A76D`), with 77 entries and only `archify/bin/archify.mjs` marked `100755`.

Exact-head validation includes:

- isolated real-Chrome desktop reader suite: 3/3 passed, including both Issue #250 workflows at all four viewports;
- all 149 non-browser workflow/compatibility/visual-check tests passed;
- merged #322 output-type suite: 9 passed, 0 failed, 4 Windows symlink-permission skips;
- extracted zero-dependency CLI smoke: 2/2 passed, including direct execution;
- golden, generated-validator, brand-mark, release-identity, and Gallery freshness checks passed.

GitHub now reports the PR as mergeable with no content conflicts. Please approve/run the fork CI for `ab4908b` and re-review this exact head.

tt-a1i (commented):
I compared current main `ed7f4d4` with this PR at `ab4908b`, using the public CLI, real Chrome, and the delivered screenshots. The results narrow the product need, so I would like to pause merging and agree on scope before asking for another implementation round.

First, the earlier containment and Viewer fixes are acknowledged. The archive-mode defect was also verified fixed at `a82c3e2`. This comment is a design/scope discussion, not a renewed report of those resolved defects or an approval of all integration checks on the latest head.

What the comparison showed:

- The single-container, three-stage vertical workflow is already possible on main: use schema v2, omit the fixed `meta.viewBox`, and center the authored `yOffset` values at `-90 / 0 / 90`. This preserves the group, stage order, spacing, and vertical connections. `deliver` passed 9/9 checks, and the delivered artifact passed browser checks at 1440x900, 1600x1000, 1920x1080, and 2048x1320. I also inspected the screenshot. The original v1 JSON still fails unchanged on both main and this PR.
- The clear new benefit is independent sizing in a mixed-lane workflow. In our three-lane reproduction, main produced lane heights `[276, 270, 270]` and a `768x1032` canvas. Setting the stacked lane's `height` to `104` in this PR produced `[276, 104, 104]` and a `768x700` canvas, reducing canvas height by about 32%. Main overflowed the tested desktop viewports; the PR artifact passed all four. This is a useful, demonstrated improvement.

My proposed direction is to focus on **independent automatic lane sizing in v2**, rather than adding an authored-height API to solve the original single-lane example:

1. Document the existing v2 solution for the original example, including the fact that `yOffset` is relative to the lane's content center.
2. Retain the independent-measurement work, but evaluate a v2 layout change without introducing `lane.height` in this iteration. Currently `height: 104` acts as both a minimum and an opt-in to a different sizing policy, and can also enable Viewer fitting. That is a broader public contract than a height value alone suggests.
3. Establish the compatibility boundary before changing the default v2 behavior. Independent sizing can change existing geometry; explicit offsets, routes, fixed viewBoxes, and checked-in examples need before/after evidence. We should not silently reinterpret authored constraints to obtain a smaller canvas.
4. Base Viewer fitting on the measured artifact and readability constraints, rather than the presence of an authored height field. Keep this work only as broad as the demonstrated layout need requires.
5. Leave fixed-v1 unchanged for this feature. The general visual-check improvements for scrolling, clipping, and theme verification are useful independently and can be proposed separately.

If a real use case requires authors to specify a minimum lane height, we can evaluate that API separately with a single, explicit meaning.

Please start with feedback on this scope and the compatibility tradeoffs; there is no need to do another broad rewrite before we agree on the direction. We should have made this product-scope decision earlier in the review process. Passing tests and resolving the implementation findings are valuable, but we also need to agree on the public behavior we are committing to maintain before merging.

zhouyuanxinand:
Addressed the latest review in `d9c2000` and merged current `main` (`1072200`). GitHub now reports the PR as conflict-free / mergeable.

What changed after the review:

- Removed the proposed public `lanes[].height` field. It remains rejected by the schema.
- Left fixed-v1 behavior unchanged, including the original stacked-source failure and the exact 700×400 compatibility geometry.
- Kept independent lane measurement only for readable-v2 workflows that have an implicit canvas and a real same-lane/same-column vertical stack (distinct `yOffset` values). Absolute authored pins (`via`, `labelAt`, `channelX`, `channelY`) and explicit `meta.viewBox` retain the existing shared-height v2 geometry. Route presets remain compiler-owned and can use independent measurement.
- Documented the existing v2 representation with centered offsets (`-90 / 0 / 90`) and clarified that `yOffset` is relative to the lane content center.
- Made Viewer fitting depend on compiled/measured SVG geometry (`data-reader-fit="intrinsic-height"` when a measured lane exceeds the 104px baseline), not on the presence of an authored sizing field. The Viewer preserves canonical SVG geometry and enforces a 6px projected node-text floor.
- Removed the unrelated general `visual-check` changes from this PR.

Compatibility evidence now locked by tests:

- Current-main shared layout: lane heights `[276, 270, 270]`, canvas `768×1032`.
- This branch's independently measured layout: `[276, 104, 104]`, canvas `768×700` (about 32% less height).
- Explicit viewBox and absolute-route-pin fixtures remain `[276, 270, 270]`.
- The checked-in non-stacked v2 example preserves its exact SVG SHA-256; fixed-v1 has byte-for-byte regression coverage.

Validation:

- WSL2 full suite exercised 1,096 tests: all behavior tests passed; its sole initial failure identified that `archify.zip` had been built with the Windows rather than canonical Linux toolchain. I rebuilt it with `scripts/build-zip.sh` under WSL2, and the exact cross-time-zone archive gate now passes (1/1).
- Workflow compiler suite: 42/42 passed.
- Checkout artifact + README proof reproducibility: 27/27 passed under WSL2.
- Real Chrome desktop Viewer regressions: 4/4 passed, including the 3-stage and 5-stage stacked workflows.
- Extracted final ZIP package smoke: passed on Linux.
- Final `archify.zip` SHA-256: `690b6df7dc90235025a089d8bb9157dec6c0aa1b1ecaade93b33faae8ff109c8`.

Ready for re-review.

archify/renderers/workflow/workflow-compiler.mjs
tt-a1i: Addressed by the revised intrinsic height-aware reader and integrated in d895f23. Independent real-Chrome coverage now includes exact three-stage and five-stage reproductions at all four desktop viewport sizes; the combined 47 compiler/reader tests pass with zero skips. I also viewed the five-stage result at 1440x900 in Ego. Explicitly pinned geometry keeps prior overflow diagnostics. Current-head CI is still running; local evidence and regenerated outputs are recorded in the rewritten PR description.

tt-a1i (approved):
Reviewed the revised intrinsic workflow measurement and height-aware reader at 321ef6f. The original visual blocker is resolved; local compiler/reader, pinned-overflow diagnostic, compatibility and perceptual checks are recorded in the PR. The Checkout regeneration mistake was corrected and all 34 Delta tests pass. Final CI run 35139768025 is successful across all jobs, including the required shared browser gate. Ready for dev integration and later real-task trial use.

PR #257, opened 2026-09-01
fix(renderers): measure data-flow stage headers

276970789:
Related to #235 / #236. Those cover edge and connection labels leaving the canvas; this addresses a separate data-flow structural-text path — **stage headers**, which never reach `renderers/shared/text-fit.mjs` at all. #236 also touches `render-dataflow.mjs` (+9/−1) and is currently conflicting; this change sits in different functions and I am happy to rebase behind it.

## Problem and value

`text-fit.mjs` opens by naming the failure mode it exists to close:

> Left unmeasured, an over-long value silently spills across its neighbours while validation still reports a clean receipt — the failure mode this module exists to close.

That protection is wired to node text only. A stage header belongs to no node, so `render-dataflow.mjs` emitted it at a literal `font-size="9"`, centred on a frame fixed at `stageW: 168` with `colGap: 215`. No validator rule measured it, and `stages[].label` has no `maxLength`. Two failures follow in one diagram, measured in real Chrome through this repo's own `ChromeVisualBrowser` harness after `document.fonts.ready` and `readerLayout.whenStable()`:

1. **A header leaves its frame and the canvas.** At 53 text units it needs 286.0px inside a 168px frame and, centred, starts at `x = −43.0` — outside the `viewBox`, so `01 / Str` is clipped away.
2. **Two adjacent headers overlap by 9.0px.** Header 0 ends at 243.0, header 1 starts at 234.0; `02 /` prints over the `on` of `Deduplication`.

Both with `validate`, `deliver --quality showcase` and `check` returning 9/9, composition `pass`, zero warnings.

The sharpest evidence is a contrast, not a screenshot. Put `"Stream Processing, Enrichment, and Deduplication"` in `nodes[0].label` and it is rejected with a precise measurement:

```
Label "…" (~298px) is wider than node "ingest" (112px) — shorten the label or increase node.width.
```

Move the identical string to `stages[0].label` and everything passes. The capability was already there, just not wired to text with no owning node — and the width model was already right: measured advance is **5.3964 px/unit at 9px = 0.5996 em** against `nodeTextFit.widthFactor = 0.6`, 0.07% apart. Nothing had to be re-derived.

## One question before you merge

This introduces a rule that can reject schema-v1 input which previously validated, in `standard` as well as `showcase`. CONTRIBUTING treats that as a contract change, so I would rather you choose the shape than assume it:

- **(a) as submitted** — mode-independent hard error, matching the existing node label / sublabel / tag rules in this same function. `validateDataflow()` takes no profile argument, so every text rule it owns already behaves this way.
- **(b) shrink everywhere, error only in `showcase`** — needs a quality profile threaded into `validateDataflow()`, which is a larger change to the layout-validation contract than this PR. Say the word and I will do it as a separate PR.
- **(c) warn only** for now, error later behind a version bump.

I implemented (a) because it is the only option that matches existing precedent in the file, and because 38 units is generous — the longest header anything shipped uses is 21. But it is your call, and I will switch without argument.

## Scope

`archify/renderers/dataflow/render-dataflow.mjs` (+34/−1), `archify/test/layout-rules.test.mjs` (+151/−0), and a rebuilt `archify.zip`.

- `nodeTextFit` gains `stageHeaderPreferred: 9` / `stageHeaderMinimum: 7`. These per-field sizes live in the renderer because `text-fit.mjs`'s own header comment says that is where they belong.
- `stageHeaderText(stage, index)` returns `` `${NN} / ${stage.label}` `` so the 5-unit ordinal prefix cannot drift between what is measured and what is drawn.
- `stageHeaderFontSize` wraps `fittedNodeFontSize(header, layout.stageW, 9, 7)`; `renderStage` emits that instead of a literal `9`.
- `validateDataflow` gains a stage loop shaped like the neighbouring sublabel/tag rules: `minimumNodeTextWidth(header, 7)` against `availableNodeTextWidth(168)` = **160px**, rejecting with `Stage header "…" needs ~Npx at the 7px legible minimum, but the stage frame provides 160px — shorten stages[N].label.`

**The two halves close different failures**, which caught me out while testing:

- **Frame overflow** is closed by the render half (shrink-to-fit).
- **The header-vs-header collision is closed by the validation half, arithmetically.** Every admissible header is at most 160px wide once fitted, so the worst pair gives `(w_a + w_b) / 2 = 159.6px` against a 215px `colGap` — 55.4px of clearance. No admissible pair can overlap.

The render half does **not** close the collision. I verified that by reverting it: a test that renders a long pair and asserts `gap > 0` still passed, because the pair I first used was never wide enough to collide. The test now asserts the invariant and pins the cap instead.

**Deliberately unchanged:** stage frame widths — adaptive column widths would change global layout and belong in a separate design decision, not this PR. Also checked and left alone: `render-lifecycle.mjs` band titles (schema caps `/lanes` at 4 and titles are left-aligned with 326px of slack — not a defect) and `render-sequence.mjs` segment labels (real coefficient mismatch, mask plate sized `units * 5.2 + 14` against an actual 5.3964/unit, but I could not produce a collision at 55, 80 or 82 units — and it is the case #235 already names). No `CHANGELOG.md` entry: CONTRIBUTING reserves release identity for maintainers.

## Stability impact

Below 30 units nothing changes: fitting is a no-op, so existing diagrams render exactly as before. No schema, CLI, receipt or typed-JSON shape changed. The new failure names the offending `stages[N].label` and the width it needs, matching the remedy path users already know from node labels.

`grep -rl 'data-composition-frame-kind="stage"' --include='*.html'` finds exactly four checked-in files with a data-flow stage frame:

```
archify/examples/dataflow-product-analytics.html
examples/dataflow-product-analytics.html
docs/gallery/artifacts/product-analytics.dataflow.html
docs/gallery/artifacts/event-stream.dataflow.html
```

All four are **byte-identical**: regenerated on this branch and on an unmodified worktree at the same base commit, `cmp` clean, and `diff -r` of the two whole `build-gallery.mjs` output trees is empty. The longest shipped header is 21 units (`04 / State + recovery`). `assets/template.html` is untouched.

**Regenerated files:** `archify.zip` only. CONTRIBUTING requires a rebuilt archive when a renderer change alters packaged contents. Built with `scripts/build-zip.sh` on Node 22.23.2; the only entry whose CRC changes is `render-dataflow.mjs`. I confirmed the toolchain is canonical by rebuilding on unmodified `main` first and getting byte-identical output to the committed archive.

## Tests

```
$ node --test test/layout-rules.test.mjs
# tests 103   # pass 103   # fail 0
```

Five cases added: the rejection message for a 53-unit header; shipped headers still render at 9px (so artifacts cannot drift); a 37-unit header shrinks below 9px, stays ≥7px and **actually fits** 160px; no admissible pair can overlap, asserted as the invariant rather than by sampling one pair; and the 38-unit cap renders while 39 is rejected, with the cap derived from the frame width and the floor rather than hardcoded.

Each half is genuinely covered — I reverted them one at a time:

| reverted | result |
| --- | --- |
| render half (`font-size="9"` restored) | 1 fail |
| validation half (the stage loop's `push`) | 3 fail |

Full suite from `archify/`, this branch versus an unmodified worktree at the same base commit:

| | tests | pass | fail | skipped |
| --- | --- | --- | --- | --- |
| this branch | 1027 | 994 | 2 | 31 |
| base, unmodified | 1021 | 988 | 2 | 31 |

The 2 failures are **pre-existing, not mine**: both in `test/clean-skill-staging.test.mjs`, both `ERR_FS_EISDIR` from `fs.rmSync` on a `symlink-probe` path the test creates, both failing identically on the unmodified worktree. It is a Node 25 `fs.rmSync` change; the file passes 6/6 on Node 22, so CI should be green.

Boundary, by CLI. Model and Chrome columns are kept separate — they are two estimates of the same thing, and I am not offering their agreement as corroboration:

| units | `font-size` | model | Chrome | result |
| --- | --- | --- | --- | --- |
| 37 | 7.2 | 159.84px | 159.7px | renders (was 199.7px in a 168px frame before) |
| 38 | 7 | 159.60px | 159.5px | renders, at the floor |
| 39 | — | — | — | rejected: `needs ~164px … provides 160px` |

38 clears 160px by only 0.4px in the model — the cap falls out of the 7px floor rather than being chosen, so a future font wider than 0.6 em would push it over.

## Visual evidence

**Visual review: passed.** Before/after on a 37-unit header — above the point where it leaves its frame, below the point where it is now rejected. Same fixture, same command, base versus this branch, cropped to the header band; the faint vertical rules are the frame borders.

Before — starts left of its frame border and runs past the right one:

stage header overflowing its frame on both sides before the change

After — shrunk to 7.2px, inside the 160px available, both borders clear; the other headers stay at 9px:

the same stage header shrunk to fit inside its frame after the change

The 53-unit case has no "after" picture because it no longer renders:

```
$ node bin/archify.mjs validate dataflow minimal.dataflow.json
Stage header "01 / Stream Processing, Enrichment, and Deduplication" needs ~223px at the
  7px legible minimum, but the stage frame provides 160px — shorten stages[0].label.
```

Its pre-fix behaviour, from a 694-byte fixture written from scratch — short labels as the control, then the defect:

data-flow stage headers with short labels, evenly spaced inside their frames

first stage header clipped at the left edge and printing over the second header

**One confound, disclosed rather than buried.** `visual-check` reports `status=fail` with 6 `viewer/viewport-overflow` findings on this artifact. Pre-existing and unrelated: the unmodified example and the repo's own checked-in `examples/dataflow-product-analytics.html` report the identical 6 on clean `main`. They are vertical-scroll findings, and `visual-check` measures no text geometry, which is why it never caught this defect. So the accurate claim is that schema/layout validation, `deliver` and static `check` pass and `visual-check` matches baseline — **not** that all gates pass.

## Checklist


tt-a1i (changes requested):
The stage-header overflow is worth fixing: the proposed measurement reuses the existing text-fit model, and fitting the complete ordinal-plus-label string avoids drift. The author-provided Chrome measurements and boundary tests make the intended behavior reviewable.

For the scope question in the description, use option **(b): shrink-to-fit in both profiles, and reject an unfit header in `showcase`**. Preserve `standard` acceptance and emit a structured warning there when the header still cannot fit. This follows the current contribution contract's broad compatibility for `standard`; it avoids turning this rendering repair into a new unconditional schema-v1 rejection. This is a maintainer scope decision, not a claim that the submitted arithmetic is wrong.

Please refresh against `dev` and use the renderer's current effective quality profile (including the public CLI override and direct compiler options). Give the failure/warning a stable diagnostic code, `stages[index].label` subject, measured available/required width, and supported repairs that preserve the author's required meaning. Keep the existing fit/boundary cases, and add public-command coverage showing the same overlong input remains accepted with a warning in `standard` and fails in `showcase`; short shipped labels should remain unchanged.

Regenerate the canonical ZIP from the final source and attach current-head CI. The previous author-provided visual results are useful evidence at their recorded revision; I have not independently repeated that visual review. Disposition: keep open, author revision requested for this bounded compatibility adjustment.

tt-a1i:
Thank you for identifying the separate stage-header measurement gap. We incorporated your measurement approach into #503, merged to `dev` as `0b8b955`, and retained your contribution in the commit history. The integrated implementation includes the stage ordinal in measurement, wraps long headers when there is clearance above nodes, and keeps the compatible single-line layout in tightly constrained geometry rather than introducing a new failure. The [matched-input evidence and regression record](https://github.com/tt-a1i/archify/blob/e608d6af18ae51852c52d91fa5a826849f7e1487/docs/reviews/issue-199/README.md) are public. This has not been released yet.

PR #266, opened 2026-09-01
refactor(architecture): extract connection routing into routing.mjs

AzScep:
## Problem and value

Architecture routing was embedded in a CLI renderer with top-level I/O and exit behavior. Extracting it lets internal callers inspect routed geometry without running the whole renderer or duplicating routing logic. This is a private module extraction; it adds no automatic-layout feature, schema, or new routing policy.

## Implementation and compatibility

`createRouter(components, connections)` owns the existing per-scene port spread and path cache and returns the existing routing functions. I compared the extracted routing block with current dev `24285de8`: function bodies are identical apart from passing `connections` instead of closing over `arch.connections`. A short module comment documents that callers create a new router when scene geometry changes.

Final integration head: `c4ae7daa35496f467745ee2cc820c79e2c617182`. The original `e7dd6f4` history is preserved. The only changed source files are the renderer and new routing module; `archify.zip` was rebuilt with official Node 22 from the combined tracked source. Existing Viewer/rendering changes on dev remain present.

## Verification

- Local Node 22: 222 geometry, layout, output-checker and release-package tests passed; 0 failed; 2 explicit platform/version-gated skips.
- `node test/golden.mjs`: passed. Five modes' root and packaged examples remain byte-identical, plus schema, template and version checks.
- Source comparison: current-dev routing body and extracted body match after the argument substitution and whitespace normalization.
- Final CI: https://github.com/tt-a1i/archify/actions/runs/35136472913 (check live result).
- No new visual behavior; no new perceptual review or actual trial-use acceptance claimed. Targets dev only.


tt-a1i (approved):
Reviewed final c4ae7daa35496f467745ee2cc820c79e2c617182 against dev24285de. The routing body is unchanged after the argument substitution, all five golden modes remain byte-identical, 222 focused tests pass with two explicit platform/version skips, and the canonical ZIP is fresh. Final CI35136472913 succeeded, including the shared real-browser gate and platform/package jobs. No unresolved review threads. Ready for dev integration; main promotion and actual trial-use acceptance remain separate.

PR #267, opened 2026-09-01
fix: ship third-party mark notices

tt-a1i:
## Summary
- add a repository and packaged `THIRD_PARTY_NOTICES.md` for bundled brand marks
- record the pinned Simple Icons source, all eight individual licenses present in 16.28.0 metadata, the OpenAI source, and the no-endorsement / no-additional-rights boundary
- fail package staging and smoke checks when the notice is absent, changed, or incomplete
- advance the development identity to `2.17.0-dev.1` and rebuild the deterministic ZIP and generated artifacts

## Boundary
This documents provenance and known terms. It does not claim that every bundled mark has been cleared for every commercial, promotional, or redistribution use.

## Verification
- Node 22 `npm test`: 1024 tests, 997 passed, 27 skipped, 0 failed
- packaged notice byte-matches the repository notice
- deterministic ZIP contains `archify/LICENSE` and `archify/THIRD_PARTY_NOTICES.md`
- package smoke passes on the extracted ZIP

PR #276, opened 2026-09-02
fix(evidence): compare --repo-root with filesystem path semantics on Windows

### LINKED ISSUE #274: [Bug]: --repo-root is rejected as not-top-level when the Windows drive letter is lowercase
### Archify version or commit

`2.17.0-dev.0`, at `main` commit `7a16d30`.

### Installation method

Git clone

### Diagram type

Architecture

### Exact command

```shell
# Same directory both times. Only the drive letter's case differs.
node archify/bin/archify.mjs deliver architecture c:\work\evidence-repo\minimal.architecture.json ^
  c:\work\evidence-repo\out.html --repo-root c:\work\evidence-repo --json   # exit 1
node archify/bin/archify.mjs deliver architecture C:\work\evidence-repo\minimal.architecture.json ^
  C:\work\evidence-repo\out.html --repo-root C:\work\evidence-repo --json   # exit 0
```

A lowercase drive letter is what a shell hands you routinely: `cd`, tab
completion, Git Bash / MSYS `cygpath`, `%CD%` after some launchers, and
copy-paste from tooling all produce `c:\...`.

```json
{
  "schema_version": 1,
  "meta": {
    "title": "Drive-case reproduction",
    "repository": {
      "url": "https://github.com/example/evidence-repo",
      "revision": "7f6a16e90328d2956b508f4070f5d55b75ade2f2"
    }
  },
  "components": [
    { "id": "router", "type": "backend", "label": "Router", "pos": [40, 40], "size": [130, 60],
      "sources": [{ "path": "src/router.js", "line": 1, "end_line": 3 }] },
    { "id": "store", "type": "database", "label": "Store", "pos": [260, 40], "size": [130, 60] }
  ],
  "connections": [{ "id": "read", "from": "router", "to": "store" }]
}
```

The fixture repository is one commit with a three-line `src/router.js` and
`origin` set to `https://github.com/example/evidence-repo.git`.

Verbatim `deliver --json` receipt for the lowercase run, with the private path
prefix replaced by `C:/work`. Note that `supportedFixes` asks for the directory
that was already passed, and that the diagnostic carries the proof of the defect
itself: `subject.repoRoot` and `evidence.gitTopLevel` are the same directory in
two spellings.

```json
{
  "schemaVersion": 1,
  "ok": false,
  "command": "deliver",
  "stage": "render",
  "type": "architecture",
  "input": "C:\work\evidence-repo\minimal.architecture.json",
  "output": "C:\work\evidence-repo\out.html",
  "error": "Evidence root must be the Git top-level directory: C:/work/evidence-repo",
  "diagnostics": [
    {
      "code": "repository-evidence/root-not-top-level",
      "severity": "error",
      "message": "Evidence root must be the Git top-level directory: C:/work/evidence-repo",
      "subject": {
        "surface": "repository-evidence",
        "repoRoot": "c:\work\evidence-repo"
      },
      "evidence": {
        "gitTopLevel": "C:/work/evidence-repo"
      },
      "supportedFixes": [
        "pass --repo-root C:/work/evidence-repo"
      ]
    }
  ]
}
```

The uppercase run on the same directory returns `ok: true` with
`evidence.verified: true` and one reference.

### Expected behavior

`--repo-root` identifies a directory. On Windows the volume is
case-insensitive, so `c:\work\evidence-repo` and `C:\work\evidence-repo` are the
same directory and both should verify.

### Actual behavior

The lowercase form fails with `repository-evidence/root-not-top-level` and no
artifact is written; the uppercase form verifies. `verifyRepositoryEvidence`
compares `path.resolve(repoRootInput)` (which preserves the case the caller
typed) against `git rev-parse --show-toplevel` (which always reports the
repository's own form) with a strict `!==`:

https://github.com/tt-a1i/archify/blob/7a16d30/archify/renderers/shared/repository-evidence.mjs#L133

The guard is right; the comparison is not. `realpathSync` does not normalize the
drive-letter case on Windows, so the two spellings of one directory compare
unequal. The failure is also hard to act on: the diagnostic's own
`supportedFixes` repeats the path that was just rejected, differing only in a
character whose case Windows does not distinguish.

POSIX is unaffected — paths there are case-sensitive and the strict comparison is
correct.

### Environment

Windows 11 Pro (10.0.26200), Node.js 22.22.3, Git Bash and PowerShell, Git
clone of `main` at `7a16d30`. Reproduced from a clean worktree with
`npm ci` in `archify/`.

### Final checks

- [x] I searched existing issues for the same problem. The closest is #144
      (CRLF checkout on Windows, closed), which is a different defect.
      personal or customer data.

jmvassoler:
## Problem and value

Fixes #274.

`--repo-root` names a directory, but `verifyRepositoryEvidence` compared it to
`git rev-parse --show-toplevel` with a strict `!==`. On Windows the volume is
case-insensitive and `path.resolve` preserves whatever case the caller typed,
while `--show-toplevel` always reports the repository's own form. The same
directory therefore verified as `C:\repo` and failed as `c:\repo` with
`repository-evidence/root-not-top-level` — and a lowercase drive letter is what a
shell hands you routinely (`cd`, tab completion, `cygpath`, copy-paste).

The failure was also hard to act on: the diagnostic's `supportedFixes` repeated
the path that had just been rejected, differing only in a character whose case
Windows does not distinguish.

## Scope

- **What changed:** one comparison in
  `archify/renderers/shared/repository-evidence.mjs`. A `samePath` helper
  compares two already-`realpath`-resolved absolute paths with filesystem
  semantics — case-insensitive on `win32`, exact everywhere else. Plus the
  regression tests, and the rebuilt `archify.zip` that embeds the changed
  renderer.
- **What deliberately did not change:** the guard itself, every diagnostic code,
  message, `subject`, `evidence` and `supportedFixes` shape, POSIX behavior
  (paths there are case-sensitive and the strict comparison was already
  correct), and the release identity — `archify/package.json` stays
  `2.17.0-dev.1`, with no tag or manifest version touched.
- **No unrelated changes:** confirmed. Three commits: fix, tests, archive
  rebuild.

Rebased onto `main` @`06dd052` (after #267) and fully re-verified there; every
number below was measured on that base.

## Stability impact

- **Compatibility and migration risk:** none for typed JSON. No schema,
  contract, or authored-geometry surface is involved. The change can only turn a
  previous `root-not-top-level` rejection into a verified render, and only on
  `win32`, and only when the two paths differ solely by case — i.e. when they
  were already the same directory. On POSIX `samePath` reduces to the original
  `===`.
- **Renderer, validator, package, or generated-artifact risk:** rendered output
  is unaffected; the golden-file tests pass unchanged. The package risk is the
  archive rebuild, covered below.
- **Failure behavior and rollback path:** evidence still fails closed. A
  subdirectory root still returns `repository-evidence/root-not-top-level`, a
  non-repository still returns `repository-evidence/git-command`, and neither
  writes an artifact — both are now pinned by tests. Rollback is reverting the
  three commits; nothing persists outside the tree.

## Tests run

**Red–green on the regression test** (Windows 11 Pro 10.0.26200, Node 22.22.3),
with the final test text and this base. Renderer reverted to the strict `!==`:

```
node --test test/repository-evidence.test.mjs
# tests 9 | pass 8 | fail 1 | skipped 0
not ok 7 - evidence root accepts either drive-letter case for one Windows directory
    -> deliver exits 1 with repository-evidence/root-not-top-level
```

Fix restored:

```
node --test test/repository-evidence.test.mjs
# tests 9 | pass 9 | fail 0 | skipped 0
```

The two guard tests pass in both states by design: `samePath` must relax the
drive case without relaxing the guard.

**Linux, the CI shape** (`node:22` container, Node 22.23.2) — full suite green:

```
npm test
# exit 0
# tests 1029 | pass 1001 | fail 0 | skipped 28

node --test test/repository-evidence.test.mjs
# tests 9 | pass 8 | fail 0 | skipped 1     (the win32-only test skips)
```

**Node floor and ceiling**, `node --test test/repository-evidence.test.mjs`:

```
Node 18.20.8   # tests 9 | pass 8 | fail 0 | skipped 1
Node 24.20.0   # tests 9 | pass 8 | fail 0 | skipped 1
```

Please treat remote CI as the authority — I reproduced its shape in containers,
I could not run your CI.

**Windows full suite** (Node 22.22.3), next to unmodified `main` @`06dd052`
measured the same way on the same machine:

```
main    # tests 1026 | pass  968 | fail 25 | skipped 33
branch  # tests 1029 | pass  970 | fail 26 | skipped 33
```

The branch adds three tests, all three of which pass here. The single extra
failure is `an empty precheck snapshot cannot start a second concurrent network
request` in `test/update-notifier.test.mjs` — intermittent rather than mine, see
the caveat below. The other 25 are pre-existing local-environment failures:
symlink creation needs privilege here, and the archive and `core.autocrlf`
checkout tests cannot pass on this box. None of them touch repository evidence.

One caveat I would rather state than hide: `test/update-notifier.test.mjs` is
intermittently red in my environment on unmodified `main` too — in the Linux
image it failed 1 of 3 isolated runs on `main`, on a different test each time.
It passed in the green full run quoted above, so if you see it fail on a retry,
it is not from this branch.

**Reproduction from the issue**, run against this branch: `--repo-root` with a
lowercase and an uppercase drive letter both return `ok: true` with identical
`evidence` blocks (`verified: true`, 1 reference, same revision). On `main` the
lowercase form exits 1.

## Visual evidence

Not applicable. No rendered output changes; the golden-file tests confirm
identical bytes. Visual review: skipped, deliberately — this change cannot reach
the artifact's appearance.

## Generated artifacts

`archify.zip` only.

`archify/renderers/shared/repository-evidence.mjs` ships inside the skill
archive, so leaving the committed container untouched made `archive build is
byte-for-byte reproducible across caller time zones without system zip` fail on
the Node 22 lane. Rebuilt with `scripts/build-zip.sh`:

```
before  d2052f494e79d0c826cd20220ac5071c5f10c8b8ffa759b7bd5af24a32e81fee  1321533 bytes
after   6c42d87ddb7e4c2565f21352c2660599933f55472033be19fae34565e61b9934  1321839 bytes
```

The only content difference is that renderer file. I verified these bytes are
canonical rather than merely mine: on the same toolchain, rebuilding `main`
@`06dd052` from its tracked inputs reproduces its committed archive bit for bit
(`d2052f49…`).

Nothing else was regenerated. Gallery, guide, start page, and README showcase
inputs are unchanged, and no schema, example, viewer, or `SKILL.md` surface is
touched. Release identity is untouched — say the word if you would rather carry
the archive rebuild separately under a release identity you choose, and I will
drop that commit.

One incidental finding, offered only as information and not addressed here:
building the archive on Windows with Node 22 from the same tracked inputs
produces different bytes, so I built the canonical archive in a Linux Node 22
container. That predates this change — it is also why the archive gate sits
among the pre-existing Windows failures above.

## Checklist


sunsunsun-java:
Thanks for the focused fix. The CLI regression and the negative cases for subdirectory/non-repository roots cover the intended behavior well.
I reviewed commit 5008613. On macOS with Node 22, the targeted suite passed with 8 passes and 1 Windows-only skip. Rebuilding archify.zip also reproduced the committed archive byte-for-byte.
Before merging, please address two items:
1. Run the new regression in Windows CI. Currently, npm test runs only on Ubuntu, while the Windows package-smoke job does not exercise --repo-root. A focused Windows step is sufficient:
   node --test --test-name-pattern="evidence root" test/repository-evidence.test.mjs
   Run this from archify/.
2. Update the branch against the latest main and resolve the conflicts. Rebuild archify.zip with Node 22 from the final source, then rerun the affected checks.
The implementation looks reasonable for #274. Once the branch is current and the Windows regression runs successfully in CI, it should be ready for final approval.

tt-a1i:
Consolidating #274 into #309, which includes the same root-path spelling problem plus native realpath handling for Windows 8.3 aliases. This PR’s tests and rationale remain useful comparison evidence. Neither Windows fix is being declared shipped by this closure: #309 still needs current-main integration and actual filesystem-identity verification, including preserving the top-level-directory guard on case-sensitive filesystems rather than assuming every Windows directory is case-insensitive. We will maintain one implementation path for this bug.

PR #287, opened 2026-09-03
fix: stop reader width oscillation when page overflows the viewport

szsip239:
## Problem

When a generated page is taller than the viewport (e.g. any content appended after the `.cards` section), the adaptive reader shell enters an infinite width oscillation and the page visibly flickers nonstop:

- `measure()` recomputes the desired width from the remaining SVG height budget and calls `applyWidth(desired)`.
- `settleOverflow()` sees `scrollHeight > innerHeight` and shrinks the width again via `applyWidth(lastWidth - overflow * ratio - 4)`.
- Every `applyWidth` reflows the observed `header`/`guided`/`cards` elements, retriggering the `ResizeObserver`, which schedules another `measure()` — forever.

Reproduction (measured with headless Chrome against a showcase workflow artifact with a tall appended section, 1440x900): **66 `--archify-reader-width` writes in 6 seconds, alternating 1009px ↔ 960px**. Without the extra content the same page converges after 3 writes.

## Change

`assets/template.html`, adaptive reader shell only:

- Track the width that `settleOverflow()` actually settled on (`settledWidth`).
- `measure()` never grows the width past `settledWidth` while the overflow persists.
- `settledWidth` resets to 0 as soon as the page no longer overflows, so the reader widens again when the viewport allows.

No behavior change when the page fits the viewport; embed/present/print and narrow-viewport paths are untouched (they were already ineligible).

## Evidence

- Same artifact after the patch: **2 width writes, then stable** (`1007px → 960px`, `data-reader-overflow="authored"`), verified at 1440x900, 1280x800, 1920x1080, 1280x2400.
- New regression test in `archify/test/adaptive-reader-layout.test.mjs` asserting the sticky-settle contract at source level.
- `npm test` (full suite incl. golden renders): all checks passed, 992 tests pass.
- Examples re-rendered and `archify.zip` rebuilt with Node 22 via `scripts/build-zip.sh` so the CI byte-compare passes.

## Rollback

Single commit touching only the reader shell JS in the template plus regenerated artifacts; reverting restores the previous layout loop logic with no data or schema impact.


tt-a1i (changes requested):
The flicker is real and worth fixing, but the sticky-width change only handles overflow that remains after shrinking.

I rendered the current dev architecture example, kept its content, changed the test viewBox to 0 0 3000 1000, appended a fixed-height block before the Viewer script, and observed reader widths for 180 animation frames at 1440x900 in real Chrome. I compared current dev with the PR's exact settledWidth logic transplanted into that same artifact, so this is a current-base integration probe, not a claim that the untouched old PR head was tested.

- With 500px additional content, dev repeatedly alternates 1376/960; the proposed logic settles at 960. This confirms the useful case.
- With 180px additional content, the proposed logic still alternates 1216/1376 throughout observation (the page fits at the smaller width).
- With 220px additional content, it likewise continues alternating 1096/1376.

The cause is the new `settledWidth = 0` branch: as soon as one shrink removes overflow, the next observer-triggered measure grows back to the width that caused overflow. Retain convergence across that internal feedback cycle while allowing growth again when actual viewport/content inputs change. Please add a real-browser regression for both persistent and shrink-solvable overflow, plus removal/resize recovery; the current source-text assertion does not protect this behavior.

Also port the change to `viewer/reader-layout.js`, regenerate its template/package outputs, and clear any new state when the reader becomes ineligible. The current `assets/template.html` is generated. Keep diagram geometry and the existing embed/present/print behavior intact. Author revision on current dev is the next step; no new layout abstraction is required solely for this fix.

PR #289, opened 2026-09-03
feat(viewer): make story and journey autoplay dwell authorable

### LINKED ISSUE #57: autoplay dwell
Both autoplay dwell constants in `assets/template.html` are hardcoded at `1100`, and neither is authorable:

- `STORY_FOLLOW_MIN_DWELL_MS = 1100` (line 8507) — guided chapter story
- `JOURNEY_DWELL_MS = 1100` (line 11236) — node-by-node journey

Because `storyBeatDwell` takes the max of the floor and the share of the view interval (line 9322):

```js
var VIEW_INTERVAL_MS = 3200;
var STORY_FOLLOW_MIN_DWELL_MS = 1100;
function storyBeatDwell(total) {
  return Math.max(STORY_FOLLOW_MIN_DWELL_MS, VIEW_INTERVAL_MS / Math.max(1, total));
}
```

the floor always wins for a story of three or more chapters — `3200 / 3 = 1066` — so every beat gets exactly 1.1 seconds regardless of how many there are.

### Why it matters

Readers reported this as far too fast to read a beat's note *and* look at the region the camera just panned to. A beat that pans somewhere new and shows a sentence needs more than 1.1s.

### The gap

No schema field and no CLI flag exposes either constant. The only workaround is patching the installed `assets/template.html` and re-delivering every diagram, which does not survive a skill upgrade.

### Suggested direction

An authored `meta.motion` object (or similar) carrying a dwell multiplier or an explicit per-beat dwell, threaded into the template the way other `meta` values already are.

Happy to open a PR if you'd like it in a particular shape — the open question is whether this belongs in `meta` or as a delivery-time flag, and that's your call.

---

Verified against `main` at `19213e1` (v2.13.0).

SomSamantray:
## Problem and value

Fixes #57.

Both autoplay dwell constants in the delivered viewer were hardcoded at `1100`, and neither was authorable. `storyBeatDwell()` takes the max of that floor and `viewInterval / beatCount`, so for a story of three or more beats the floor always won and every beat got exactly 1.1s regardless of how much content it carries. The only workaround was patching the installed `assets/template.html` and re-delivering every diagram, which does not survive a skill upgrade.

Diagrams can now author playback pacing directly. `meta.motion.dwell` carries optional `storyFollowMinMs`, `journeyStepMs`, and `viewIntervalMs` values (integers, 1100-60000ms) that reach the delivered page as a JSON data block and drive the story and journey schedulers. Authoring nothing keeps today's exact 1100/3200 defaults and byte-identical rendering.

## Scope

- What changed:
  - `meta.motion` added to the shared schema `$defs` and referenced from all five diagram types; standalone validators regenerated.
  - `writeDiagram`/`applyTemplate` emit an optional `archify-motion-data` JSON block; authored-less renders emit no block.
  - The viewer resolves the block once at boot into `Archify.motionConfig`; the guided-story and route-journey schedulers read their timing from it, keeping the readability floor as the fallback.
  - New `autoplay-dwell.test.mjs` regression covering all five renderers (authored, absent, partial, out-of-bounds); three pinned literal assertions rebased onto the authored-default shape.
- What deliberately did not change: the 1100ms/3200ms defaults, the `max(floor, interval/count)` story-dwell formula, camera-settle design, route animation durations, and the schema-v1 typed JSON contract (the addition is optional and additive).
- No unrelated changes: confirmed. Every file in the diff serves the dwell contract or is a regenerated artifact of the template change.

## Stability impact

- Compatibility and migration risk: low. Existing schema-v1 documents validate identically (verified against all five checked-in examples with and without `meta.motion`). Artifacts without `meta.motion` render byte-identically modulo the timing-code region of the embedded template.
- Renderer, validator, package, or generated-artifact risk: standalone validators are regenerated from the schemas (`check:validators` green). A corrupt or hand-edited motion block degrades to the floor defaults instead of killing the viewer boot script (guarded; browser-verified with a literal `null` block).
- Failure behavior and rollback path: out-of-bounds dwell values fail schema validation with a machine-readable diagnostic before delivery. Rollback is a plain revert; no data or format migration is involved.

## Tests run

```
cd archify && npm test
# tests 1030 | pass 1003 | fail 0 | skipped 27 (browser-gated, no ARCHIFY_CHROME in runner)

node --test test/autoplay-dwell.test.mjs
# tests 4 | pass 4 | fail 0

node --test test/story-follow-camera.test.mjs test/route-journey.test.mjs \
  test/guided-views.test.mjs test/story-trail.test.mjs test/story-beat-navigator.test.mjs
# tests 30 | pass 30 | fail 0

node bin/archify.mjs validate architecture <motion-fixture>.json --json   # exit 0
node bin/archify.mjs validate architecture <storyFollowMinMs:10>.json --json
# exit 1 — schema/minimum: /meta/motion/dwell/storyFollowMinMs must be >= 1100
```

`npm test` includes `check:validators`, `test/golden.mjs`, gallery freshness, and the version-sync gates. Browser verification (Chrome, real playback): authored dwell held beat 1/5 past the 1.2s mark and advanced to 2/5 at ~2.2s, matching the authored 2000ms floor; a hand-tampered `null` motion block degraded to the 1100/3200 defaults with the viewer fully alive.

## Visual evidence

Static rendering is unchanged by design (geometry/layout verified byte-identical in the regenerated goldens; zero geometry-touching diff lines). Visual review of the delivered page: passed — diagram, guided-views panel, and playback all render and animate correctly in Chrome with authored, corrupt, and absent motion blocks. No layout or visual regression observed.

## Generated artifacts

Regenerated because the viewer template changed (authoritative inputs listed per `CONTRIBUTING.md`):

- `archify/examples/*.html` (5) and `examples/*.html` mirrors (5) + `examples/web-app.html`
- `examples/checkout-platform-delta.html` + `.receipt.json`
- `docs/gallery/*` (11 artifacts) + `docs/gallery/manifest.json` + `docs/gallery.html`
- `docs/assets/archify-live-proof.gif` + `.json` (54 frames / 5.4s — visual proof content unchanged, only embedded template bytes)
- `archify.zip` (rebuilt with Node 22, byte-reproducibility gate green)

## Unapplied review findings

Residuals from the pre-ship review (run 20260903-214252-3c7dcf46) that need a maintainer decision; finding #1 (corrupt-block guard) was applied in this PR.

- [ ] P3 — `archify/SKILL.md` — `meta.motion` has no author-facing documentation or changelog entry; add a SKILL.md bullet, schemas/README paragraph, authoring-contract section, and CHANGELOG entry (suggested fix in review).
- [ ] P3 — `archify/test/autoplay-dwell.test.mjs` — runtime wiring is proven only by source-shape regexes; a DOM-shimmed execution probe of the resolver would close the gap a DEFAULTS-mapping typo could slip through.
- Residual (informational): authored `storyFollowMinMs` above the default `viewIntervalMs` lets the floor dominate short chapters — the documented `max()` semantics; a cross-field validation rule is the alternative if that surprise is unwanted.
- Residual (informational): authoring any `meta.motion` lowers the attainable `viewIntervalMs` minimum from 3200 to the shared 1100 floor; worth a line in the contract docs either way.

## Checklist


tt-a1i (commented):
The value in #57 is clear: authors should be able to allow enough time to read a beat without patching the installed Viewer. The bounded optional dwell fields are a proportionate approach; preserving the 1100/3200 defaults keeps existing diagrams compatible. I reviewed the schema, serialization and scheduler wiring at `0e43e9d471ee80b12cf9de835f413b35f90120f4`.

For the open questions in the description:

- Keep the existing `max(storyFollowMinMs, viewIntervalMs / beatCount)` rule. A large authored floor dominating a short chapter is expected from an explicit minimum; no new cross-field rejection is needed.
- Document the three fields, units, bounds, defaults and that formula in one canonical authoring reference, with a short Skill link if needed. We do not need the same contract copied into four documents. Keep released version history unchanged; an Unreleased note can follow current dev's existing structure.
- Close the scheduler verification gap with a focused real-browser regression for authored and absent values, partial configuration, and the existing malformed/null-block fallback. Assert actual progression/pause behavior using generous bounds or observable scheduler state; source-string checks alone do not establish playback timing. The reported manual Chrome run is useful historical evidence, but should be tied to the artifact/head and not relabeled as an independent current-dev result.

Next, refresh onto current dev and put runtime changes in the canonical Viewer sources. Preserve the current motion ownership/pause/reset behavior and add the browser test to the shared gate. Regenerate affected examples/Gallery/ZIP once from the combined source, then provide the final CI result. The old full-suite numbers in this description belong to the older base and are not an integration result.

This is a positive scope assessment with concrete integration/evidence work remaining, not a newly reproduced runtime defect. I have not rerun this branch's full tests or independently accepted its playback on the current dev runtime. Please keep this focused on authored pacing; no additional animation controls are required.

PR #294, opened 2026-09-04
feat(viewer): support bounded multi-role comparison

### LINKED ISSUE #83: [Feature][Viewer] Support broader comparison and consistent hover previews
## Context

I used an Architecture diagram as a product competition map. The semantic Lens is useful, but it allows only one or two semantic kinds. For business comparison tasks, I often need to compare three or more selected modules, or several price-band roles, in one view.

There is also an interaction inconsistency: hovering input/output relationship controls can preview the chain by highlighting the target path and dimming unrelated elements, while the `Upstream` and `Downstream` controls require a click before the same kind of visual focus appears.

## Current behavior

- Lens supports one or two semantic kinds.
- Broader comparison requires repeating separate two-way comparisons.
- Some relationship controls provide hover preview; upstream/downstream exploration is click-only.

## Suggested direction

Either extend Lens to a small bounded multi-selection, or add a separate multi-node comparison mode. A limit of three or four selections would cover common comparison use cases without turning the viewer into a general-purpose analytics tool.

For relationship controls, use a consistent model:

- hover/focus: temporary preview with unrelated elements dimmed;
- click/keyboard activation: pin the selection and update the stable link;
- Escape/Clear: restore the overview.

## Acceptance criteria

- Users can compare more than two selected roles or nodes in a bounded way.
- The viewer clearly distinguishes role comparison from node comparison.
- Upstream/downstream controls provide the same temporary hover/focus preview pattern as other relationship controls.
- Click still pins a truthful authored state and stable link.
- Keyboard and reduced-motion behavior remain supported.

## Environment

- Archify 2.14.0
- macOS, Node.js 24.19.0
- Codex Desktop

## Final checks

- [x] This report contains no secrets, private repository content, personal data, or customer data.

seekskyworld:
## Summary
- extend Semantic Lens from two to a bounded maximum of four semantic kinds
- distinguish single-kind node inspection from multi-kind role comparison
- add temporary hover/focus previews for Upstream and Downstream controls; click/keyboard activation pins the authored state and URL
- preserve Escape/Clear, keyboard navigation, reduced-motion, embed, print, and export boundaries
- refresh all generated artifacts, gallery outputs, README proof, and canonical archify.zip

## Issue
Closes #83

## Verification
- `node --test test/semantic-lens.test.mjs test/authored-reachability.test.mjs` passes 8/8
- `npm test` passes 987/987 executed tests (1,018 total; 31 skipped)
- Existing two-kind direct relationship comparison remains unchanged
- Canonical SVG remains free of viewer-only lens/reachability state


yunaremaia:
Closing this PR — the branch is conflicting and a cleaner implementation exists elsewhere. Will revisit with a fresh branch when capacity allows.

tt-a1i (changes requested):
The bounded four-kind comparison can be useful and does not require changing authored topology. However, the bundled hover preview currently destroys a user's pinned reachability selection.

At df6ca4dac79f06ca92a4f738bda497355d32d2b6, I ran the eight Semantic Lens/authored-reachability tests (all pass, zero skips). I then reproduced this in real Chrome using the packaged web-app diagram:
1. Select `users` and pin Downstream. `Archify.focus.reachability()` contains the downstream graph, the button is pressed, and the URL is `#focus=users&reach=downstream`.
2. Enter and leave the Downstream button with a mouse pointer without clicking. The pointer events were dispatched in the real Viewer DOM.
3. Reachability becomes null and the button becomes unpressed, while the URL still says downstream. A reload restores a state the current page has silently lost.

`previewReachability()` calls `applyReachability()` even when an authored state is already pinned; that marks it as a preview, and pointerleave/blur clears it. Preserve the committed state separately from the transient preview, or avoid starting a preview when a pin is active. Leaving a preview must restore the committed state and keep the URL consistent. Add a real-browser regression covering hover/focus cancellation with an existing pin, switching direction, and click/keyboard promotion from preview to pin; current string-match checks cannot catch this.

For the three/four-kind feature, also demonstrate the intended displayed relationship set in the browser: two kinds currently show only between-kind links, while three/four use all links touching any selected kind, including unselected peers. That difference can be reasonable, but the UI description and test should make it explicit rather than imply the same comparison filter at every cardinality.

Please port the final behavior to current dev's canonical `viewer/focus.js` and `viewer/semantic-lens.js`, include the browser regression in the shared PR/release gate, regenerate affected outputs and the canonical ZIP, and pass final-head CI. The author owns this revision. No requirement to add more than four kinds or invent inferred relationships.

Issue #301, opened 2026-09-04
repository-evidence: identity binding is hard-coded to github.com, blocking self-hosted Git (Gitea/GitLab/Forgejo)


macawi-ai:
### Summary

`repository-evidence` is the most valuable part of Archify for teams who need diagrams that fail when they go stale — but it is currently unusable on self-hosted Git (Gitea, GitLab CE, Forgejo, Bitbucket Server), because the *identity binding* is hard-coded to github.com while every other part of the verification is already host-agnostic.

### Where it is pinned

Two places, both identity-only:

1. `schemas/architecture.schema.json` — `meta.repository.url` is regex-locked:
   ```
   ^https://github\.com/[A-Za-z0-9_.-]+/[A-Za-z0-9_.-]+(?:\.git)?/?$
   ```
2. `renderers/shared/repository-evidence.mjs` — `githubSlug()` matches only `https://github.com/…`, `git@github.com:…`, `ssh://git@github.com/…`, and the origin check is `githubSlug(origin) !== authoredSlug`.

Everything downstream is already portable: `rev-parse --show-toplevel`, `cat-file -e <rev>^{commit}`, path-escape rejection, and line-range checks make no assumption about the host.

### Effect

With a self-hosted origin, `authoredSlug` is always a valid GitHub slug (the schema permits nothing else) while `githubSlug(origin)` returns `null`, so validation always fails with `repository-evidence/origin-mismatch`. There is no configuration that reaches it. The practical result is that the whole `sources[]` + `--repo-root` capability is off-limits to self-hosted repositories, including air-gapped and internal-only ones — arguably the users with the strongest need for architecture diagrams that are pinned to a revision.

### Suggested direction (deliberately minimal)

Generalise the *identity* check only, leaving the verification logic untouched:

- widen the schema pattern to any `https://<host>/<owner>/<repo>` (plus the `git@`/`ssh://` forms), and
- generalise `githubSlug()` to a `originSlug()` that normalises host + owner + repo, comparing all three.

That keeps the guarantee intact — the diagram is still bound to a specific origin and revision — while making it work anywhere. Optional, if you would rather be conservative: an explicit `meta.repository.host` field, or an allowlist, so the default behaviour is unchanged for existing users.

### Why it matters

[...]

PR #315, opened 2026-09-05
fix: stabilize offline reader geometry

### LINKED ISSUE #277: [Bug]: adaptive reader width depends on reaching Google Fonts (960 -> 987 offline), and the pinned desktop-reader test only passes online
### Archify version or commit

`2.17.0-dev.1`, at `main` commit `06dd052`.

### Installation method

Git clone

### Diagram type

Generated viewer

### Exact command

Nothing modified. The only variable is whether the webfont host is reachable.

```shell
# Both runs: same commit, same container, chromium + fonts-dejavu installed.
node bin/archify.mjs render architecture \
  examples/production-deployment.architecture.json /tmp/a.html --quality showcase
node bin/archify.mjs visual-check /tmp/a.html --json

# Then make the font host unreachable, exactly as an air-gapped machine sees it:
printf "127.0.0.1 fonts.googleapis.com\n127.0.0.1 fonts.gstatic.com\n" >> /etc/hosts
node bin/archify.mjs visual-check /tmp/a.html --json

# The pinned regression test shows the same split:
ARCHIFY_CHROME=/usr/bin/chromium node --test test/desktop-reader-browser.test.mjs
```

None needed — the checked-in `archify/examples/production-deployment.architecture.json`
reproduces it, and it is the same input `test/desktop-reader-browser.test.mjs`
already uses. No custom or private JSON is involved.

`visual-check` receipt, 1440x900 desktop viewport, same artifact in both states:

```json
{
  "fontHostReachable": {
    "readerWidth": 960,
    "diagramWidth": 930,
    "viewBoxWidth": 1376,
    "minimumProjectedNodeTextPx": 6.000,
    "minimumRequiredNodeTextPx": 6,
    "readabilityOk": true,
    "overflowX": false
  },
  "fontHostBlocked": {
    "readerWidth": 987,
    "diagramWidth": 957,
    "viewBoxWidth": 1376,
    "minimumProjectedNodeTextPx": 6.174,
    "minimumRequiredNodeTextPx": 6,
    "readabilityOk": true,
    "overflowX": false
  }
}
```

And the pinned test, with the host blocked:

```
not ok 1 - production showcase is readable in the real 1440 by 900 adaptive reader
  error: 987 !== 960
  expected: 960
  actual: 987
  stack: test/desktop-reader-browser.test.mjs:46  assert.equal(observation.readerWidth, 960)
```

### Expected behavior

The adaptive reader settles at its designed width — `readerWidth` 960 and
`diagramWidth` 930 at 1440x900 — whether or not the artifact could reach
fonts.googleapis.com. A delivered artifact is offline evidence, so its geometry
should not depend on a network fetch succeeding.

### Actual behavior

With the font host unreachable, the reader settles 27px wider: `readerWidth`
987, `diagramWidth` 957. `test/desktop-reader-browser.test.mjs` fails on the
pinned `readerWidth` assertion.

I want to be precise about the severity, because it is narrower than it first
looks: **readability does not degrade.** `readabilityOk`, `overflowX`, the
legend/dock intersection areas, `dockStageGap` and `viewerChromeStageOk` all
pass in both states, and the smallest projected label is actually marginally
larger offline (6.174px vs 6.000px against the 6px floor) because the reader is
wider. So this is an off-spec width and a network-dependent pinned test, not a
broken artifact.

Two things seem worth deciding, and both are yours:

1. The reader measures rendered text at runtime, so its chosen width tracks
   whichever face actually loaded. `archify/assets/template.html:354` already
   prefers a locally installed `'JetBrains Mono'` and then a system monospace
   chain, and the export path emits `@font-face` with
   `src: local('JetBrains Mono')`. The template comment says the stack "falls
   back to system monospace until it lands", which reads as a transient state —
   but for anyone air-gapped, behind a proxy, or in a locked-down enterprise
   network, the fallback is the steady state, and they silently get different
   geometry from everyone else.
2. `desktop-reader-browser.test.mjs` therefore only reproduces 960 when the
   machine running it can reach Google Fonts. That is a hidden network
   dependency in a pinned layout contract, which also means the test cannot
   verify the offline case it arguably cares most about.

Note the control also shows `minimumProjectedNodeTextPx` sitting at exactly
6.000 against a 6px floor when the font does load — zero margin. That is your
design call, not something I measured breaking, but it is why the geometry is
sensitive to which face renders.

I have no attachment to a particular fix, and I am not proposing one here since
the contract is yours to set. For the record, embedding the face would hold 960
and remove the request, but it changes distribution: JetBrains Mono is OFL 1.1
and `THIRD_PARTY_NOTICES.md` has no entry for it today, precisely because the
font is only linked.

Happy to open a PR for whichever direction you choose, or to add an
offline-mode variant of the desktop-reader test if that is the useful half.

### Environment

`node:22` Debian bookworm container, Node 22.23.2, Chromium 151.0.7922.173,
`fonts-dejavu` installed so the fallback chain had a real font from the
template's own CSS stack. Also observed on Windows 11 Pro 10.0.26200 with
Node 22.22.3. Discovered while making artifacts request nothing on a fork, so
the same numbers appear there with the `<link>` removed rather than blocked.

### Final checks

      personal or customer data.

testikun:
## Problem and value

Fixes #277. At the canonical 1440x900 desktop viewport, the adaptive reader previously settled at 960px when Google Fonts loaded but 987px when the font hosts were unreachable. Compact desktop artifacts now use the canonical 960px reader width without changing authored SVG geometry; taller desktop viewports keep the existing height-budgeted expansion.

## Stability impact

- Impact class and changed behavior/shared callers: moderate viewer-layout change in the wide adaptive reader only when the viewport height is 900px or less.
- Existing behavior preserved / intended compatibility changes / failure behavior: embed, present, print, non-wide diagrams, and taller desktop expansion remain unchanged. Offline 1440x900 changes intentionally from reader/diagram 987/957 to 960/930.
- No unrelated changes: the branch was merged with current main, the accidentally mixed #301 repository-evidence changes were removed, and the earlier unsupported screenshots were replaced.

## Tests run

Comparison: `daced37` (current main) to `b8cd6f0`.

- Node 22.23.2: `node --test archify/test/adaptive-reader-layout.test.mjs archify/test/desktop-reader-browser.test.mjs archify/test/visual-check.test.mjs` -> 18 passed, 0 failed, 0 skipped.
- Real Chrome: font host reachable ran 3 times; explicit CDP blocking of `fonts.googleapis.com` and `fonts.gstatic.com` ran once. Every 1440x900 observation reported readerWidth=960, diagramWidth=930, no overflow, and readability pass.
- Ablation: temporarily removing the compact-height bound reproduced readerWidth=987 and diagramWidth=957 with the font hosts blocked, and the 960px assertion failed; the bound was restored.
- Distribution: Node 22.23.2 rebuild was byte-identical on a second build; `unzip -tq archify.zip` passed; the extracted package `doctor` command passed.

## Visual evidence

| Blocked font hosts, main | Blocked font hosts, this PR |
| --- | --- |
| Before | After |

- Comparison conditions: same production deployment input, Chrome, 1440x900 viewport, dark theme, Blueprint preset, read detail, still motion, and blocked Google font hosts.
- Automated or browser checks: pass; before=987/957, after=960/930, containment and readability pass in both.
- Perceptual visual review: passed. The expected difference is the narrower canonical reader; no overlap or clipping was observed.

## Generated artifacts

- Rebuilt `archify.zip` with Node 22.23.2; SHA-256 `d0b362c25f45cb1b272bed96839b810c1c8a5cd0651378705cd90c550b03bf98`.
- Regenerated the two controlled 1440x900 dark screenshots shown above.


testikun:
Controlled offline reproduction is now complete and the branch has been updated.

| Font hosts blocked, current main | Font hosts blocked, PR head |
| --- | --- |
| Before | After |

Both captures use the same production deployment input, Chrome, 1440x900 viewport, dark theme, Blueprint preset, read detail, and still motion. Chrome DevTools explicitly blocked `fonts.googleapis.com` and `fonts.gstatic.com` before navigation.

- Current main `daced37`: readerWidth=987, diagramWidth=957.
- PR head `b8cd6f0`: readerWidth=960, diagramWidth=930.
- Both remain contained and readable; the PR restores the canonical compact-desktop geometry.

The browser regression now runs both the normal font-loading path and the blocked-host path. Node 22.23.2 targeted tests report 18 passed, and the rebuilt distribution ZIP is deterministic and passes extracted-package `doctor` validation.

sunsunsun-java:
Thanks for working on this. After checking the overlap, #256 already addresses the underlying cause of #277 by embedding the font, preserving the adaptive layout without introducing the additional 960px cap for viewports up to 1600px.
I suggest consolidating the fix in #256 and closing this PR as superseded. Any controlled offline reproduction or regression coverage from this work would still be valuable there, especially the original 1440×900 case with the 960px reader / 930px diagram assertions.
#277 should remain tracked until #256 is merged and its offline acceptance case is verified.

Issue #143, opened 2026-08-27
repository-evidence 建议支持 Gitee 等非 GitHub 仓库，并将本地证据校验与源码链接生成解耦


chendlcom:
## 使用场景

我们在一个包含多个独立 Git 仓库的企业项目中使用 Archify。代码托管在 Gitee，而不是 GitHub。

希望使用 Architecture 的 repository evidence 能力，为节点绑定固定 commit 下的源码文件和行号。但目前即使本地仓库、commit、文件和行号全部有效，也无法启用该功能。

## 当前限制

目前 repository evidence 与 GitHub 强绑定：

1. Architecture Schema 将 `meta.repository.url` 限制为 `https://github.com/...`：
   https://github.com/tt-a1i/archify/blob/9a5060566c832832fb843e457e58c8ee6bac82fd/archify/schemas/architecture.schema.json#L29-L33

2. 运行时要求 URL 以 `https://github.com/` 开头，否则返回 `repository-evidence/url-invalid`：
   https://github.com/tt-a1i/archify/blob/9a5060566c832832fb843e457e58c8ee6bac82fd/archify/renderers/shared/repository-evidence.mjs#L94-L118

3. 本地仓库 `origin` 通过 GitHub 专用的 `githubSlug()` 与声明地址匹配：
   https://github.com/tt-a1i/archify/blob/9a5060566c832832fb843e457e58c8ee6bac82fd/archify/renderers/shared/repository-evidence.mjs#L132-L146

因此，非 GitHub 仓库无法使用经过验证的 `SRC` 引用，即使 Archify 实际需要校验的本地证据都已具备。

## 期望行为

建议将两类能力解耦：

1. 本地证据校验：
   - Git 仓库根目录
   - origin 一致性
   - commit 是否存在
   - 文件是否存在
   - 路径是否越界
   - 行号范围是否合法

2. Web 源码链接生成：
   - GitHub
   - Gitee
   - GitLab
   - 自建 GitLab/Gitea
   - 不生成链接的 local-only 模式

源码证据是否可信，主要取决于本地 Git、commit、文件和行号校验，不应取决于仓库是否托管在 GitHub。

## 建议的最小实现

可以先支持以下方式之一：

### 方案 A：Provider Adapter

在 `meta.repository` 中增加可选的 provider：

```json
{
  "repository": {
    "provider": "gitee",
    "url": "https://gitee.com/owner/repository",
    "revision": "<full-commit-sha>"
  }
}
```

[...]

PR #316, opened 2026-09-05
fix: support self-hosted repository evidence

### LINKED ISSUE #301: repository-evidence: identity binding is hard-coded to github.com, blocking self-hosted Git (Gitea/GitLab/Forgejo)

testikun:
## Summary

Fixes #301 by supporting self-hosted Git origins in revision-pinned repository evidence. HTTPS, SSH, origin matching, revision verification, path containment, and line-range validation remain enforced.

## Validation

Repository evidence tests pass for GitHub and self-hosted HTTPS/SSH remotes. The ablation experiment confirms the generalized origin parser is required for self-hosted repositories.


sunsunsun-java:
Thanks for the contribution and follow-up fixes. The functionality requested in #301 is now covered by the merged #354, which addresses the broader scope of #143 by separating local evidence verification from web source-link generation.
Closing this PR as superseded by #354. No separate merge is needed. Thank you for helping identify and validate the self-hosted repository use cases.

PR #317, opened 2026-09-05
feat(architecture): allow connections to named boundaries

### LINKED ISSUE #284: Allow connections to bind to a boundary (region/security-group), not just a component
## Problem

In the `architecture` diagram type, `connections[].from`/`to` can only reference an `id` in `components[]`. There's no way to make an arrow visually begin or end at a `boundaries[]` entry (a dashed `region` or `security-group`).

This matters when the *real* relationship being diagrammed is between two zones/containers, not between two specific leaf boxes — e.g. "Team A's service mesh talks to Team B's data platform" where the accurate arrow should touch the dashed perimeter of each team's boundary, not one arbitrarily-chosen box inside it. Today the only workaround is to pick a proxy leaf component near the boundary edge and shrink that boundary's `pad` so the arrow visually "punches through" the border on its way to the proxy — which gets close but isn't a real boundary-anchored endpoint, and breaks down when the boundary has no leaf sitting near the true attachment point (e.g. the original relationship binds to the frame's geometric center-top, not any child).

## Proposed fix

Scoped to the `architecture` renderer/schema:

1. **Schema** (`schemas/architecture.schema.json`): add an optional `id` field to `boundaries[]` entries (same pattern as `common.schema.json#/$defs/id`), plus a validation rule that a boundary `id` can't collide with any `components[].id`.
2. **Renderer** (`renderers/architecture/render-architecture.mjs`): `boundaryRect(boundary)` already computes `{ x, y, width, height, ... }` for every boundary — the same shape `measureComponent()` produces for components. After boundary rects are finalized (i.e. after the label-fit expansion pass in `resolveBoundaryTitles()`/`layoutBoundaryTitles`, so the anchor matches what's actually drawn), build a second `id → rect` lookup for boundaries and have connection-endpoint resolution check both maps (currently every endpoint lookup goes through the single `components` Map, e.g. `components.get(conn.from)`). Boundary entries should **not** be merged into the `components` Map itself, since that map also drives the type legend/counts.
3. **Geometry**: confirm `anchor()`/`routeVia()`/`connectionSides()` work against a boundary rect the same way as a component rect (may just need `cx`/`cy` added alongside `x/y/width/height`).
4. **Validators**: spot-check `clean-flow/container-border-run` and `clean-flow/edge-through-node` don't misfire when a connection's declared endpoint *is* the boundary it touches (a terminating perpendicular segment at a border shouldn't be flagged as "running along" it).
5. Add a test fixture exercising a boundary-bound connection (`test/` already has broad coverage for the existing composition checks).

## Scope

Proposing this for `architecture` only to start; the other diagram types (`workflow`, `sequence`, `dataflow`, `lifecycle`) don't have the same region/security-group boundary concept, so this wouldn't apply to them.

Happy to help test against real-world specs if useful.

ouchihao:
## Problem and behavior

Fixes #284. Architecture connections can address optional boundary IDs so zone relationships land on visible frames without proxy components. Component-to-boundary and boundary-to-boundary links use the shared router and the existing label, composition, Viewer, and delta contracts.

This remains an additive schema-v1 change: anonymous boundaries work as before; component inventory and legend counts exclude boundaries. Explicit sides, `via`, routes, and label offsets remain authoritative. Boundaries expose outward-facing interfaces; links to contained endpoints report `architecture/boundary-endpoint-contained`, and `wraps` creates membership rather than graph edges.

## Revision for the maintainer review

Addresses [the September 16 review](https://github.com/tt-a1i/archify/pull/317#pullrequestreview-5227921604). The PR targets `dev`. Merge commit `978687e7d4067acb406db0a2990b6a1f5fcc3d5c` integrates dev `548f3b9e0e3aaf8ce9938e659574ec9fc3871d45` with the previously reviewed head `9de26f1`.

- Uses dev's extracted `createRouter` instead of retaining an inline router. Endpoint lookup/port spreading accept named frames, while component obstacle checks retain only real components. An unrelated named frame does not become an opaque obstacle.
- Boundary-title convergence measures routes and labels for each candidate frame layout. The returned final scene supplies rendering, label masks, canvas sizing, diagnostics, layout receipts, and legend avoidance. Long titles and offset labels therefore use the same final frame geometry.
- Retains current-dev dominant-axis inference and port spreading. Auto canvas includes boundary connection labels; an undersized authored viewBox reports `composition/label-canvas-containment` with supported fixes.
- Regenerates validators and `archify.zip` from the combined sources. No release-version change or separate Viewer feature work.

## Validation

Windows, Node 22.23.2; commands from `archify/`:

```sh
node --test --test-concurrency=2 test/architecture-boundary-endpoints.test.mjs test/geometry.test.mjs test/layout-rules.test.mjs test/automatic-port-spread.test.mjs test/architecture-delta.test.mjs test/engineering-profile.test.mjs test/renderer-import-isolation.test.mjs test/generate-validators.test.mjs
```

Run as the boundary file and a separate seven-file related suite: **20 + 263 passed**, no failures or skips. Four added integration cases cover long-title/offset-label geometry and canvas containment, vertical dominant-axis routing, boundary fan-out/order independence, and named-frame transparency. Against the previous head's package, the same new cases produce three failures and one preservation pass; all four pass on this revision.

```sh
ARCHIFY_CHROME=/path/to/chrome node --test --test-concurrency=1 test/desktop-reader-browser.test.mjs test/finder-browser.test.mjs test/focus-browser.test.mjs test/route-probe-browser.test.mjs test/guided-views-browser.test.mjs
```

Real Chrome 150: **45 passed**, no failures or skips.

Local `npm test` is **not green**: 1,679 tests, 1,560 passed, 42 failed, 77 skipped, no cancellations. Viewer/brand-mark/validator/release-identity/golden preflights passed. Browser coverage was run separately as above.

Thirty-one failure names match the September 11 Windows run; that is a historical name comparison, not a current-dev baseline. For the 11 failure names not present in that run, fresh isolated candidate/current-dev runs both yield 1 pass and 10 failures: atlas symlink creation fails with `EPERM`, the Windows archive-path test fails in its build helper, and eight repository-evidence replacement fixtures fail during Git initialization (`\\.\nul`). The delivery-lock `empty-write` case passes on both isolated runs, so the cause of its original full-suite failure remains unresolved. This is a focused comparison, not a claim that a full dev baseline was run. No unrelated assertion, timeout, or product behavior was changed to make the suite pass.

Final-head [CI run 35193142996](https://github.com/tt-a1i/archify/actions/runs/35193142996) is **successful** for `978687e7d4067acb406db0a2990b6a1f5fcc3d5c`: 11 jobs passed, including Node 18/20/22/24, three-platform package smoke, Windows portability, ZIP freshness, and the shared browser/WebM gate. Pages deployment was conditionally skipped. No workflow approval is pending; maintainer re-review remains outstanding.

## Browser and compatibility evidence

Reproduce the representative component-to-boundary-to-boundary-to-component example from the repository root:

```sh
node archify/bin/archify.mjs deliver architecture docs/issue-284-visual-evidence/boundary-endpoints.architecture.json /tmp/boundary-endpoints.html --quality showcase --json
ARCHIFY_CHROME=/path/to/chrome node archify/bin/archify.mjs visual-check /tmp/boundary-endpoints.html --json
```

- Fresh showcase delivery passes 9/9 checks. Chrome visual-check passes at 1440x900, 1600x1000, 1920x1080, and 2048x1320. The 1440x900 light/dark captures were inspected separately for readable titles/labels, frame alignment, and clipping.
- Finder/Focus and Route Probe pass nine operations: the three-edge `client -> application-zone -> data-zone -> audit` path works; six membership-only routes remain unreachable; both named boundaries are selectable.
- The new long-title regression case, with `labelDx: 1000` and `labelDy: 150`, passes public showcase validate/deliver and all four Chrome viewports. At 1440x900 in both themes, titles fit the final frames, projected boundary title size is about 8.684px (above the 6px floor), the displaced label fits the 1665x379 auto canvas, and route endpoints meet final frame side centers within 0.01 screen px. This is a geometry stress case, not a proposed ideal visual layout.
- The same input with a 1000x500 authored viewBox is rejected by validate and deliver with `composition/label-canvas-containment`; no artifact was forced past validation.
- Frozen component-only `web-app` and `production-deployment` inputs produce byte-identical diagram SVG on current dev and this revision.

The checked-in screenshots remain illustrative captures from original head `c4fe801`; they are not relabelled as the fresh browser evidence above.

## Generated package

The current official tracked-only stager and Node 22 ZIP writer (including its mode manifest) were used through a local Windows plaintext I/O adapter, with no archive postprocessing. A control build exactly reproduces current dev's tracked ZIP. Two candidate builds are byte-identical; all 84 entries match staged source bytes/modes and pass CRC checks. Candidate SHA-256: `68672577c7214158a349b2064bd5a55f90336e147f00a1d44fe282bdf2c3319f`.

Extracting outside the repository without `node_modules`, then running showcase validate/deliver, passes all 9 checks with no errors or warnings; delivered HTML matches the worktree artifact. Generated-validator freshness and `git diff --check` pass.


anubhavp94:
Please resolve these conflicts and merge this change. Would be very helpful.

anubhavp94:
Gentle reminder on this. Seems like some tests are pending.

tt-a1i (changes requested):
The feature is worthwhile: an explicitly named zone can expose an outward-facing interface without inventing a proxy component. Keeping `wraps` as membership rather than implicit graph edges is a sensible scope boundary. I ran the boundary-endpoint, architecture-delta, and engineering-profile suites at `9de26f130cae6a344fcfd52451359ed54c228938`: 46 passed, zero skipped.

Required integration before acceptance (not a reproduced defect on the original head): please update onto current `dev`, which now includes the shared architecture router (#266), connection-label canvas containment (#236), and dominant-axis endpoint inference (#379). The old inline-router patch cannot be accepted unchanged:

- Current rendering measures routed connection labels before boundary-title/canvas convergence. This PR creates boundary endpoints from the final resolved frames. Integrate these so routing, label measurement, and the final drawn boundary use the same geometry, including a long boundary title and a boundary-connection label extending past the component extent. An automatic canvas must contain the label; a deliberately too-small explicit viewBox must still report its diagnostic.
- Preserve the distinction between endpoint lookup and component obstacles when using the shared router. Passing all boundary frames as ordinary component obstacles would change route acceptance. Retain the existing explicit `via`/side behavior and outward-only contained-endpoint diagnostic.
- Keep the existing component inventory/legend/Delta distinctions covered, then run the relevant current-dev tests and a browser check of a representative component-to-boundary and boundary-to-boundary scene. Regenerate affected validators/package outputs and obtain green final-head CI.

The author owns this bounded integration follow-up. No need to broaden this into implicit containment edges or nested boundary membership. This stays open targeting `dev`; `main` promotion is separate.

THE-NAVVV:
Tested this against my use case from #410 (subsystem-to-subsystem edges without proxy components).

Setup: a "Payments Service" boundary wrapping an API component and a Retry Worker component, with a connection from a Client component straight to the boundary, and another from the boundary to a Database component.

Result: works as intended. Both connections resolved cleanly against the boundary's frame - render passed all 9 validation checks, and inspecting the output SVG confirmed the edges terminate on the boundary's actual rect edges (not reaching into an internal component). This is exactly the abstract-level connection semantics described in #284 and #410.

One thing worth flagging for anyone else testing: wraps still only accepts component ids, not another boundary's id (tried it, got "wraps unknown component"). Boundary containment currently only happens by accident, when two boundaries' wraps lists happen to overlap geometrically. That's a separate, pre-existing gap (tracked in #410) - not a regression from this PR, just noting it since it's adjacent to what this PR touches.

No issues found with the boundary-endpoint behavior itself. Would be good to see this merged.

tt-a1i:
A real-world test on #284 now gives us a specific boundary-endpoint failure. In a three-component/two-boundary repro, a bottom boundary and a top boundary overlap on the horizontal axis, but automatic inference chooses `right`/`left`; the generated route then fails `clean-flow/endpoint-side-direction` and follows the target border long enough to trigger `composition/container-border-run`. Explicit `fromSide: "top"` / `toSide: "bottom"` works. Please include that minimal spec from [the #284 report](https://github.com/tt-a1i/archify/issues/284) as an automatic-routing regression, while retaining the existing explicit-side path and obstacle rules. Another tester confirmed the basic boundary attachment for #410, so this is a focused gap rather than a rejection of the feature. I have not independently reproduced the reported failing spec yet; the contributor's exact diagnostics are the current evidence.

PR #322, opened 2026-09-05
fix(cli): prevent non-HTML output overwrites without restricting directories

### LINKED ISSUE #124: [Bug]: CLI deliver output path can overwrite non-HTML files (meta.output sandboxed; CLI is not)

tt-a1i:
## Problem and value

Fixes #124. A command such as `archify deliver workflow input.json marker.env` could report success after replacing an existing non-HTML file with HTML. The shared output resolver now rejects mismatched artifact and receipt extensions before writing and during existing pre-commit rechecks.

Related implementation and discussion: #125. This PR preserves explicit absolute/parent-directory output paths and the public validation/inspection paths while addressing the reported overwrite. #125 remains available for its original discussion and author attribution.

## Scope

- HTML outputs from render, deliver, preview, and compare require `.html`; compare receipts require `.json`. Checks cover authored CLI suffixes and resolved symlink targets, case-insensitively.
- Return `output/cli-extension` or `output/cli-resolved-extension` with structured evidence and a repair suggestion. Compare checks artifact/receipt aliases before extension validation in both preparation and pre-commit checks, including the derived default receipt path.
- Give layout-only validation/inspection an unused typed HTML target instead of `/dev/null`; layout mode continues to emit JSON without writing that target.
- Update race and commit-failure fixtures to start with valid HTML targets, preserving their original failure-stage assertions. Document the output contract and rebuild the Skill ZIP.
- No schema, viewer, renderer geometry, release identity, dependencies, or workflow changes.

## Stability impact

Explicit nonstandard output extensions now fail, including new files; callers must choose `.html`/`.json`. Absolute paths and paths outside cwd remain supported. Authored `meta.output` retains its existing cwd sandbox. This is file-type misuse protection, not a sandbox for explicit CLI paths or a promise to preserve existing files with the expected extension. Reverting this commit restores prior behavior.

## Tests run

- Diagnostic-precedence regressions on prior PR head `117ef92`: 1 passed / 3 failed. The failing direct, explicit-symlink, and default-receipt-symlink cases returned `output/cli-extension` instead of `output/target-alias`; distinct invalid targets retained their extension diagnostic. All four pass after the fix and preserve existing bytes.
- Linux Node 22.23.2, from `archify/`: `node --test test/cli-output-types.test.mjs test/output-path.test.mjs` — 32 passed, 0 failed, 0 skipped.
- Linux Node 22.23.2, from `archify/`: `npm test` — 1045 total, 1018 passed, 0 failed, 27 skipped; generated-contract and golden checks also passed.
- Canonical Node 22: `bash scripts/build-zip.sh`; rebuild to a separate path and `cmp` — byte-identical.
- Extracted ZIP outside the repository: `node scripts/package-smoke.mjs <extracted-skill>` — passed on macOS without installed package dependencies.
- `git diff --check` — passed.

Earlier validation at `117ef92` included 73 related Linux tests passing. One initial full run at that head failed in the unchanged update-notifier concurrency test; unmodified main passed that file 87/87 in isolation and a full candidate rerun passed. The current full run above passed without a retry.

## Visual evidence

Not applicable: output-path validation only; no visible artifact changes. Browser-specific full-suite skips are reported as skips.

## Generated artifacts

`archify.zip` rebuilt with canonical Node 22. Only the CLI, output resolver, and delivery-contract document changed inside the 77-file archive; no unrelated artifacts regenerated.

## Checklist

- [x] Focused change with explicit compatibility impact.
- [x] Behavioral regression coverage and focused validation.
- [x] Full `npm test` completed; skips and the initial transient failure are disclosed.
- [x] Canonical archive rebuilt and freshness verified.
- [x] No secrets, private content, or unrelated files included.


archify/renderers/shared/output-path.mjs
```diff
@@ -314,6 +315,24 @@ export function resolveOutputPath({
...
+  // Keep explicit CLI directories unrestricted, but reject mistaken file types.
+  // Alias checks above retain priority when a target would overwrite an input.
+  if (source === 'cli') {
+    const resolvedOutput = canonicalFuturePath(outputPath);
+    const authoredMatches = path.extname(rawOutput).toLowerCase() === requiredExtension;
+    const resolvedMatches = path.extname(resolvedOutput).toLowerCase() === requiredExtension;
+    if (!authoredMatches || !resolvedMatches) {
+      const message = `CLI output must ${authoredMatches ? 'resolve to' : 'target'} a ${requiredExtension} file.`;
+      throw new OutputPathError(message, {
+        code: authoredMatches ? 'output/cli-resolved-extension' : 'output/cli-extension',
```
sunsunsun-java, line 327: [P3, non-blocking] Preserve target-alias precedence when the artifact also has an invalid extension.

The PR says existing input/target-alias diagnostics retain precedence, but compare resolves the HTML target before passing the receipt into the output-alias checks. This new throw can therefore prevent the pairwise alias check from running.

Reproduction from `archify/`:
```bash
node bin/archify.mjs compare architecture \
  examples/checkout-platform.base.architecture.json \
  examples/checkout-platform.head.architecture.json \
  same.json --receipt same.json --json
```
On baseline `d8e4daf`, this returns `output/target-alias`; on `117ef92`, it returns `output/cli-extension`. Both exit 1 at the prepare stage without writing, so this is a diagnostic-contract issue rather than a file-protection failure.

Consider checking the two output paths for aliases before checking their extensions, and adding a regression covering this combination.

sunsunsun-java (commented):
Reviewed 117ef92 against d8e4daf. The extension gate addresses the overwrite reported in #124 while preserving explicit absolute and parent-directory outputs. I found no blocking issues; one non-blocking diagnostic-precedence regression is noted inline.

Validation: all 73 focused tests across cli-output-types, output-path, cli, and cursor-onboarding were verified on macOS / Node 22.23.1 (72 passed in the sandbox; the preview test initially hit the sandbox's loopback-listen restriction and passed when rerun outside it). The ZIP contains the same 77 entries, with only the three expected files changed and their contents matching the candidate source.

sunsunsun-java (approved):

PR #325, opened 2026-09-06
fix: keep packaged examples within desktop reader

### LINKED ISSUE #324: 3 of 5 packaged examples fail archify's own visual-check containment gate
## Summary

Three of the five packaged `examples/*.html` files fail `archify visual-check` — the containment gate the skill itself enforces — at **every** desktop viewport.

This matters more than a normal doc nit because `SKILL.md` instructs the agent to read an example **before** authoring ("Read one matching schema in `schemas/`, `schemas/common.schema.json`, and one matching JSON example in `examples/`"). An agent that infers geometry from those examples inherits a layout that cannot pass `deliver` + `visual-check`, and only discovers it after a full authoring loop.

## Reproduction

```bash
cd <skill dir>
for f in examples/*.html; do
  echo "$f"
  node bin/archify.mjs visual-check "$f" --json | python3 -c "import json,sys; d=json.load(sys.stdin); print(d['containment']['status'])"
done
```

## Result

| Example | containment | viewports passing | worst scrollHeight |
|---|---|---|---|
| `dataflow-product-analytics.html` | **fail** | 0/4 | 1457 |
| `lifecycle-agent-run.html` | **fail** | 0/4 | 1399 |
| `sequence-cache-miss-request.html` | **fail** | 0/4 | 1751 |
| `web-app-rendered.html` | pass | 4/4 | — |
| `workflow-agent-tool-call-rendered.html` | pass | 4/4 | — |

All failures are vertical only (`overflowX: false` everywhere); the diagnostic is `viewer/viewport-overflow`, e.g. `scrollHeight 1302` vs `innerHeight 900` at 1440x900 for `dataflow-product-analytics`.

## Environment

- archify `2.17.0-dev.1`, commit `d8e4daf2610d512821365f41b139d874b29efe81`
- Node v26.8.1, macOS 15 (darwin arm64), Google Chrome via the skill's own `findChrome`
- Unmodified skill package; `ARCHIFY_UPDATE_CHECK_DISABLED=1` set (no other env)

## What made this expensive in practice

While authoring five diagrams I hit the same wall repeatedly, and the measured constraints were not obvious from the docs:

1. The dataflow renderer pins stage centres and row tops at **fixed** pixel positions. `meta.viewBox` sets the canvas and aspect ratio but does **not** rescale them, so the row count — not the viewBox — determines the minimum height. Empirically: 3 rows needs viewBox height ~505-520, 4 rows ~606, and **5 rows cannot fit at 1440x900 with cards at all**.
2. Rows appear to be **0-indexed**. A spec starting at row 1 silently wastes a row and can push an otherwise-fitting layout over the limit.
3. With sublabels present, viewBox **width** is effectively capped near 1085: the authored sublabel font is fixed and `composition/desktop-readability` assumes ~930px of available diagram width, so a wider canvas pushes text under the 6px floor. Widening to fix a *height* problem therefore trades one hard error for another.

If (1) is intended, it would help a lot for the docs to state the row→minimum-height relationship explicitly, since `validate` can pass 9/9 with 0 errors and 0 warnings while the delivered artifact still fails containment.

## Suggested fix

Either regenerate the three failing examples so the shipped set passes the skill's own gate, or add a CI check that runs `visual-check` over `examples/*.html` so they cannot drift out of compliance again.

Happy to send the exact diagnostics JSON for any of the three.

zhouyuanxinand:
## Problem and value

Three of the five packaged HTML examples overflowed Archify's own desktop reader at every viewport exercised by `visual-check`. This makes the shipped examples fail the same containment gate users are expected to trust.

Closes #324.

## Scope

- What changed:
  - Compacted the dataflow example into three rows and rerouted its affected flows.
  - Widened and vertically compacted the sequence example, using spread participant columns.
  - Reduced the lifecycle summary dock from three verbose cards to two concise cards.
  - Added a real-browser regression that runs every packaged HTML example through the full desktop `visual-check` viewport matrix.
  - Regenerated the affected packaged/root HTML, Gallery artifacts and manifest, and deterministic Skill ZIP.
  - Renderer, schema, validator, and Viewer behavior are unchanged.
  - The two packaged examples that already passed were not regenerated.

## Stability impact

- Compatibility and migration risk: none; the changes are limited to example inputs, their generated outputs, and regression coverage.
- Renderer, validator, package, or generated-artifact risk: low. Existing exact fixed-column sequence tests now explicitly request `column_fit: fixed` instead of inheriting that choice from the changed example fixture.
- Failure behavior and rollback path: CI now fails when any packaged HTML example fails the desktop containment gate. Reverting this commit restores the previous examples and CI coverage.

## Tests run

- Node v22.23.2: `node --test test/brand-marks.test.mjs test/layout-rules.test.mjs test/gallery.test.mjs` — 127 passed, 0 failed.
- Node v22.23.2 with a real Chrome runtime: `node --test test/desktop-reader-browser.test.mjs` — 2 passed, 0 failed. The packaged-example test covers all five HTML examples at 1440x900, 1600x1000, 1920x1080, and 2048x1320.
- `node scripts/generate-brand-marks.mjs --check` — passed.
- `node scripts/generate-validators.mjs --check` — passed.
- `node ../scripts/check-release-identity.mjs` — passed (`2.17.0-dev.1`).
- `node test/golden.mjs` — all renderer, packaged-example, schema, template-freshness, and version-sync checks passed.
- Node 22 deterministic ZIP rebuild performed twice — exact byte match; SHA-256 `946736334D9A9DA50DB8A64ED5B301AA763F1AA2335B4A05CA95CD01F62C6988`.
- `npm test` was also run on the Windows development host. Its first run reported 960 passed, 35 failed, and 38 skipped. Six fixture-coupling failures introduced by the example changes were fixed and are included in the 127/127 targeted rerun above. The other 29 failures are existing host limitations or timing failures involving Windows symlink privileges, the `\\.\nul` Git path, unavailable `/bin/bash`/`unzip`, preview signal behavior, and notifier concurrency. The authoritative Ubuntu/Node matrix remains delegated to CI.

## Visual evidence

- Evidence provided: regenerated HTML/Gallery artifacts in this diff, the reproducible real-browser command above, and the containment receipt summary below.
- Comparison conditions: packaged examples in Chrome at the four standard desktop viewports, with the adaptive reader enabled. Light 1440x900 and dark 2048x1320 screenshots were also inspected for each changed example.
- Automated or browser checks: all 12 originally failing example/viewport combinations now pass. For dataflow, lifecycle, and sequence respectively, document scroll height equals viewport height at 900, 1000, 1080, and 1320 pixels.
- Perceptual visual review: passed. No clipping, route/label collisions, or unreadable text was observed in the inspected light and dark outputs.

## Generated artifacts

- `archify/examples/{dataflow-product-analytics,lifecycle-agent-run,sequence-cache-miss-request}.html`
- `examples/{dataflow-product-analytics,lifecycle-agent-run,sequence-cache-miss-request}.html`
- `docs/gallery.html`
- `docs/gallery/manifest.json`
- The three corresponding files under `docs/gallery/artifacts/` and `docs/gallery/sources/`
- `archify.zip`

## Checklist


sunsunsun-java:
The fix looks good. It makes the three examples more compact while preserving their node relationships and message order, without changing the renderer or relaxing validation rules. The new browser regression test also helps prevent shipped examples from failing their own checks again.

I reran the 127 targeted tests, the real-browser tests, and the generated-artifact consistency checks—all passed. The browser tests cover all five examples at four desktop viewport sizes, confirming that the reported overflow is resolved.

Before merging, please resolve the current merge conflicts and rerun CI on the updated branch. There are currently no remote CI results, so integration still needs verification.

One optional documentation improvement: explain that changing `viewBox` does not automatically reposition nodes, and widening the canvas can make text appear smaller. That would help users understand the layout constraints when adapting these examples.

sunsunsun-java (approved):

PR #321, opened 2026-09-05
fix(checker): flush large artifact receipts before exiting

jackie-cqz:
## Problem and value

Fixes #311.

When the artifact checker writes a large JSON receipt to a pipe, its immediate `process.exit()` can terminate Node before stdout drains. Valid diagrams then fail `validate`, `deliver`, or `compare` with truncated-JSON errors. On Linux Node 22.22.0, the regression fixture's 262,519-byte successful receipt was truncated to 146,176 bytes before this fix.

## Scope

- What changed: use `process.exitCode` after emitting the final checker receipt; add five regressions covering successful and failing large receipts and all three consuming commands; update the changelog.
- What deliberately did not change: receipt fields, validation rules, exit codes, renderer geometry, Viewer behavior, or release version.
- No unrelated changes: confirmed. Related PR #314 includes overlapping checker work alongside CLI argument and preview changes; this PR is limited to #311.

## Stability impact

- Compatibility and migration risk: no schema or CLI contract changes; no migration required.
- Renderer, validator, package, or generated-artifact risk: only the checker exit strategy changes. Both exit 0 and exit 1 receipts are compared byte-for-byte between file and pipe output.
- Failure behavior and rollback path: invalid artifacts still return exit 1 with complete diagnostics. Delivery and comparison retain their existing validation and atomic replacement paths.

## Tests run

Revalidated after merging upstream `main` at `ed7f4d4`; the PR head is `4d6ecb2`.

Linux / WSL, Node 22.22.0, from `archify/`:

```sh
npm test
```

- Full suite: **1023 passed / 0 failed / 27 skipped** out of 1050 tests, including all five #311 regressions and the upstream output-path tests.
- Brand mark generation, generated validators, release identity, and golden artifact checks also passed.
- Before the original production fix, `node --test test/artifact-receipt-flush.test.mjs` produced **0 passed / 5 failed** on Linux; after the fix, all five passed.
- Environment-dependent skips are not counted as passes.

Windows, Node 24.15.0, from `archify/`:

```sh
node --test test/artifact-receipt-flush.test.mjs test/cli-output-types.test.mjs test/output-path.test.mjs
```

**23 passed / 10 failed / 4 skipped**. All five #311 regressions passed. The ten failures occur at fixture setup with `EPERM` from `fs.symlinkSync` because this Windows account lacks symbolic-link privileges; four other symlink tests explicitly skip. These failures are not counted as passes. The same output-path tests pass in the Linux full suite above.

From the repository root with Node 22.22.0:

[...]

Issue #344, opened 2026-09-08
Tracking: improve first-diagram reliability, readability, and maintenance focus


tt-a1i:
## Objective

Make an ordinary Agent's path from a bounded system description to a readable, verified diagram more reliable and less expensive to repair. Use the existing issues below as the implementation owners, with one selected approach per behavior.

Assessment baseline: `920543baa1c6137803c5b45a69d8977152773d35`, checked on 2026-09-08. Local reproduction used that commit's tracked `archify.zip` (`2.17.0-dev.1`) on macOS, Node `v26.3.0`, and Chrome. The [CI run for the same commit](https://github.com/tt-a1i/archify/actions/runs/34122654563) passed.

## Current maintenance queue — 2026-09-16

Latest inventory before this cleanup: 77 open issues and 102 open PRs; 42 open issues have an explicitly linked implementation PR. The immediate work is to finish and consolidate existing candidates. Keep at most three issue workstreams active. A primary implementation label selects the review path; it is not approval or permission to bypass required checks.

| Active issue | Primary implementation | Maintenance owner / next step | Acceptance boundary |
| --- | --- | --- | --- |
| #420 pinned evidence must ignore Git replacement objects | #421 | @tt-a1i coordinates current-head review; use the existing PR | Original SHA semantics survive commit/blob replacement and custom replacement refs; preserve valid evidence and user Git configuration. |
| #423 valid semantic IDs must work in the Viewer | #424 | @tt-a1i coordinates current-head review; use the existing PR | Validate all five modes, focus/chapter/route behavior, ordinary IDs, browser CI execution and regenerated artifact consistency. |
| #375 failed delivery must not make an old artifact look current | #380 | @tt-a1i re-reviews the updated candidate; earlier requested changes refer to an older head | Recheck failure provenance, output boundaries, transaction/rollback and alias cases on the latest head. A passing happy-path example does not establish failure recovery. |

Issue assignment to @tt-a1i denotes maintenance/review coordination; it does not transfer authorship or ask contributors to start a replacement implementation. No PR is approved or merged by this triage pass.

### Next queue and consolidation

- #427 browser gate wiring, then #429 compiler validation parity and #430 profile metadata consistency, become candidates when an active slot is available. These are current-main defects, not a new library API workstream.
- #431 and #432 are acceptance issues for unmerged feature PR #367. Keep them attached to that feature's readiness; do not describe them as released regressions.
[...]

PR #351, opened 2026-09-08
fix(renderer): flush complete failure diagnostics before exiting

### LINKED ISSUE #350: Renderer failure diagnostics are truncated through pipes, losing Agent repair hints
## Problem

On current `main` (`06bd6fe`), a large renderer failure loses its structured repair diagnostics when stderr is a pipe. `validate --json`, `deliver --json`, and `compare --json` then report `internal/unclassified` instead of the actual schema rule and supported fixes. This blocks the normal Agent repair loop on an invalid input with many repeated mistakes.

Reproduced on macOS arm64 with Node v22.23.2 (also observed on Node v26.7.0), using the source checkout. The input is 16,426 bytes; the complete failure is 107,256 bytes with 200 `schema/additionalProperties` diagnostics. File-directed stderr contains all of it, while piped stderr contains exactly 65,536 bytes of incomplete JSON. All processes correctly exit 1, and an existing delivery target is preserved; the defect is loss of repair evidence, not a false successful delivery.

## Reproduction

Run this from the repository root. It creates and removes only a temporary fixture directory and uses no external services or private data.

```js
// Save as reproduce.mjs in the repository root, then run: node reproduce.mjs
import fs from 'node:fs';
import os from 'node:os';
import path from 'node:path';
import { spawnSync } from 'node:child_process';

const tmp = fs.mkdtempSync(path.join(os.tmpdir(), 'archify-stderr-repro-'));
try {
  const input = path.join(tmp, 'input.json');
  const output = path.join(tmp, 'output.html');
  fs.writeFileSync(input, JSON.stringify({
    schema_version: 1,
    diagram_type: 'architecture',
    meta: { title: 'Large schema diagnostic regression' },
    components: Array.from({ length: 200 }, (_, i) => ({
      id: `node-${i}`, type: 'backend', label: 'Node',
      pos: [40, 40], unexpected: true,
    })),
    connections: [],
  }));
  const args = ['archify/renderers/architecture/render-architecture.mjs', input, output];
  const env = { ...process.env, ARCHIFY_DIAGNOSTIC_FORMAT: 'json' };
  const reference = path.join(tmp, 'reference.json');
  const fd = fs.openSync(reference, 'w');
  try {
    spawnSync(process.execPath, args, { env, stdio: ['ignore', 'pipe', fd] });
  } finally {
    fs.closeSync(fd);
  }
  const piped = spawnSync(process.execPath, args, { env, encoding: 'utf8' });
  console.log({ fileBytes: fs.statSync(reference).size, pipeBytes: Buffer.byteLength(piped.stderr) });
  const result = spawnSync(process.execPath, [
    'archify/bin/archify.mjs', 'validate', 'architecture', input, '--json',
  ], { encoding: 'utf8' });
  console.log(JSON.parse(result.stdout).diagnostics);
} finally {
  fs.rmSync(tmp, { recursive: true, force: true });
}
```

Expected: a complete renderer JSON failure reaches the parent; validation identifies `schema/additionalProperties`, the affected nodes, and `remove unsupported property "unexpected"`.

Actual: the parent reports `Renderer failed before emitting a structured diagnostic.` with `internal/unclassified` and no supported fixes.

## Cause and proposed scope

`installRendererDiagnosticBoundary()` in `archify/renderers/shared/diagnostics.mjs` calls `fs.writeSync(process.stderr.fd, payload)` once and immediately exits. A pipe can accept only a prefix of the payload; the returned byte count is ignored. The same shared boundary is used by all five renderers.

Proposed fix: flush the existing stderr JSON before terminating with status 1, retaining termination on a stream failure. Add real pipe-versus-file regression coverage and public CLI assertions for preserved repair diagnostics and unchanged trusted outputs. Keep schemas, error fields, quality rules, renderer geometry, and successful artifacts unchanged.

This is separate from #311/#321, which fixed the artifact checker's stdout exit path. The renderer's stderr boundary still uses the one-write path. I found no open PR covering this boundary.

Visual evidence: Not applicable; diagnostic transport only.

276970789:
## Problem and value

Fixes #350.

On macOS/Node 22, a 16 KB invalid input produces 200 schema diagnostics (107,256 bytes), but piped stderr receives only 65,536 bytes. `validate`, `deliver`, and `compare` then report `internal/unclassified` instead of the actual errors and repair hints. #350 has the reproduction.

## Scope

Replace the single `fs.writeSync()` with a stream write and exit 1 after its callback. #321 fixed checker stdout; this fixes renderer stderr. The three changed files are the shared boundary, regression tests, and Skill ZIP.

Since reviewed head `af34f53`: guard duplicate fatal writes, test two exceptions under backpressure, and sync `main@b86b607`. Current head: `b3b2ffc`.

## Stability impact

Shared behavior across all five renderers. Schemas, error fields, quality rules, buffer limits, and successful rendering stay unchanged. Stream errors still exit 1; active handles cannot keep the renderer alive after the write completes. Failed delivery and comparison preserve existing outputs. Rollback is a revert and ZIP rebuild.

## Tests run

macOS arm64; base `b86b607`, candidate `b3b2ffc`. Commands run from the repository root unless noted. These are local results.

- Initial regression baseline (`06bd6fe`): 1 passed / 8 failed. The new concurrent-exception test also failed on reviewed head `af34f53` before adding the guard.
- `node --test archify/test/renderer-diagnostic-flush.test.mjs archify/test/repair-receipt.test.mjs`: 16 passed / 0 failed / 0 skipped on Node 18.20.8, 20.20.2, 22.23.2, and 24.20.0, using `npx --yes node@<version>` for alternate versions.
- From `archify/`, `ARCHIFY_CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" npm test`: 1076 passed / 0 failed / 4 skipped on official Node 22.23.2, including package/release, generated-file, and golden checks.
- Two `scripts/build-zip.sh` builds and `cmp`: identical. `node scripts/package-smoke.mjs <extracted-skill>` passed without installed dependencies; packaged validate/deliver retained all 200 diagnostics and preserved existing output. `git diff --check` passed.

[Remote CI for this head](https://github.com/tt-a1i/archify/actions/runs/34219517025) awaits maintainer approval.

## Visual evidence

Not applicable: diagnostic output only; no appearance or interaction changes.

## Generated artifacts

Only `archify.zip`: official Node 22.23.2, 78 entries, 1,327,741 bytes (+137 against current main). The entry list matches main; only `renderers/shared/diagnostics.mjs` differs, and its bytes match the source. Release identity is unchanged.

## Checklist

- [x] Focused change; existing typed JSON preserved.
- [x] Targeted and full tests completed.
- [x] Regression tests fail before the fix.
- [x] Package rebuilt and verified.
- [x] No sensitive data in the submission.


tt-a1i:
Current-head scope review: `b3b2ffc2ac072ede65f78298dec0142dbe1a59fc`, compared with current main `d673e8300df60a5c8166abe78787fdc78f6b8000`.

#374 has resolved #350 on main. The original reproduction now receives all 107,256 bytes through stderr, and the three boundary regressions pass. Your additional regression coverage still has value, so I am keeping this PR open for a narrower test-only revision rather than closing it as a duplicate.

Please rebase onto current main and:

- Remove the superseded diagnostic runtime change and ZIP rebuild. Remove `Fixes #350` and describe this as additional regression coverage; #350 is now closed with verification evidence.
- Retain the useful end-to-end coverage for the five renderer entrypoints and validate/deliver/compare, plus closed-reader/active-handle behavior, avoiding redundant copies of the existing boundary tests.
- Make the concurrent-error test independent of the asynchronous implementation. Its parent currently waits for IPC from the second callback before draining stderr; a valid synchronous writer can block before that callback runs, producing a test-created deadlock. Have the parent resume reading after a bounded delay independently of child IPC, then assert one complete receipt, nonzero exit and no hang.

Read-only verification of the PR tests on main using Node 26.8.1: the nine non-implementation-bound tests from this PR passed. The original concurrency test timed out under the synchronous implementation; a separate delayed-reader probe with two queued exceptions and an active timer passed when pipe draining did not depend on that second callback. The current main boundary suite also passed 3/3. These are focused local checks, not fresh CI or full-suite acceptance of a revised PR.

Marking this awaiting-author for the scope revision above. It is queued regression work outside the three active issue workstreams in #344; this comment does not approve a merge.

tt-a1i:
Closing as superseded by merged #374 (d673e83). Current main drains complete renderer receipts, handles EAGAIN backpressure, and preserves the original failure exit. I ran this PR’s regression file against main: its 9 public-behavior/closed-reader cases pass. The remaining test deliberately waits for a second in-process callback before it starts reading stderr; that handshake specifically assumes this PR’s asynchronous writer and deadlocks against main’s synchronous drain, so it is not evidence that normal receipt delivery remains broken. There is no need to maintain a second implementation for #350. A separately reproduced normal-caller defect can be tracked on its own.

PR #363, opened 2026-09-08
docs(authoring-contract): add viewport height budget and concrete via direction contract

ljcjeans-dev:
## Problem and value

Current-main trigger: a candidate that passes `validate --quality showcase` with 0 errors / 0 warnings can still overflow the 1440×900 first screen in `visual-check`. The docs describe the goal qualitatively ("over-compressed Y layout", "vertical rhythm") but give no quantitative budget, so an author rediscovers the page-height arithmetic by trial and error after the spec is already frozen. Separately, the via direction contract is enforced by `clean-flow/endpoint-side-direction` but stated only abstractly ("a side is a direction contract"), so the exact constraint on `via[0]` coordinates is learnable only by triggering the error.

Intended outcome: an author can budget the viewport height at authoring time and write compliant `via` waypoints on the first try.

Evidence from a real authoring session on 2.16: official example `web-app-rendered.html` at viewBox 1080×588 (ratio 1.84) passes all four viewports, while my candidate at 1225×648 (ratio 1.89) overflows 1440×900 by 79px. A higher ratio overflowing while a lower one passes proves "keep the ratio above a threshold" is the wrong mental model; the determinant is absolute rendered height (viewer chrome ~51px + 930px reader-width × viewBoxH/viewBoxW + cards). Tightening card copy recovered 34px; stretching node x-coordinates (viewBox 1225→1315) recovered the remaining 11px.

Approach: docs-only — one new subsection and one bullet extension (below). Narrow documentation correction; per CONTRIBUTING "Choose the right path", no separate planning issue needed.

## Stability impact

- Impact class: documentation only. No code, schema, defaults, or acceptance rules changed.
- Existing behavior preserved: all existing text unchanged except the one direction-contract bullet, which is strictly extended (original sentence retained verbatim).
- No unrelated changes: confirmed — single file, +23/−1.

## Tests run

- Comparison base: `main` @ 1072200. Candidate: this branch (1 commit).
- Markdown structure verified: "Viewport height budget" nests under "Executable geometry rules" alongside "Spacing and labels" / "Repair order"; no heading regressions.
- No automated checks apply to prose; `archify doctor` unaffected (docs not loaded by the CLI).

## Visual evidence

Not applicable — documentation-only change; no rendering path touched.

## Generated artifacts

None; no generated files reference `references/authoring-contract.md` content.

### DIFF
--- a/archify/references/authoring-contract.md
+++ b/archify/references/authoring-contract.md
@@ -104,12 +104,34 @@ in the generated viewer.
 ## Executable geometry rules
 
 - Node anchors start at side midpoints. `left`/`right` change the horizontal endpoint; `top`/`bottom` change the vertical endpoint. For an automatic Architecture relationship, unobstructed facing ports whose axis offset is under 16px may share one horizontal or vertical axis when both endpoints retain the 16px corner gutter. If exactly one endpoint belongs to a spread group, only its unshared counterpart moves; relationships spread at both endpoints keep their distinct ports and outside bridge.
-- A side is a direction contract. The first and final route segment must be perpendicular and outward/inward in the named direction.
+- A side is a direction contract. The first and final route segment must be perpendicular and outward/inward in the named direction. Concretely, for an explicit `via`: `via[0]` must share its x with the departure point when `fromSide` is `top`/`bottom` (vertical first segment) and share its y when `fromSide` is `left`/`right` (horizontal first segment); the same holds inverted for the arrival point and the last `via` entry. Departure and arrival points sit at side midpoints — they are not freely choosable positions on the edge.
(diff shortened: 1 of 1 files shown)


tt-a1i:
The useful guidance is now covered by the integrated layout-repair guide in #371 and the existing delivery contract: explicit via points are ordered absolute waypoints between node anchors, endpoint directions must be respected, and viewport acceptance budgets the whole page including necessary cards. The recipe also directs authors to inspect the final artifact at the supported desktop sizes.

Closing this competing documentation proposal to keep one consistent repair path. Its additional fixed suggestions (aspect ratio >= 2, at most four card items, an approximately 930px reader width) are not general acceptance conditions: Viewer reading width adapts to viewport/content, and the PR's own example shows a lower-ratio diagram can fit while a higher-ratio one overflows. We should use measured page bounds and readability, not prescribe stretching the x span as a general remedy.

The distinction between geometry validation and actual browser acceptance remains important and is retained in the current Skill and delivery contract. This closure does not claim this PR's historical authoring observation was false. #364's separate concern about treating the initial 12-node budget as a semantic limit can be considered independently.

PR #370, opened 2026-09-08
feat(dataflow): multi-line sublabels and a viewBox-derived row grid

josephgimenez:
## Problem

Generating data-flow funnels with many consumers hits two hardcoded walls in the dataflow renderer:

1. **Sublabels are a single `<text>` line.** A one-line sublabel cannot carry the per-consumer detail (what the handler actually reads/writes) that makes a fan-out diagram readable; authors were forced to move detail off-canvas.
2. **The row grid is a hardcoded five rows** (`rowYs: [128, 242, 356, 470, 584]`). A taller `meta.viewBox[1]` bought nothing — a sixth node row was rejected by validation while the canvas below the fifth row stayed empty.

## Change

- A node `sublabel` may contain `\n`: each segment renders as its own `<text>` line on a 12px pitch from `y+37`. Validation measures **every line** against the node's legible width, and requires enough node `height` for the block (plus the tag band when a tag is present).
- The row grid derives from `meta.viewBox[1]`: `rowCount = max(5, floor((viewBox[1] − 74 − 128) / 114) + 1)`, keeping the fixed top (128) and pitch (114). The historical 720px canvas keeps exactly the old five rows.

## Compatibility

- Single-line sublabels render byte-identically (same `y+37`, same shrink-to-fit call per string).
- The default canvas keeps rows 0–4 at their historical coordinates; row 5+ remains a validation error there, now asserted by rejection rather than by absence of SVG (unused rows emit no SVG, so an output-only check cannot see a wrongly-grown grid).
- `archify.zip` regenerated with `scripts/build-zip.sh` on Node 22; `test/release-package-gates.test.mjs` passes.

## Tests

`test/dataflow-multiline-sublabels.test.mjs` (6 cases): tall canvas places row 7; 720px canvas keeps the historical grid and rejects row 5; multi-line sublabel renders one `<text>` per line; single-line geometry unchanged; overlong line rejected with a repairable message; too-short node rejected. Mutation-checked: widening the grid to six rows fails the rejection case.

Full suite (`npm test` from `archify/`): failure set identical to the clean-`main` baseline on this machine (52 pre-existing environment-dependent failures — network fetches, sandbox limits — zero new).

Vibe coded using Qwen 3.8 Flash Next FP8


josephgimenez:
Closing — we've vendored the patch in our own tree instead, so we no longer need this upstream. Thanks for the project.

PR #373, opened 2026-09-09
fix(checker): scope finite_svg to numeric attributes (#372)

### LINKED ISSUE #372: Bug: `artifact/finite-svg` false positive when a component `tag` contains the text "NaN" or "Infinity"
## Summary

`artifact/finite-svg` fires on specs whose coordinates are all finite, when a component `tag`
contains the literal substring `NaN` or `Infinity`. The check appears to scan rendered text for
those tokens rather than testing numeric values, so ordinary prose in a label is reported as a
non-finite coordinate.

The diagnostic carries `"evidence": {"details": []}`, so there is nothing in the output pointing
at the real cause.

## Environment

- archify `2.17.0-dev.1` (skill install via `npx skills add tt-a1i/archify`)
- Node v24.13.0, Linux

## Minimal reproduction

`repro.json` — two nodes, one connection, no evidence, no overrides:

```json
{
  "schema_version": 1,
  "meta": { "title": "Minimal repro", "quality_profile": "showcase" },
  "components": [
    { "id": "a", "type": "backend",  "label": "Solver", "tag": "returns a NaN leaf",
      "pos": [40, 40], "size": [200, 74] },
    { "id": "b", "type": "database", "label": "Result",
      "pos": [400, 40], "size": [200, 74] }
  ],
  "connections": [ { "from": "a", "to": "b", "label": "writes" } ]
}
```

```
node bin/archify.mjs validate architecture repro.json --quality showcase --json
```

## Expected vs actual

**Expected:** passes. Every coordinate in the spec is a finite literal.

**Actual:** exit 1, `artifact/finite-svg`, "Final artifact failed finite_svg."

## Behaviour matrix

Only the tag text is varied; everything else is byte-identical.

| `tag` value | result |
|---|---|
| `returns a NaN leaf` | ❌ `artifact/finite-svg` |
| `returns a nan leaf` | ✅ pass |
| `Infinity` | ❌ `artifact/finite-svg` |
| `plain text` | ✅ pass |

Case-sensitive, and matches on a substring rather than a whole token.

## Scope

- Only the component `tag` field triggers it. The same wording inside `cards[].items` or a
  component `sublabel` renders fine, which is what makes it look arbitrary from the outside.
- `finite_svg` runs after the layout checks, so it only surfaces once every
  `layout/constraint` error is cleared. While layout errors remain it is masked, and the
  failure appears to arrive out of nowhere on the round where the diagram finally becomes clean.

## Why it bites

Real diagrams of numerical code legitimately need to say `NaN` — describing a guard that returns
an all-`NaN` result is exactly the kind of thing a node tag is for. The workaround is to reword
the tag, but the error message gives no way to discover that.

## Suggested fix

Test the numeric coordinate values before serialisation instead of scanning the rendered output
for the strings `NaN` / `Infinity`. If the post-render scan is deliberate belt-and-braces, it
could be restricted to attribute values that are meant to be numeric (`x`, `y`, `d`,
`transform`, …) rather than text nodes, and populate `evidence.details` with the offending
attribute so the message is actionable.

apetcu:
## Problem and value

`artifact/finite-svg` incorrectly rejected diagrams whose prose mentioned `NaN` or `Infinity` (#372), because it searched the whole SVG. Check numeric/geometry attributes instead and identify the offending element and attribute. Authored prose no longer causes a false positive; actual non-finite geometry still fails.

Fixes #372

## Stability impact

Local behavior in the shared artifact checker, consumed by `check`, `validate`, and `deliver`. No renderer, schema, layout, Viewer, dependency, or release-identity changes.

Global and element-specific numeric attribute checks remain enforced. Text, titles, aria/data attributes, comments, and CDATA are not numeric geometry. HTML descendants of SVG integration points (including `foreignObject`) are excluded; the `foreignObject` element's own geometry and SVG descendants re-entering through `<svg>` remain checked. Numeric character references are decoded once for finite-value matching only; diagnostic evidence retains raw attribute values. Shared `attrEntries`, `parseAttrs`, and outer SVG extraction are unchanged.

## Review follow-up

Candidate `6b9357c2874ba5d2e6a70acc131cf9a07cc024f3` follows reviewed head `7935d41` and addresses all three CodeRabbit findings:

- Consume comments and CDATA as complete non-element tokens, so tag-like prose does not become geometry.
- Detect decimal/hexadecimal character references encoding non-finite tokens, without recursively decoding escaped prose or changing other attribute consumers.
- Track SVG/HTML context across `foreignObject`, HTML void elements, and nested SVG; preserve validation of actual SVG numeric attributes.

All three regression cases failed against the reviewed code before the fix. The earlier PR test counts have been replaced with current-source evidence below. Review threads remain available for reviewer confirmation.

## Tests run

Base `6db72a9aea3d0f67a6a034e41f8a5491476a11c1`; candidate `6b9357c2874ba5d2e6a70acc131cf9a07cc024f3`. macOS, official Node 22.23.2; tested working tree is identical to this commit.

- `node --test test/render-output-checks.test.mjs` from `archify/`: **33 passed, 0 failed, 0 skipped**. Covers prior prose/geometry cases plus comments/CDATA, encoded attributes, and SVG/HTML context.
- `npm test`: **1354 total; 1306 passed, 1 failed, 47 skipped**. The sole failure is `checked-in MCO artifacts are byte-reproducible from the pinned repository revision`; rerunning that test on clean base `6db72a9` also fails. The complete local suite is therefore not claimed green. Optional/browser skips are not passes.
- Extracted rebuilt ZIP outside the repository, with no `node_modules`: the same **33** checker tests passed.
- Two ZIP builds produce identical bytes, retaining all 79 entries. Relative to reviewed head `7935d41`, only packaged `scripts/check-render-output.mjs` changes. `git diff --check` passed.
- New-head hosted CI is running; inspect PR Checks for its final result. Previous revisions' CI does not establish this head's status.

## Visual evidence

Not applicable: only artifact validation changes; generated geometry, styling, and Viewer interactions are unchanged. No browser/perceptual acceptance is claimed.

## Generated artifacts

`archify.zip` rebuilt from current tracked source with Node 22.23.2. Gallery, guide, start, and README showcase remain unchanged because their rendering inputs are unchanged.

---
Original contribution prepared with AI assistance (Claude Code), reviewed and tested locally by @apetcu. Subsequent maintainer integration and review fixes were prepared with Codex; current validation is recorded above.

## Final maintainer integration

Integrated current main `50e2dcc` at final head `b2acd3d`; only the ZIP conflicted and was regenerated with official Node 22.21.1. The checker source and its 33 regression tests are unchanged from `6b9357c`. Full local `npm test` on the merged source passed; exact counts are in the final review. Required final-head CI remains the merge gate.

The additional nested-SVG extraction observation was reproduced with the same synthetic outer SVG containing a nested SVG followed by `circle cx="NaN"` on both current main and the candidate: both report `finite_svg: true`. This is a pre-existing limitation of outer SVG extraction, not a regression in this prose false-positive fix. It is recorded here as a separate follow-up, not claimed fixed by #372.


tt-a1i (approved):
Reviewed final integration b2acd3d against main 50e2dcc. This fixes #372 by separating authored prose from numeric SVG attributes while preserving non-finite geometry rejection and actionable attribute evidence. The previously addressed comment/CDATA, numeric-reference and foreign-content cases remain covered. Checker source and tests are unchanged from reviewed 6b9357c; the merge regenerated the ZIP from combined tracked source.

Local evidence: 33/33 focused checker tests passed; full npm test on the merged source with official Node 22.21.1 passed 1309 tests, 0 failures, 48 skips. No browser acceptance is inferred from local skips. Final-head ZIP freshness and platform smoke have passed; all required remote checks must finish successfully before merge.

The outside-diff nested-SVG extraction observation reproduces identically on current main and candidate. It is a pre-existing outer-extraction limitation, recorded in the PR description for separate follow-up; it is not fixed or introduced here. No remaining blocking regression found within this change.

PR #381, opened 2026-09-10
refactor(viewer): modularize source while preserving standalone HTML

sunsunsun-java:
## Problem and value

Viewer changes previously shared a 14,911-line template containing styles, export pipelines, layout, motion ownership and interaction lifecycles. Extract 13 cohesive internal source fragments and assemble them deterministically into the existing standalone HTML. The maintainer template is now 6,237 lines; consumers still receive one inline artifact.

Closes #343.

## Stability impact

- **Shared Viewer source organization:** Reader Layout, Chrome Layout, Camera, Radar, Motion Governor, Finder, Intent Trace, Semantic Lens, Route Probe, Guided Views, Focus/flow tokens, Export and private Export Cleanup now have explicit maintenance locations and contracts in `viewer/README.md`.
- Preserve classic-script scope, initialization order, shared helpers, URLs, schema, authored geometry/IDs, export formats and zero-dependency installed runtime. Assembly allows exactly one nested relationship, Export → Cleanup; invalid inputs cannot overwrite the generated template.
- Two bounded scrolling fixes followed review: selecting a Story stop used the wrong offset coordinate system, and narrowing chapter/Trail viewports could hide the active item. Use viewport-relative positioning and recenter only when viewport width changes; ordinary manual scrolling remains untouched.
- The final review also found a clipboard-test gap. Assert that ClipboardItem is constructed before the actual click handler returns, rather than relying on user activation surviving asynchronous work. Production clipboard behavior is unchanged.
- No framework, runtime asset requests, public-interface redesign, CSS split, version change or unrelated feature work.

## Tests run

**Current acceptance candidate: `1579cc946567ff7eff9b2bc20cf2ad171d4e50ae`.** Native pointer verification and the shared native-click helper preserve real browser input: the helper waits for stable, visible, unobstructed targets and verifies a trusted click reached the expected element. Finder focus and Route playback assertions remain strict. Motion initialization is scoped to each navigation; each reload verifies its new-document identity, and persisted Still mode is checked across five consecutive reloads. CI browser concurrency is bounded at two. No behavior assertion is skipped or retried.

[...]

PR #382, opened 2026-09-10
perf(template): drop backdrop-blur and filter transitions so viewer stays smooth

iqiiidi-sudo:
## Problem

Delivered HTML artifacts (~840 KB for a 17-node architecture diagram) stutter badly during pan/zoom and when opening modal panels (radar, guide, export menu). Four costs in the viewer repaint the huge SVG every frame:

1. **`backdrop-filter: blur(10–18px)` on 10 panels/modals** — the browser re-samples and re-blurs the entire SVG underneath on every frame; the main cause of freezes when opening modals.
2. **`filter` in 6 `transition` lists** (e.g. `transition: opacity 0.18s ease, filter 0.18s ease`) — animating `drop-shadow` is not composited and repaints the filter per frame on hover/focus/lens.
3. **The scene `svg` was not a compositing layer** — the pan/zoom `transform` triggered full repaints.
4. **A permanent `drop-shadow` on the radar viewport frame** — recomputed on every camera sync.

## Change

Applied in `viewer/template.source.html` (the source of truth since #381), regenerated into `archify/assets/template.html` via `npm run generate:viewer`:

- `backdrop-filter: blur(...)` → `none` in all 10 places; panel backgrounds are already semi-opaque over their own base color, so the visual delta is negligible.
- Remove `filter` from all `transition` lists; hover glow now applies instantly instead of animating a filter per frame.
- Promote the scene canvas to its own compositing layer: `will-change: transform; contain: layout style;` — pan/zoom moves a ready raster, no scene repaint.
- Drop the permanent `drop-shadow` on the radar viewport frame.

## Scope

- The global Viewer visual change is the intended product change of this PR: the stutter is inherent to the shared template, so it cannot be fixed per-diagram. There is no linked issue; scope decision is deferred to the maintainer at review.
- **Regenerated stale artifacts**: `examples/archify-repo.html`, `examples/archify-repo-grid.html`, `examples/maka-architecture.html`, `examples/rag-pipeline.html`, `docs/cases/mco-runtime.architecture.html`, `experiments/mco-showcase/mco-runtime.html`, `generated/maka-regenerated.workflow.html` (+ its `visual-check` receipt and screenshots). These were stale against the current renderer (last built before #256/#381 and this PR), so rebuilding them from the final template picks up accumulated upstream template behavior (export cleanup, source-evidence links) in addition to the CSS delta. That is required, not unrelated: with a shared template, any stale tracked render contradicts the golden/freshness tests and the claim that no tracked artifact retains blur.
- `generated/maka-regenerated.workflow.visual-check.json` was regenerated with `visual-check` so the receipt sha256 matches the rebuilt artifact (ok=true, containment + readability pass).
- **Rollback/failure**: pure presentation-layer change in the shared template; reverting the single commit restores prior rendering byte-for-byte. No API, schema, or contract change. Not applicable beyond revert.

## Validation evidence (final head `e277148`)

Base is `8c3af8a` (upstream `main`). All local runs below executed on the exact final-head tree; the last amend only added the two regenerated artifacts whose rebuild was verified before the amend.

1. `cd archify && npm test` — **1337 tests: 1290 pass, 0 fail, 47 skipped** (Node-version gates; skipped classification visible in the TAP output). Covers `check:viewer` freshness, golden byte-identity, `check:release-identity`, validators, brand marks, and the checked-in Checkout delta reproducibility test.
2. `ARCHIFY_CHROME=<chrome> node --test test/desktop-reader-browser.test.mjs` (real Chrome 152) — **2 tests: 2 pass, 0 fail, 0 skipped**.
3. `node bin/archify.mjs compare architecture ... --quality showcase --json` — 28/28 checks, completeness `complete`; receipt sha256 matches `examples/checkout-platform-delta.receipt.json`.
4. `scripts/build-zip.sh archify.zip` on Node 22 — reproduces committed archive bytes (release-package-gates inside `npm test`).
5. `node bin/archify.mjs visual-check generated/maka-regenerated.workflow.html --json` — ok=true, containment + readability pass on 1440x900/1600x1000/1920x1080/2048x1320, both themes; receipt and screenshots committed.
6. Real-world regression: 17-node architecture diagram re-delivered with the patched skill — `deliver` 9/9 checks; `visual-check` containment + readability pass at all four viewports in both themes.
7. **Upstream CI**: `.github/workflows/ci.yml` has a `pull_request` trigger, so the matrix (Node 18/20/22/24) and webm-artifact jobs run on this head — see the Checks tab of this PR for `e277148`.

Automated browser evidence: items 2, 5, 6 (receipts). **Perceptual review, reported separately**: the author visually compared panels in both themes at 1440x900/2048x1320 before and after — no readability regression observed; the maintainer's perceptual call is welcome.

## Trade-offs

- Panels lose their blur effect — they were mostly opaque already, so readability is unchanged; exports were never affected (static serialization strips viewer chrome state).
- Glow on hover is instant instead of eased. If the easing is desired, an `opacity`-only transition on a pre-rendered glow layer would be the compositable alternative; happy to iterate.


iqiiidi-sudo:
@coderabbitai run pre-merge checks

iqiiidi-sudo:
@coderabbitai run pre-merge checks

iqiiidi-sudo:
@coderabbitai run pre-merge checks

iqiiidi-sudo:
@tt-a1i Rebased onto current `main` (`0857afc`) and regenerated `archify.zip`, so the PR no longer conflicts. The full CI matrix is green on this head; the failing `webm-artifact` job also fails on `main` itself, so it is unrelated to this change. Local run: 1379 tests, 1331 pass, 0 fail, 48 skipped (Node-version gates).

If removing the blur entirely is too large a visual change, an opacity-only transition on a pre-rendered glow layer keeps the easing without repainting a filter per frame. Happy to iterate.

tt-a1i (changes requested):
The performance problem is worth addressing, and keeping the change in the authoritative Viewer source is the right maintenance boundary. Reviewed head `0146307bdaf3499376e38500e75901afe06f79f3` and its reported evidence; I have not independently measured its performance or accepted the visual trade-off.

Before integration, please provide one reproducible before/after measurement on the same diagram, browser/device, viewport, and pan/zoom or panel-opening sequence. A short performance trace or frame/long-task comparison is enough. The existing test and visual-check passes establish correctness/containment, but do not establish the stated cause or speedup. In particular, evaluate the permanent `will-change: transform` / `contain: layout style` addition separately from removing animated filters, so we can retain only the changes that earn their ongoing cost. A large benchmark framework is unnecessary.

Please refresh onto current `dev`, preserve the files removed by the already-merged cleanup #383, and regenerate the affected retained outputs and ZIP from the final source. The earlier claim that the browser CI failure is pre-existing is historical; the shared Motion regression has since been repaired on `dev`, so final-head CI must pass. Update the description's stale `e277148` evidence boundary or identify explicitly which unchanged evidence is reused.

Disposition: retain this PR for author revision and measured performance evidence. This is an unresolved benefit/scope claim, not a reproduced runtime defect. Prefer the smallest demonstrated improvement; wholesale removal of visual effects has not yet been accepted.

PR #387, opened 2026-09-11
feat(import): add Mermaid sequenceDiagram importer for typed sequence IR

### LINKED ISSUE #93: Import Mermaid sequenceDiagram sources as Archify sequence artifacts
## What to build

Extend the Mermaid import path with a documented `sequenceDiagram` subset. A user can import an existing Mermaid interaction trace and receive typed Archify Sequence IR that is validated and delivered through the normal Archify workflow.

The imported result must preserve interaction meaning and ordering rather than copy Mermaid layout or styling.

## Acceptance criteria

- [ ] Supported participant declarations, aliases, ordered messages, direction, labels, and request/return distinctions are preserved in typed Sequence IR.
- [ ] A supported source can pass the existing Sequence schema and quality gates and be delivered as a standalone Archify artifact.
- [ ] The supported subset is documented with runnable examples.
- [ ] Unsupported control structures or malformed syntax exit non-zero with a stable named diagnostic and useful source location; no message or participant is silently discarded.
- [ ] Existing JSON-authored Sequence behavior remains unchanged.
- [ ] Regression coverage includes valid, malformed, unsupported, and adversarial sequence fixtures.

## Blocked by

- #92

samiksha-shreya:
## Problem and value

Closes #93 (maintainer-labelled `enhancement`, `ready-for-agent`).

Current `main` has no path from a Mermaid `sequenceDiagram` into Archify's typed IR. Authors retype the topology by hand, and that is the step where a participant or message that was never in the source gets invented.

This PR adds `archify import sequence <input.mmd> [output.json] [--json]`. A self-contained parser in `archify/importers/sequence.mjs` emits sequence IR that passes `archify validate sequence`. The package still has no runtime dependency. Box measurement reuses `textUnits` from `renderers/shared/utils.mjs`, so it cannot drift from the renderer.

The mapping refuses to guess:

| Mermaid | Archify `variant` |
|---|---|
| `->`, `->>` (solid) | `default` |
| `-->`, `-->>` (dotted) | `return` |
| `-)`, `--)`, `-x`, `--x` (open or cross head) | `dashed` |

- `emphasis` and `security` are never emitted. They mean "main path" and "authorization step", and Mermaid expresses neither. Inferring them from label text would invent facts.
- Participant kinds use no name sniffing: `actor` becomes `external`, and `participant` becomes `backend`.
- Constructs the sequence schema cannot express (`loop`, `alt`/`else`, `opt`, `par`, `critical`, `break`, `rect`, `box`, `create`/`destroy`, `link`, `%%{init}%%`) fail with a diagnostic that names the construct, line, and column. Nothing is silently dropped.
- A note before the first message is rejected. Attaching it to a later message would invent placement.

## Stability impact

- Impact class: **Local behavior**. It adds one new CLI path and one new module. No schema, renderer, Viewer, or existing CLI path changes.
- Existing behavior is preserved. A regression test asserts that JSON-input commands are untouched. Every failure exits non-zero with `diagnostics[]` (stable `code`, line and column, `evidence`, `supportedFixes`), and nothing is written on failure. An importer exception becomes an `import/internal` diagnostic instead of a stack trace. An unknown format or option exits 2 through the existing `fail()`.
- Untrusted input: labels stay literal text. Ids are re-derived against `^[a-zA-Z][a-zA-Z0-9_-]*$` with `-2` suffixes, so distinct names cannot collapse. Control characters and unpaired surrogates are rejected. Tests assert that a `<script>` payload in a label is escaped in the delivered artifact, and that a JSON-injection payload adds no top-level IR key.
- No unrelated changes. The only edits to `bin/archify.mjs` are the usage line, the `IMPORT_FORMATS` registry, `commandImport`, and one `case 'import'` arm.

## Tests run

- Comparison base: `origin/main` @ `8c3af8a`. Candidate head: `aad8c79`.
- `npm test` from `archify/` on the base: **1337 tests, 1286 pass, 0 fail, 51 skipped**.
- `npm test` on the candidate: **1422 tests, 1371 pass, 0 fail, 51 skipped**. The delta is exactly the 85 new tests in `test/sequence-import.test.mjs`. The 51 skips are the `ARCHIFY_CHROME` browser tests, identical on base and candidate: skipped, not passed.
- The CLI seam is exercised. Every IR emitted from the 12 valid and adversarial fixtures runs through `archify validate sequence` and must pass.
- 41 fixtures: 8 valid, 14 unsupported-construct, 14 malformed, 5 adversarial.
- Remote CI has not run yet on this head.

## Visual evidence

Not applicable. This change adds an importer that emits typed JSON IR. It does not change any renderer, the Viewer, or generated artifact layout. Rendering of the emitted IR goes through the existing, unchanged sequence renderer, and each fixture's IR passes `archify validate sequence`.

## Generated artifacts

- `archify.zip`: rebuilt under Node 22 from the combined source, because `archify/importers/sequence.mjs` and `archify/references/mermaid-sequence-import.md` are new packaged files. It was verified by SHA-256 against the committed blob and contains the importer.
- No other generated output changed. The Viewer, templates, examples, gallery, and README showcase stay fresh because none of their inputs changed.

## Notes for review

- **Overlap with #140 (flowchart importer, open) and the Mermaid state importer PR (#388).** All three add `archify import`. This PR uses an `IMPORT_FORMATS` registry, so a later importer adds one row. The expected conflict is confined to three hunks in `bin/archify.mjs`. Whichever lands second should also merge the per-format reference docs into one canonical page, as CONTRIBUTING asks.
- This adds a new CLI command surface, `archify import <format>`. Happy to adjust the shape if you prefer a different one.


tt-a1i (changes requested):
The documented subset has value and follows the existing issue scope. Focused tests at `aad8c79e1c92a9eec9e40c70fca92e4f648a40d3` passed: 85 passed, zero skipped. I independently reproduced two CLI failures that block merging.

**P1 — importing onto the input destroys the Mermaid source.** With a valid `sequence` fixture, `archify import sequence input.mmd input.mmd --json` exits 0 with `ok: true` and replaces the source with JSON. The same happens when `output.json` is a symlink to `input.mmd`. These were actual separate CLI processes against this head, not a source-only inference.

**P2 — output errors break the machine-readable failure contract.** With an existing directory named `output.json`, the same command exits 1 with an uncaught `EISDIR` stack and empty stdout, so a caller requesting `--json` has no receipt.

Please integrate through one shared import-output path with #140: use the existing output-path resolver for extension/alias checks and an atomic candidate/rename commit with commit-time recheck and cleanup. Do not copy the direct `writeFileSync` implementation into a second command. Regression coverage should preserve source bytes for same-file/symlink/hardlink aliases, preserve a previous valid output on failed import/write, and assert structured diagnostics for a directory/unwritable target and commit-time failure.

Then update onto current `dev`, retain a single import format registry and canonical per-format references, and rerun import → validate → deliver on representative supported inputs. Because the importer chooses geometry, an unchanged renderer alone does not establish that the resulting diagrams are usable: inspect representative delivered artifacts and record the viewport/theme. Regenerate the package from the combined source and obtain green final-head CI. These are integration/evidence requirements in addition to the two reproduced defects; no full final-head browser acceptance is claimed here.

Author owns the bounded follow-up; keep this PR open targeting `dev` after the shared flowchart import path is ready. No main promotion.

PR #388, opened 2026-09-11
feat(import): add Mermaid stateDiagram importer for typed lifecycle IR

### LINKED ISSUE #94: Import Mermaid stateDiagram sources as Archify lifecycle artifacts
## What to build

Extend the Mermaid import path with a documented `stateDiagram` / `stateDiagram-v2` subset. A user can import an existing state model and receive typed Archify Lifecycle IR that is validated and delivered through the normal Archify workflow.

The imported result must preserve state and transition meaning rather than copy Mermaid layout or styling.

## Acceptance criteria

- [ ] Supported state declarations, aliases, initial and terminal states, directed transitions, and transition labels are preserved in typed Lifecycle IR.
- [ ] A supported source can pass the existing Lifecycle schema and quality gates and be delivered as a standalone Archify artifact.
- [ ] The supported subset is documented with runnable examples.
- [ ] Unsupported nested, concurrent, ambiguous, or malformed state syntax exits non-zero with a stable named diagnostic and useful source location; no state or transition is silently discarded.
- [ ] Existing JSON-authored Lifecycle behavior remains unchanged.
- [ ] Regression coverage includes valid, malformed, unsupported, and adversarial state fixtures.

## Blocked by

- #92

samiksha-shreya:
## Problem and value

Closes #94 (maintainer-labelled `enhancement`, `ready-for-agent`).

Current `main` has no path from a Mermaid `stateDiagram` into Archify's typed IR. Authors retype the topology by hand, and that is the step where a state or transition that was never in the source gets invented.

This PR adds `archify import state <input.mmd> [output.json] [--json] [--title text] [--outcome state=success|failure]`. A self-contained parser in `archify/importers/state.mjs` emits lifecycle IR that passes `archify validate lifecycle`. It has no runtime dependency. Geometry and text measurement reuse `renderers/shared/geometry.mjs` and `text-fit.mjs`, so the importer cannot drift from the renderer.

| Mermaid | Lifecycle kind |
|---|---|
| `state id <<choice>>` | `decision` |
| target of `[*] -->` | `start` |
| source of `--> [*]`, with `--outcome id=success\|failure` | `success` / `failure` |
| source of `--> [*]`, no override | `neutral` |
| every other state | `active` |

The importer refuses to guess:

- `[*]` is a pseudo-state. It sets the kind of the state it touches and does not create an unnamed box.
- A terminal state defaults to `neutral`. `--> [*]` says that the lifecycle ended, not how it ended. The importer does not read "Failed", "Done", or guard text. `success` and `failure` come only from the explicit `--outcome` flag. An override that names an unknown state fails with `import/state-unknown-outcome-target`.
- `waiting` and `external` are never emitted, because Mermaid has no construct with that meaning.
- Composite states are rejected, not flattened. Flattening would have to invent transitions between the inner region and the composite's siblings.

Also rejected with a diagnostic: `--` concurrency, `<<fork>>`/`<<join>>`, styling directives, floating notes, self-transitions (the renderer needs at least 32px between endpoints), and ids outside `^[a-zA-Z][a-zA-Z0-9_-]*$`.

## Stability impact

- Impact class: **Local behavior**. It adds one new CLI path and one new module. No schema, renderer, Viewer, or existing CLI path changes.
- A real limit, stated plainly: the lifecycle renderer uses fixed bands (5 phase, 3 event, 3 outcome), so **capacity is 11 states**. A larger diagram is rejected with a diagnostic instead of being shipped broken. The importer computes the renderer's exact geometry and scores candidate orthogonal routes. An unplaceable route or guard label becomes an import diagnostic. Imported IR declares `quality_profile: standard`, and `showcase` may need manual routing.
- Failure behavior: every failure exits non-zero with `diagnostics[]` (`code`, `severity`, `subject.line`, `evidence`, `supportedFixes`), and nothing is written. An unknown format or option exits 2.
- Untrusted input: quotes, `<script>`, `</text>`, entity codes, and a `-->` inside a quoted description are inert in the IR and escaped in the artifact. The artifact SVG is re-parsed with `saxes` to prove it. C0 controls, `U+007F`, `U+2028`, `U+2029`, and `U+FEFF` are rejected by name, not silently stripped.
- No unrelated changes. The only edits to `bin/archify.mjs` are the usage line, the `IMPORT_FORMATS` registry, `commandImport`, and one `case 'import'` arm.

## Tests run

- Comparison base: `origin/main` @ `8c3af8a`. Candidate head: `b10bcca`.
- `npm test` from `archify/` on the base: **1337 tests, 1286 pass, 0 fail, 51 skipped**.
- `npm test` on the candidate: **1364 tests, 1313 pass, 0 fail, 51 skipped**. The delta is exactly the 27 new tests in `test/state-import.test.mjs`. The 51 skips are the `ARCHIFY_CHROME` browser tests, identical on base and candidate: skipped, not passed.
- The CLI seam is exercised. Every emitted IR runs through `archify validate lifecycle`. Valid fixtures also assert `composition.summary.warnings === 0`, so a layout regression that only avoids hard errors still fails.
- 15 fixtures: 5 valid, 4 malformed, 6 unsupported or adversarial. There is also one runnable documented example, `archify/examples/release-train.state.mmd`.
- Remote CI has not run yet on this head.

## Visual evidence

Not applicable. This change adds an importer that emits typed JSON IR. It does not change any renderer, the Viewer, or generated artifact layout. Rendering of the emitted IR goes through the existing, unchanged lifecycle renderer, and each valid fixture's IR validates with zero warnings.

## Generated artifacts

- `archify.zip`: rebuilt under Node 22 from the combined source, because `archify/importers/state.mjs` and `archify/importers/README.md` are new packaged files. It was verified by SHA-256 against the committed blob and contains the importer.
- No other generated output changed. The Viewer, templates, examples, gallery, and README showcase stay fresh because none of their inputs changed.

## Notes for review

- **Overlap with #140 (flowchart importer, open) and the Mermaid sequence importer PR (#387).** All three add `archify import`. This PR uses an `IMPORT_FORMATS` registry, so a later importer adds one row. The expected conflict is confined to three hunks in `bin/archify.mjs`. Whichever lands second should also merge the per-format reference docs into one canonical page, as CONTRIBUTING asks.
- `--outcome` goes beyond the minimum in the issue. Without it, the `success` and `failure` legend kinds are reachable only by guessing from state names.
- This adds a new CLI command surface, `archify import <format>`. Happy to adjust the shape if you prefer a different one.


tt-a1i (changes requested):
The documented subset has value and follows the existing issue scope. Focused tests at `b10bccabfb60088342e18b683ea00be045703266` passed: 27 passed, zero skipped. I independently reproduced two CLI failures that block merging.

**P1 — importing onto the input destroys the Mermaid source.** With a valid `state` fixture, `archify import state input.mmd input.mmd --json` exits 0 with `ok: true` and replaces the source with JSON. The same happens when `output.json` is a symlink to `input.mmd`. These were actual separate CLI processes against this head, not a source-only inference.

**P2 — output errors break the machine-readable failure contract.** With an existing directory named `output.json`, the same command exits 1 with an uncaught `EISDIR` stack and empty stdout, so a caller requesting `--json` has no receipt.

Please integrate through one shared import-output path with #140: use the existing output-path resolver for extension/alias checks and an atomic candidate/rename commit with commit-time recheck and cleanup. Do not copy the direct `writeFileSync` implementation into a second command. Regression coverage should preserve source bytes for same-file/symlink/hardlink aliases, preserve a previous valid output on failed import/write, and assert structured diagnostics for a directory/unwritable target and commit-time failure.

Then update onto current `dev`, retain a single import format registry and canonical per-format references, and rerun import → validate → deliver on representative supported inputs. Because the importer chooses geometry, an unchanged renderer alone does not establish that the resulting diagrams are usable: inspect representative delivered artifacts and record the viewport/theme. Regenerate the package from the combined source and obtain green final-head CI. These are integration/evidence requirements in addition to the two reproduced defects; no full final-head browser acceptance is claimed here.

Author owns the bounded follow-up; keep this PR open targeting `dev` after the shared flowchart import path is ready. No main promotion.

PR #253, opened 2026-09-01
fix(lifecycle): hide decorative rail during semantic interactions

sacloudy:
## Problem and value

Closes #244.

Lifecycle diagrams render a decorative primary rail behind the authored states. During Intent Trace, direct-relationship preview, and committed Semantic Passport focus, semantic nodes and edges dim but that unmarked decoration stays fully opaque. It can then read as a false direct edge across unrelated states.

This change keeps the rail in the ordinary READ view while removing that false semantic signal from all three public interaction states.

## Scope

- What changed:
  - Mark only the decorative lifecycle path with `data-lifecycle-rail`.
  - Hide that decoration while `data-intent-trace-active`, `data-relationship-preview-active`, or `data-focus-active` is present.
  - Add final-HTML contract coverage plus a real-Chrome regression that exercises node hover, relationship hover, and committed focus.
  - Refresh all generated HTML, Gallery metadata, README motion proof, Architecture Delta proof, and the packaged skill.
- What deliberately did not change:
  - No JSON schema, validator, layout, authored topology, route, reachability, relationship identity, or default READ presentation changed.
  - The rail does not receive `data-edge-from`, `data-edge-to`, or any semantic relationship metadata.
- No unrelated changes: confirmed.

## Stability impact

- Compatibility and migration risk: low. Existing typed JSON and public viewer APIs are unchanged; this is one renderer marker and three interaction-scoped CSS rules.
- Renderer, validator, package, or generated-artifact risk: lifecycle output gains one inert data attribute. The validator remains unchanged. Checked-in artifacts and `archify.zip` were regenerated and verified.
- Failure behavior and rollback path: unsupported or non-lifecycle diagrams do not emit the marker, so the selectors are no-ops. Reverting this single commit restores the previous presentation.

## Tests run

```shell
ARCHIFY_CHROME='/Applications/Google Chrome.app/Contents/MacOS/Google Chrome' \
  npx -y -p node@22 -c 'node --test \
    archify/test/architecture-delta.test.mjs \
    archify/test/cli.test.mjs \
    archify/test/generated-artifact-xml.test.mjs \
    archify/test/intent-trace.test.mjs \
    archify/test/lifecycle-rail-browser.test.mjs \
    archify/test/readme-showcase.test.mjs \
    archify/test/relationship-direct-explorer.test.mjs \
    archify/test/semantic-passport.test.mjs'
```

[...]

PR #397, opened 2026-09-12
fix(checker): preserve explicitly authored straight route intent

tt-a1i:
## Problem and value

Schema-valid `route: "straight"` relationships in architecture, data-flow, and lifecycle diagrams can render successfully but fail `check`, `validate`, and atomic `deliver` solely because the final artifact checker treats every diagonal as unintended. On base `6db72a9`, the lifecycle reproduction renders `M 307 157 L 493 307`, then fails `orthogonal_arrows`.

Preserve explicit straight intent in a small renderer-generated SVG attribute. The checker accepts that attribute only on a semantic relationship with one actual direct segment. This recovers the supported authoring behavior without altering paths or broadly accepting unmarked diagonals. This is a recovered fuzz finding; the minimal redacted reproductions are in `archify/test/authored-straight-routes.test.mjs`.

## Stability impact

- Contract repair across three renderers and the shared artifact checker. Automatic routes, workflow routing, explicit endpoint-side validation, opaque-node clearance, and other quality gates retain their existing behavior.
- Previously rendered unmarked HTML keeps the strict diagonal check; rerender the supported JSON to carry its authored intent into the artifact.
- Nonempty `via` overrides `straight` and cannot acquire the direct-route exception. Empty `via` remains compatible for architecture/lifecycle. Data-flow retains its existing orthogonal-via validation.
- Only actual two-point authored diagonals receive metadata; axis-aligned straight relationships and existing example bytes stay unchanged. The checker rejects stale markers on bent/curved paths and markers without semantic edge identity.
- The lifecycle design guidance now distinguishes the axis-aligned recommendation from the supported explicit straight option.
- No schema, dependencies, Viewer behavior, finite-SVG validation, default router, version, or release identity changes. This does not incorporate the separate #373 finite-SVG or #379 inferred-side changes.

## Tests run

Base `6db72a9aea3d0f67a6a034e41f8a5491476a11c1`; candidate `4f8e59416a4488b9dd7e95d642ad1486fd938619`. Node 22.23.2 on macOS:

- New end-to-end regression file: **14 passed**. Covers all three modes under standard/showcase through render → check → validate → deliver, exact path/point preservation, marker removal, auto routing, explicit-side rejection, via precedence, and invalid-marker controls.
- Existing `render-output-checks`, `layout-rules`, `v1-compatibility`, and `gallery` suites: **138 passed**.
- Base reproduction: all three minimal inputs render successfully and fail only `orthogonal_arrows`; the candidate retains the same endpoint coordinates and SVG `d`.
- Golden checks passed, including both development and packaged examples. Gallery generation reproduced its 11 checked-in artifacts and 99 checks.
- Extracted ZIP outside the repo, without `node_modules` or installation: the same **14 end-to-end checks passed**.
- Full `npm test`: **1359 total; 1311 passed, 1 failed, 47 skipped**. The sole failure is the pre-existing checked-in MCO byte-reproducibility drift (its old HTML still uses network fonts). The same test fails on clean base `6db72a9`; base and candidate regenerate the pinned MCO artifact to identical bytes (SHA-256 `3c53dea72ae62c91cb367a0e835f0ebd4aa1e4aa8719747278593aea60177180`, 823676 bytes). The unrelated MCO artifacts are unchanged. Chrome-gated/optional tests remain explicitly skipped; this is not a claim of a green complete local suite.
- Independent review found an empty-via gap, which was fixed; both new empty-via delivery cases passed independent replay. No remaining in-scope review blocker.

## Visual evidence

Not applicable: the SVG geometry, styling, and Viewer code are unchanged. Fixed inputs preserve the exact paths/points, and existing example/Gallery output remains byte-identical. The visible difference is successful delivery of explicitly authored straight routes. Browser/perceptual acceptance is not claimed.

## Generated artifacts

`archify.zip` regenerated with official Node 22.23.2 and reproduced byte-for-byte by a second build. Its 79-entry inventory is unchanged; only the seven affected renderer/checker/documentation entries differ from base. Examples, Gallery, guide/start pages, README showcase, and Viewer outputs need no regeneration because their authoritative generated output remains unchanged.

Final-head hosted CI passed on `4f8e594` across Node 18/20/22/24, browser/WebM, ZIP freshness, published-manifest and all three package-smoke platforms: [CI run](https://github.com/tt-a1i/archify/actions/runs/34677407961). This does not reclassify the documented pre-existing local MCO failure.


sunsunsun-java (approved):

tt-a1i:
The reviewed commits are now preserved in integration PR #449 targeting `dev`. It combines #397, #166 and #253 on the current base, ports Viewer changes to authoritative source, corrects the unintended workflow heading recoloring, and rebuilds the shipped artifacts/ZIP. Full local Node 22 + Chrome validation passed (1,599 passed, 0 failed, 5 explicit gated skips); the final CI-only/test-guard follow-up also passed its focused browser/control checks. Full final-head CI is running on #449. This PR remains open until that integration is accepted; merging #449 with ancestry preserved should mark the original PR merged automatically. `main` remains unchanged.

PR #398, opened 2026-09-12
feat(viewer): authored node explanations in the semantic passport

cilladev:
## Problem and value

The Semantic Passport  for each node shows facts about a node but has nowhere for the author to say what it actually does.  (a short text explanation for maybe non technical people and let the diagram explain itself). This adds an optional explanation text field to nodes in all five diagram types. When a reader clicks a node that has one, an Explain button appears next to Copy link. Clicking it flips the passport to show the explanation; clicking again flips back. The text is stored outside the SVG so exports are unaffected. Diagrams that don't use the field work exactly as before.

## Stability impact

- Impact class: new optional schema field + shared viewer behavior. Changed: `utils.mjs`, `cli.mjs`, all five renderers, `i18n.mjs`, `focus.js`, `template.source.html`, five schemas + regenerated validators.
- Existing behavior preserved: documents without `explanation` render identically. The Explain button is `hidden` and the data block absent when unused. Exports, focus, reach, route probe and pins are untouched.
- No unrelated changes: confirmed. Other diffs are regenerated artifacts.

## Tests run

Base `6db72a9`, head `5aeb586`. Node v22.17.0, Chrome 152.

- `npm test`: 1302 pass, 0 fail.
- Full suite with `ARCHIFY_CHROME`: 1435 pass, 0 fail.
- New `test/passport-explanation.test.mjs`: validates all five types, rejects bad input, prose never in SVG, no block when unused, zh-CN localised.
- New `test/passport-explanation-browser.test.mjs`: flip, focus, position stability, export isolation, Escape.

## Visual evidence

Click Explain:
image

Passport flips:
image

Click  Enter/Space/Esc: Flips back to passport
image

## Generated artifacts

Validators, viewer template, all examples, compare artifact, gallery, README proof and `archify.zip` regenerated. All freshness checks green.


tt-a1i (changes requested):
Authored plain-text explanations are useful for readers who need more context than a node label. Keeping the bounded prose outside SVG and rendering it through textContent is a reasonable implementation. Reviewed current head `5c0f0f9c679817427047b3f4f5f7a080ec6d97bc`; I found a long-content keyboard/viewport issue that the short desktop fixture does not cover.

**Blocking: the explanation opens with its focused Back control outside both the visible face and the viewport.** Using the existing web-app input with a valid 1,190-character, paragraph-separated explanation on `users`, at 320×300 I scrolled Explain into view and physically clicked it. The panel grew from about 214px to 464px high despite the 300px viewport. The explanation face had clientHeight 462, scrollHeight 899 and scrollTop 0; the focused `focus-explanation-back` was around y=936–959, below the face's bottom around y=533 and below the viewport. `openExplanation()` caps growth against the diagram container rather than the available viewport and focuses the bottom action with preventScroll, so a keyboard user receives no visible focus/control after opening the prose.

Keep the explanation bounded by usable viewport space, make all prose scrollable, and keep Back visible/reachable with a visible keyboard focus. A fixed header/footer and scrollable body is one possible fix, not a required redesign. Add a real pointer/keyboard regression with near-limit prose at reflow/short-height size, including closing and restoring focus. Please also check resizing while the explanation is already open; the current inline minHeight persists until close.

Local verification: all 6 explanation tests passed with Node 22 and real Chrome at this head. The separate long-content browser probe above fails the missing scenario. The supplied browser test uses a short two-paragraph desktop case, programmatic click and synthetic Escape; its export assertion inspects SVG markup, not an actual exported PNG. Those checks remain useful, but should not be described as full physical keyboard/export acceptance.

The PR already edits canonical Viewer sources; retain that approach when refreshing onto current dev. Preserve the newer Passport placement work, add the explanation browser suite to the shared browser gate, and regenerate affected outputs/ZIP after the fix. Update the description's older `5aeb586` full-suite evidence with the current-head results or clearly label the reused revision. No change to the optional plain-text scope or schema length limit is requested.

PR #399, opened 2026-09-12
feat(erd): add an entity-relationship diagram type (#115)

TuGou-a:
## Problem and value

Current `main` has no type for the shape of the data itself. A `database`-typed architecture node names a store, but not a key, a foreign-key target, or whether an end is one or many, so "map this service's data model" can only be approximated.

This implements #115: `erd` becomes a sixth typed diagram — entities, key attributes, and both-end cardinality — authored as JSON, validated fail-closed, delivered as a self-contained HTML artifact that reuses the existing viewer with no new viewer concepts (it rides the `data-node-id`/`data-edge-*` hooks and a `data-node-kind` the viewer already knows).

It is a second implementation of the same issue as #183, with different decisions, so a maintainer can take either. Differences, and the reasoning:

| | #183 | this PR |
| --- | --- | --- |
| cardinality | `cardinality: { from, to }` free strings | `fromCardinality` / `toCardinality` (`one`/`many`) + `fromOptional` / `toOptional` |
| entity kind | required `kind` enum, 5 values | none in this slice |
| attribute keys | `role: primary/foreign/attribute` | `key: pk/fk/uk` |
| attribute count | capped in-schema | no cap; per-row fit + diagnostic, full table in cards |
| orphan entity | error, with a `standalone` escape | no orphan error |
| repository evidence | `sources` + `repository` in-slice | deferred in this slice; `#389` already generalizes evidence to every type on `dev`, so the deferral is a scope choice here rather than a missing contract |
| routing | per-type routing; the shared-geometry change was reverted | shared router extracted (`#266` landed the same extraction as `architecture/routing.mjs` on `dev`); obstacle-aware routing |

The cardinality point is the substantive one: `"0..N"` is display notation, so no gate can check it and the crow's-foot marker can only be parsed from arbitrary text. `one|many` plus an optionality boolean is checkable, and the marker follows from the fact rather than from string parsing. The `kind` vocabulary is the second: `junction` and `external` are derivable from the key facts (two foreign keys forming the identity; no inbound foreign key), so an authored enum can contradict the keys the same document declares.

## Stability impact

- Impact class: contract change (new type keyword, schema, acceptance rules, Skill instructions) plus shared behaviour (router extraction).
- Changed behaviour and shared callers: `renderers/shared/orthogonal-router.mjs` is new; `renderers/architecture` now imports it instead of owning the candidate families, `strictClearance`/`preferredCandidates`/`extraCandidates` are opt-in and unused by architecture, and `renderDefinitions(extra = '')` appends type-owned definitions inside the same `<defs>`. This extraction was written before `#266` landed the same split on `dev`, so it is a second one; the hooks ERD needs (`preferredCandidates`, the spread `ports` map) are the only additions over the landed module, and they belong folded into it rather than kept parallel.
- Existing behavior preserved: architecture's `web-app` example renders **byte-identical** to the checked-in golden, and `check:viewer`, `check:brand-marks`, `check:validators`, `check:release-identity` and the golden suite pass.
- Intended compatibility changes: `guide` gains a twelfth recipe and routes data-model questions to `erd`; the gallery gains a twelfth case; the packaged Skill gains a type.
- Failure behavior: invalid ERD exits non-zero with machine-readable diagnostics (`layout/constraint` carrying `subject`, `evidence`, `supportedFixes`), and `deliver` preserves the previous artifact. A relationship that cannot clear an unrelated entity fails with `clean-flow/edge-through-node` rather than being drawn through a box.
- No unrelated changes: the ZIP and the site surfaces are regenerated only where the new type and recipe reach them. `docs/assets/archify-live-proof.{gif,json}` were deliberately left untouched — regenerating them on unmodified `main` already produces a diff, so that staleness belongs to its own change.

## Review round 2 — fail-closed contract

Both reproductions from the review are fixed at head `91f781c`, with the schema as the gate:

```
# every relationship loses fromCardinality/toCardinality
node bin/archify.mjs validate erd /tmp/no-cardinality.erd.json --json
# ok:false · stage "render" · erd schema validation failed:
#   /relationships/0 (id/label: "order_customer") must have required property 'fromCardinality'

# every key:"fk" attribute loses its references target
node bin/archify.mjs validate erd /tmp/no-references.erd.json --json
# ok:false · erd schema validation failed:
#   /entities/1/attributes/1 (id/label: "order") must have required property 'references'
```

`fromCardinality`/`toCardinality` are now required on a relationship, `references` is required whenever `key: "fk"` is asserted, and a `references` value must read `"entity.attribute"`. The renderer keeps the checks the schema cannot make — that the named entity and attribute exist — and its cardinality fallback is now documented as unreachable after validation instead of as a default that made an omission validate. Regression coverage: two `erd-rendering` tests strip the fields across the whole orders example the way the review reproduced it, and three schema-enforcement cases join `golden.mjs`.

The two stale-claim notes are accepted: `#266` landed the router extraction on `dev` as `architecture/routing.mjs`, and `#389` generalized repository evidence to every type. The table above and the stability section now say so.

## Consolidation

Taking the review's direction: `#183` is the integration base, and this branch stays open as the source/alternative rather than advancing a parallel schema. What is ready to contribute, with attribution:

- **Typed, fail-closed cardinality** — `one`/`many` plus an optionality boolean, both ends required, so nothing about an end is inferred from the other end or from a display string.
- **Obstacle-aware route coverage** — entity boxes are opaque here, so a table between two aligned anchors is routed around rather than crossed, and an unroutable relationship is an error; the fixtures and the clean-flow assertion come with it.
- **The bundling geometry** (fan-in/fan-out trunks) if it is wanted; it is independent of the schema vocabulary.

The branch is currently conflicting against `dev` (10 files, mostly regenerated artifacts plus the shared-router extraction). I have not rebased it, because re-homing the router is the consolidated contract's decision rather than this branch's; I can rebase it onto `dev` if you want it kept as a clean alternative. No further schema work will go into it meanwhile.

## Tests run

Comparison base: the PR base branch `dev`. Candidate: this branch at `91f781c`.

```
cd archify && npm ci && npm test          # Node 22, the canonical ZIP major
# tests 1357 · pass 1309 · fail 0 · skipped 48

ARCHIFY_CHROME="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  node --test test/desktop-reader-browser.test.mjs
# 2/2 pass, real Chrome, 1440x900 adaptive reader

node bin/archify.mjs deliver erd examples/orders.erd.json /tmp/orders.erd.html --quality showcase --json
# ok, 9/9 checks, showcase pass, 0 errors / 0 warnings
# spec sha256 8e725322cd16025a… 5862 B · artifact sha256 39e49f1afefe5bc2… 810800 B

node bin/archify.mjs visual-check /tmp/orders.erd.html --json
# pass · containment pass · readability pass · viewerChrome pass · 0 diagnostics
# 1440x900 → 900, 1600x1000 → 1000, 1920x1080 → 1080, 2048x1320 → 1320
```

New regression coverage in `archify/test/erd-rendering.test.mjs` (12 tests). Besides the artifact receipt, it asserts the geometry contract this type adds: every route is checked against unrelated entity rectangles, an entity placed between two aligned anchors is routed around instead of crossed, and an unroutable relationship fails with a diagnostic while writing no artifact. The schema negatives (unknown entity, unknown attribute reference, duplicate attribute name, under-wide entity) and the cardinality/optional/dashed variants are covered too.

The 48 skipped tests are the platform- and browser-gated ones; the browser test above was run explicitly rather than left skipped.

## Visual evidence

- Evidence provided: the checked-in gallery artifact `docs/gallery/artifacts/orders.erd.html` (linked from the gallery card), and the `visual-check` sidecar set (4 viewports × 2 themes) reproducible with the command above. Drawings inspected: crow's-foot markers at both ends, the dashed non-identifying payment link, key glyphs, and the legend.
- Comparison conditions: same authored input, `--quality showcase`, light and dark, static preset, no motion.
- Automated or browser checks: `visual-check` pass with 0 diagnostics at 1440×900 / 1600×1000 / 1920×1080 / 2048×1320; `desktop-reader-browser.test.mjs` 2/2 with real Chrome.
- Perceptual visual review: pending a human. The parts I would ask to be looked at are the crow's-foot markers against the existing arrow language and the row typography inside entity boxes.

## Generated artifacts

- `archify.zip` rebuilt with Node 22 and smoke tested from the extracted package (`doctor` now reports the erd renderer, schema, and example).
- `archify/renderers/shared/generated-validators.mjs` regenerated; two consecutive runs are byte-identical.
- `archify/examples/erd-orders-rendered.html` and `examples/erd-orders-rendered.html` re-rendered from the example source.
- `docs/gallery.html`, `docs/gallery/manifest.json`, `docs/gallery/artifacts/orders.erd.html`, `docs/gallery/sources/orders.erd.json`, and `docs/guide.html` regenerated from the shared recipe source.

---

---

On the consolidation: this schema's vocabulary is not the thing to defend — if the consolidated contract takes `cardinality: { from, to }` in whatever shape it settles on, the markers, the obstacle routing, and the fail-closed requirements above port to it directly. What I would argue to keep either way is that both ends are stated rather than defaulted, and that an `fk` names its target: those are the two places where a diagram otherwise asserts a fact nobody authored.


TuGou-a:
@coderabbitai review

tt-a1i (changes requested):
Compared `fecaa6091b1e620e1263d2d1226c5bff53a5787c` with #183 and issue #115. All 10 focused ERD tests pass. Typed cardinality, optionality and obstacle-aware route tests are valuable; maintaining a second competing ERD schema is not.

Two public-CLI reproductions still fail the issue's stated fail-closed contract:

- Removing `fromCardinality` and `toCardinality` from every orders-example relationship still makes `validate erd --json` return exit 0 / `ok:true`. `cardinalityOf()` silently supplies many-to-one, mandatory-at-both-ends facts that the author never declared. Require both endpoint cardinalities instead of guessing them.
- Removing every `references` field from `key: "fk"` attributes also validates successfully. The reference loop skips absent strings, so a claimed foreign key need not name a target. Require and validate a target whenever FK is asserted.

There are also deliberate scope differences from #115: repository evidence is omitted, standalone/orphan policy is removed, and a grid-placement surface is added. The current PR body acknowledges some differences; their presence is not acceptance of a changed first-slice contract. Current dev already has the router extraction (#266) and generalized repository evidence (#389), so the claimed need for a second extraction/evidence deferral is stale.

Consolidation direction: prefer #183 as the issue-aligned integration base, taking this PR's typed marker and obstacle-routing coverage where useful with attribution. #183's own arbitrary-cardinality acceptance is being requested for repair; this is not approval of its current head. Owner/action: maintainers and the #183 author consolidate one public ERD contract; this author can contribute the reusable portions rather than advancing a parallel schema. Keep this PR open as the source/alternative until the consolidated replacement is actually accepted; do not close it as already superseded. No final browser or full-suite acceptance is claimed from these focused checks.

TuGou-a:
Both reproductions are fixed at `91f781c`, and the two stale-claim notes are accepted.

**Fail-closed contract.** The schema is now the gate, so an omission cannot validate at all:

```
# every relationship loses fromCardinality/toCardinality
$ node bin/archify.mjs validate erd no-cardinality.erd.json --json
ok:false · stage "render" · erd schema validation failed:
  /relationships/0 (id/label: "order_customer") must have required property 'fromCardinality'

# every key:"fk" attribute loses its references target
$ node bin/archify.mjs validate erd no-references.erd.json --json
ok:false · erd schema validation failed:
  /entities/1/attributes/1 (id/label: "order") must have required property 'references'
```

`fromCardinality`/`toCardinality` are required on a relationship, `references` is required whenever `key: "fk"` is asserted, and a `references` value must read `"entity.attribute"`. The renderer keeps the checks the schema cannot make (that the named entity and attribute exist) and its cardinality fallback is documented as unreachable after validation rather than as a default — `cardinalityOf()` no longer decides anything about a diagram that passes. Regression coverage mirrors the reproductions: two tests strip the fields across the whole orders example, and three schema-enforcement cases join `golden.mjs`. Full suite is 1309 passing on Node 22; validators and the archive are rebuilt.

**Scope claims.** `#266` landed the router extraction as `architecture/routing.mjs` and `#389` generalized repository evidence, both after this branch was cut. The PR body and the stability section now say so, and the deferral is described as a scope choice rather than a missing contract.

**Consolidation.** Agreed on the direction — `#183` as the integration base, this branch kept open as the source/alternative, no further schema work here. The portions I would offer for it, with attribution:

- typed, fail-closed cardinality (`one`/`many` + optionality boolean, both ends required and neither inferred from the other);
- the obstacle-aware route coverage, because entity boxes are opaque is not a policy here — a table between two aligned anchors is routed around, and an unroutable relationship is an error rather than a drawn crossing;
- the fan-in/fan-out trunk geometry, if useful; it does not depend on the schema vocabulary.

The branch is conflicting against `dev` in 10 files (regenerated artifacts plus the router). I have left it unrebased because re-homing the router is the consolidated contract's decision, not this branch's — tell me if you want it rebased onto `dev` to keep it usable as the alternative, and I will.

One note on CI: GitHub cannot build the `pull_request` merge ref while the branch conflicts against `dev`, so no workflow has run on `91f781c` (0 runs for that SHA). The fix is verified locally — `cd archify && npm ci && npm test` on Node 22 is 1357 tests / 1309 passing / 0 failing, including the two reproductions above. Rebasing onto `dev` would restore CI; that is the offer in the paragraph above, and I will do it as soon as you want the branch kept as a working alternative.

tt-a1i:
Thanks @TuGou-a — we support adding ER diagrams and would welcome you continuing to improve this implementation.

First, a coordination update: #183 is now closed without merging, so our earlier instruction to wait for it as the integration base is no longer useful. Sorry for leaving you waiting. If you are still interested, please proceed with updating #399 against current dev as the consolidation path, reusing the router/evidence foundations already there and carrying over useful work from #183 with attribution. We should converge on one ERD contract rather than maintain parallel implementations.

I ran your unchanged orders example at `91f781cede28aaea6935b65b31632571954ecf1f` with Node 22.23.2. The public showcase delivery passed 9/9 checks, and all 12 ERD rendering tests passed. I also opened the generated HTML in a browser, inspected both themes, and checked node search/focus and connected-table details. This establishes a working example, not full integration or merge acceptance on current dev.

The capability is useful, but the example's presentation still needs refinement. My main reading concerns are:

- The tables have nearly identical visual weight, making it difficult to identify the order model's center at a glance. Please improve hierarchy and grouping through layout and typography; extra colors or new style modes are not required.
- The relationships around order/payment/order_item are harder to follow, and the cardinality symbols and legend are faint, especially in the light theme. Make both-end meanings easier to distinguish without relying on color alone.
- There is substantial space above the tables while field text is relatively small. Aim for a better balance between readable fields, routing space, and overall compactness.
- The example places `order.total_amount`, `payment.paid_at`, and `order_item.quantity` only in prose cards. Please include these in the structured attributes and table display for this example, keeping cards for supplementary explanations. If an example intentionally presents a subset, label that scope clearly rather than implying a complete schema.

Please start by improving this example, then try a representative real or redacted data model with more fields and relationships. Before/after artifacts at the same viewport and theme would help us assess the improvement. Preserve the existing five diagram types and include relevant regression checks plus practical generation/repair timings if the changes affect them; passing validation alone does not establish readability or unchanged performance.

The direction has our support. Let's make the existing slice clear and comfortable to read before expanding its scope. Thanks again for the implementation and for coordinating the earlier review fixes.

TuGou-a:
Updated to current `dev` as the consolidation path, with the reading round from the 2026-09-21 review. The PR description above now describes this head; the previous body's consolidation section (waiting on #183, the parallel router, the evidence deferral) is superseded — #183 is closed, the router is `dev`'s single module, and `erd` verifies repository evidence like the other five types.

Head: `382923c`, rebased onto `dev` (force-pushed, hence the rewrite notice).

**Consolidation**
- One router: architecture imports `shared/orthogonal-router.mjs` and renders **byte-identical** to the checked-in golden (`cmp` on a fresh render).
- Evidence: `meta.repository` + `entities[].sources`, wired through the shared collection map and `--repo-root`, fail-closed on the same six failure modes as the other types; `erd` joins both evidence suites.
- `archify.zip` rebuilt with Node 22 from the branch index; a second build is `cmp`-identical and package smoke passes from the extracted archive.

**The four reading concerns**
1. Grouping/hierarchy: a contiguous `tag` block is drawn as a labelled domain band; rows read as text (18-unit rows, 10-unit fields, 28-unit header); `order` is the wider hub inside the `checkout` band.
2. Relationships and symbols: same-side ports keep 24 units apart (a glyph is 14 tall); the crow's foot, bar and optional ring are drawn in the muted text ink — measured 3.98:1 light / 6.02:1 dark against a table fill, where the shared arrow token sits at 2.14:1 — and the legend repeats the exact glyph and weight.
3. Compactness: the empty band under the tables was the reserved legend strip, and the legend had been **silently dropped** (the strip was four units short and the type's default hides an unfittable legend). The strip now fits all six entries, which fills the space with what it was reserved for.
4. Complete fields: `order.total_amount`, `payment.paid_at` and `order_item.quantity` are in the tables; cards are supplementary.

**Site and start surfaces** (found by a review pass over this round, not by CI): `website/src` now carries the twelfth gallery case and thirteenth recipe; the gallery card's "Create this type" link resolves to a real `database-schema` start on both surfaces, with the sixth chooser button and count. A guard asserts that every type a gallery card links to is in the start allow-list, has a start entry with bilingual prompts, and has a chooser button whose count matches the label.

**Evidence**
```
cd archify && npm test
# tests 2178 · pass 2109 · fail 0 · skipped 69
node bin/archify.mjs deliver erd examples/orders.erd.json /tmp/orders.erd.html --quality showcase --json
# ok · 9/9 checks · showcase pass · 0 errors / 0 warnings
ARCHIFY_CHROME="…" node bin/archify.mjs visual-check /tmp/orders.erd.html --json
# pass · containment pass · readability pass · 4 viewports × 2 themes · hasLegend
cd website && npm run check && npm run build && npm test    # 7/7
bash scripts/build-zip.sh && cmp archify.zip /tmp/again.zip # deterministic
```
Timings (Node 22, median of five, same input): `deliver` 146 ms → 190 ms; the renderer alone 79–84 ms → 104–111 ms. Most of that is the larger document this revision writes (bands plus complete field lists) — the profile delta lands in serialization, hashing and GC, not in a new code path. For scale: `dev`'s own schemas cost every type about 50 ms more than this branch's fork point on this machine (architecture 47–53 ms → 97–113 ms, byte-identically), so the honest baseline for these numbers is `dev`'s.

Two points I would like your call on, both in the description: the crow's foot's toe orientation (currently toes spread away from the table, as the renderer README documents; some notations draw them against the entity) and whether an unfittable default legend should hide — today it does, a declared `meta.legend` fails loudly instead.

tt-a1i:
Thanks for the careful update. I opened the current `c214e05` orders artifact at 1440×900 in both themes. The domain bands, complete field rows, and restored legend are clear improvements over the earlier example.

On the legend question, we don't require the entire ERD to fit on one screen. A larger diagram that needs normal scrolling is fine; preserving readable fields and a visible legend matters more than squeezing everything into the initial viewport. So I would not change the default to fail just because the diagram is taller than the screen. For the ERD renderer, please keep the default legend in the generated content when sizing the diagram, rather than silently hiding it when its reserved strip is short. If an explicitly fixed drawing area genuinely cannot hold it, an actionable capacity diagnostic is reasonable. The current orders example already shows the legend, so this is about the remaining edge case.

The `scrollHeight == innerHeight` result in the PR is useful evidence for this particular orders fixture, not an acceptance rule that every ERD must fit without scrolling. Please do not shrink text, omit content, or constrain the layout simply to satisfy that condition. We are revisiting the broader one-screen assumptions in Archify as part of the canvas work.

For future large-diagram navigation, you can also look at our experimental [labs/infinite-canvas](https://github.com/tt-a1i/archify/tree/labs/infinite-canvas) branch and the separate [adaptive-canvas work in #478](https://github.com/tt-a1i/archify/pull/478). They are not on `dev` yet, and neither should be a dependency of #399. Viewer panning/zooming can make existing content easier to explore, but cannot recover a legend omitted while generating the SVG. We expect to keep improving large-diagram browsing, including considering how ERD fits into it later.

On the crow's-foot question, I'd prefer the foot to open toward the entity whose cardinality it describes; the current glyph can read like a directional arrow. Please show the same relationship before and after the flip, including the optional-many end, so we can confirm the result visually without changing the underlying cardinality facts.

The remaining visual-review concern at 1440×900 is that the field types and legend labels are still fairly small and faint. A representative real or redacted model beyond the orders example would help us judge that tradeoff before integration.

TuGou-a:
Updated to current `dev` (`7d3594e`) with the reading round from the 2026-09-23 review. Head: `eab5922` — rebased again after your merge so the pull-request merge ref builds and CI can run; the only conflict was the generated `archify.zip`, everything else applied clean, and the artifacts below are from this head.

All four points are answered below, each with the artifact you asked for. The three before/after images live on a scratch branch so they stay out of the diff — `review-erd-round4` on the fork, safe to delete any time.

**1. The crow's foot now opens toward the entity it describes.**

The toes stand on that table's edge and the apex sits one foot back along the relationship; the `one` bar is unchanged and the optionality ring still sits past the apex, so the ring is never bisected by the foot. `er-many-end` and `er-many-start` remain mirrors of one shape, and only that shape moved — every declared cardinality draws exactly the glyph it drew before.

Before/after of the same relationship, plain `many` and `optional many`, same viewport and theme: [4-panel comparison](https://raw.githubusercontent.com/TuGou-a/archify/review-erd-round4/review/erd-round4/01-crow-foot-before-after.png). The legend swatch repeats the new orientation, since the legend is supposed to teach the glyph the diagram draws.

**2. Field types and legend labels.**

Measured in the real viewer at 1440×900 through computed styles (not from the source units): field names 10 → 11px, field types 9 → 10.5px, key glyphs 8 → 9, table names 12 → 13, legend labels 11 → 12, and relationship labels are now set explicitly at 9 units instead of the shared 8 — they had become the smallest text on a tall canvas. The type column also moved off the muted token: sampled from the rendered pixels against the table fill, the muted ink measured 3.98:1 (light) and 6.04:1 (dark), under the 4.5:1 text floor, which is what "faint" was measuring; the type now uses the primary ink at 14.94:1 / 15.48:1.

[Light](https://raw.githubusercontent.com/TuGou-a/archify/review-erd-round4/review/erd-round4/02-legibility-light.png) · [dark](https://raw.githubusercontent.com/TuGou-a/archify/review-erd-round4/review/erd-round4/03-legibility-dark.png) — same viewport, same crop, before on the left.

**3. The legend, and a larger diagram.**

You were right that this is about the generated content, not the viewport. An automatic canvas is now sized so the default legend is always in it: the shared measurement decides whether the reserved strip fits and the drawing area grows a band at a time until it does. If a route or a label still collides with the placed band, the renderer raises the capacity diagnostic instead of dropping the key, and an authored `meta.viewBox` — a fixed area — names exactly what it cannot hold (`legend/vertical-overflow`, `legend/label-too-wide`, or `legend/content-overlap`, each with its fix). The `unfit: 'hide'` path is gone from both routes.

For the scroll question: an automatic canvas declares `data-reader-fit="intrinsic-height"`, so a schema taller than one screen scrolls at a readable size instead of being squeezed, and field rows now declare `data-detail="context"` on the text itself so the Viewer sizes the diagram from the fields it has to keep legible. That was a real hole: the reader's own scale calculation only counted texts carrying the attribute, so a tall schema was shrunk past the floor while the check disagreed. Thanks for the pointer to `labs/infinite-canvas` and #478 — neither is a dependency here, and nothing in this branch anticipates them.

**4. A representative model beyond the orders example.**

[`examples/subscription-billing.erd.json`](archify/examples/subscription-billing.erd.json) — a redacted subscription-billing schema: eleven tables, sixty-four fields, eleven relationships, four domain bands (accounts, catalog, billing, payments). It passes the showcase profile 9/9 with 0 errors and 0 warnings, and at 1440×900 it reads as a normal vertical scroll with 7.5px minimum text (the floor is 6px).

[1440×900, both themes](https://raw.githubusercontent.com/TuGou-a/archify/review-erd-round4/review/erd-round4/04-representative-model-1440x900.png). One honest number for judging the tradeoff: the reader settles this one at scale 0.834, so its field types land at ~8.8px on screen against 10.5px for the orders example at 1440×900. The typography is the same; the diagram is simply larger. If you would rather have that compressed, the lever I would pull is rows, not type size.

**Also in this revision**

- The router is now `dev`'s current planner, moved to `renderers/shared/orthogonal-router.mjs` with the ERD contributing `sideFor`, `portSpacing`, `preferredCandidates`, `compositionFloors: false`, and a plan-free `inferredSides` read. Architecture and lifecycle changed one import line each and all six example renderings are byte-identical to the checked-in goldens. The ERD's own detour/outside candidate families are deleted — the shared obstacle grid covers them and the example renders byte-identically without them.
- Three defects my own review passes found and fixed: the bundled bus was painted over the relationship paths (a fan-in with an optional marker drew the bus through the ring's mask); an over-subscribed table side — the port spread is a cap, not a floor — drew overlapping feet and bars with nothing reported (now `layout/marker-capacity`, naming the table, the side, the spacing, the glyph height, and three ways out); and the relationship label's drawn mask was measured by the layout checks with the *old* 8-unit advance, so a label could sit on a table edge with no diagnostic. One measured box now serves all three consumers, and under the honest measurement the orders example's own label was 2px onto both tables — its corridor is widened to the room the diagnostic asks for, and the route shapes, bend counts, and gate outcome are unchanged by that.

**Evidence** (Node 22, this head)

```
cd archify && npm test
# tests 2319 · pass 2241 · fail 0 · skipped 78
npm run test:browser
# 246 · pass 246 · fail 0 on re-run (includes the desktop reader and site continuity)
# the first run flaked on two Chrome-contention cases; each passed alone and the
# full suite re-ran clean

node bin/archify.mjs deliver erd examples/orders.erd.json /tmp/orders.html --quality showcase --json
# ok · 9/9 checks · showcase pass · 0 errors / 0 warnings
node bin/archify.mjs visual-check /tmp/orders.html --json
# pass · 4 viewports x 2 themes · contained at all four · min projected text 8.6px
node bin/archify.mjs deliver erd examples/subscription-billing.erd.json /tmp/billing.html --quality showcase --json
# ok · 9/9 checks · 0 errors / 0 warnings
node bin/archify.mjs visual-check /tmp/billing.html --json
# pass · readable vertical scroll at 1440/1600/1920 · min projected text 7.51px

cd website && npm run check && npm run build && npm test   # typecheck clean · build ok · 7/7
bash scripts/build-zip.sh && cmp archify.zip /tmp/again.zip # deterministic
```

`deliver` on the orders example is 166 ms median of five (renderer alone 82 ms); for scale, `dev`'s own schemas cost every type about 50 ms more than this branch's fork point, byte-identically.

The PR description now describes this head. Both questions from the last round are closed your way — the foot flips, and the unfittable legend is a diagnostic rather than a silent drop.

TuGou-a:
@tt-a1i  review.  oh hhh

PR #402, opened 2026-09-13
fix(benchmark): reject malformed visual defect lists before passing runs

nbt0:
A passed visual review with a reviewer identity and `defects: "clipping"` is normalized to an empty list, allowing an otherwise valid first attempt to report `firstPassUsable: true`.

Require an explicit defects array before accepting a passed visual review. Missing or incorrectly typed lists now produce an invalid visual-review gate. Empty arrays still pass, and non-empty arrays retain their defects and prevent acceptance. Optional skipped reviews are unchanged.

Validation:
- `node --test archify/test/ordinary-model-floor.test.mjs` on Windows with Node 22.23.2: 19 passed, 0 failed.
- The regression covers omitted, null, string, object, boolean, numeric, empty-array, and non-empty-array values, while asserting the semantic and renderer gates pass.
- `git diff --check` passed.
- The full suite was not rerun on this isolated branch. An earlier run on the unchanged base had Windows/environment failures; this PR does not claim an all-green full suite.

### Visual evidence

Not applicable: only benchmark metadata validation and its regression test change; no Viewer or diagram presentation changes.

### Scope and artifacts

No schema or packaged runtime changes; no package regeneration is needed. This is a narrow fix with a concrete reproduction under CONTRIBUTING.md's small-fix path.

### DIFF
--- a/archify/test/ordinary-model-floor.test.mjs
+++ b/archify/test/ordinary-model-floor.test.mjs
@@ -200,6 +200,34 @@ test('benchmark never accepts a visual pass without an identified reviewer', () 
   assert.equal(receipt.firstPassUsable, false);
 });
 
+test('benchmark never upgrades missing or malformed visual defects to a clean pass', () => {
+  const caseFile = path.join(repoRoot, 'benchmarks/ordinary-model-floor/cases/web-runtime.architecture.case.json');
+  const candidate = path.join(skillRoot, 'examples/web-app.architecture.json');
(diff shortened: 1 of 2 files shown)


tt-a1i (approved):
Reviewed b4e257be592942b9e5b0d21494ccba6c3cc75980. No blocking findings. The explicit array check prevents malformed or missing defect lists from becoming a clean visual pass, while preserving empty-array acceptance and non-empty-array rejection. All 19 benchmark tests passed locally; the new regression fails against the main implementation. No packaged runtime or visual output changes. Current reported CI checks pass.

Issue #400, opened 2026-09-12
[Bug]: compare architecture validates mutable input paths instead of the exact snapshots used for hashes and delta


yixinforu:
### Archify version or commit

6db72a9aea3d0f67a6a034e41f8a5491476a11c1

### Installation method

Git clone

### Diagram type

Architecture

### Exact command

```shell
cat > snapshot.json <<'JSON'
{
  "schema_version": 1,
  "diagram_type": "architecture",
  "meta": {
    "title": "Snapshot consistency reproduction"
  },
  "components": [
    {
      "id": "api",
      "type": "backend",
      "label": "API",
      "pos": [40, 40],
      "size": [120, 60]
    }
  ],
  "boundaries": [],
  "connections": [],
  "cards": []
}
JSON

cp snapshot.json head.json

rm -f base.json
mkfifo base.json

(
  cat snapshot.json > base.json
  printf '{}\n' > base.json
) &

node archify/bin/archify.mjs compare architecture \
  base.json head.json delta.html --json
```

### Minimal redacted JSON reproduction

```json
{
  "schema_version": 1,
  "diagram_type": "architecture",
  "meta": {
    "title": "Snapshot consistency reproduction"
  },
  "components": [
    {
      "id": "api",
      "type": "backend",
      "label": "API",
      "pos": [40, 40],
      "size": [120, 60]
    }
  ],
  "boundaries": [],
  "connections": [],
  "cards": []
}
```

### Validation receipt or exact error

```json
The issue reproduces consistently using the FIFO-based reproduction above.

Exact `--json` output:


{
  "schemaVersion": 1,
  "ok": false,
  "command": "compare",
  "type": "architecture",
  "stage": "input",
  "error": "Base snapshot failed validation: architecture schema validation failed:\n  / must have required property 'schema_version' {\"missingProperty\":\"schema_version\"}\n  / must have required property 'diagram_type' {\"missingProperty\":\"diagram_type\"}\n  / must have required property 'meta' {\"missingProperty\":\"meta\"}\n  / must have required property 'components' {\"missingProperty\":\"components\"}",
  "diagnostics": [
    {
      "code": "schema/required",
      "severity": "error",
[...]

Issue #376, opened 2026-09-09
[Bug]: Auto-routing infers left/right for a hub above an offset spoke


drewdog88:
### Archify version or commit

`2.17.0-dev.1` (`10722002bb8777ecb639d93c49586fae4adf3ae4`)

### Installation method

Git clone of https://github.com/tt-a1i/archify

### Diagram type

Architecture

### Exact command

```sh
node archify/bin/archify.mjs deliver architecture hub-spoke.architecture.json /tmp/hub-spoke.html --quality standard --json
```

### Minimal redacted JSON reproduction

```json
{
  "schema_version": 1,
  "diagram_type": "architecture",
  "meta": {
    "title": "Hub above an offset spoke",
    "output": "hub-spoke.html",
    "quality_profile": "standard",
    "viewBox": [640, 360]
  },
  "components": [
    { "id": "hub", "type": "cloud", "label": "Hub", "pos": [200, 40], "size": [176, 52] },
    { "id": "spoke", "type": "cloud", "label": "Spoke", "pos": [40, 240], "size": [168, 52] }
  ],
  "connections": [
    { "id": "hub-spoke", "from": "hub", "to": "spoke" }
  ]
}
```

Hub center x is 288; spoke center x is 124. `defaultFromSide` therefore returns `"left"` even though the dominant gap is vertical.

### Expected behavior

Automatic routing should leave the hub from the **bottom** and enter the spoke from the **top** (dominant axis), or it should infer sides from the path it actually drew. `deliver` should succeed without authored `fromSide`/`toSide`.

Suggested direction: make `defaultFromSide` / `defaultToSide` pick the axis with the larger center delta so hub-and-spoke above a row works. A follow-up: `edge-through-node` `supportedFixes` should include a concrete `via` candidate.

### Actual behavior

Architecture auto-routing infers `fromSide`/`toSide` from **horizontal center order first**. A hub sitting above an offset spoke is therefore treated as “leave from the left / enter from the right.” The automatic router then draws a **vertical** dogleg, and Clean Flow rejects the path it just produced (`clean-flow/endpoint-side-direction`, `sideOrigin: "inferred"`).

The suggested fix in the diagnostic is “keep automatic routing so the renderer can use a side-aware bridge,” but this JSON already uses automatic routing (no `fromSide`, `toSide`, `route`, or `via`).

[...]

PR #405, opened 2026-09-13
fix: Windows repo-root, stderr pipe truncation, false-positive finite_svg, hub-spoke routing, compare TOCTOU (#274 #350 #372 #376 #400)

rollroyces:
## Summary

Five independent bug fixes, each scoped to a single file or a small
adjacent test, all addressing the same general theme: small toolchain
defects that produce wrong, late, or empty error output.

| # | Issue | Surface | Fix |
|---|---|---|---|
| #274 | `--repo-root` rejected on lowercase Windows drive | `archify/renderers/shared/repository-evidence.mjs` | Replace strict `===` with `pathsAlias`, which probes FS case sensitivity at runtime and lowercases only on case-insensitive mounts |
| #350 | renderer diagnostics truncated at 64 KiB through pipes | `archify/renderers/shared/diagnostics.mjs` | Loop `fs.writeSync` until the entire payload is accepted; tolerate EPIPE/ECONNRESET by swallowing write errors |
| #372 | `finite_svg` flagged text containing `NaN`/`Infinity` as non-finite coords | `archify/scripts/check-render-output.mjs` | Tokenize numeric SVG attributes (`x`, `y`, `x1`, `y1`, `x2`, `y2`, `cx`, `cy`, `r`, `rx`, `ry`, `width`, `height`, `viewBox`, `points`) and report only genuinely non-finite values, populating `details` |
| #376 | hub-and-spoke with offset spoke rendered horizontal connectors | `archify/renderers/shared/geometry.mjs` | Weight the dominant axis in `defaultFromSide` / `defaultToSide`; prefer vertical when `\|dy\| > 1.5 × \|dx\|` |
| #400 | `compare` re-read mutable input paths after validation | `archify/bin/archify.mjs` | Snapshot the validated buffers to staging files before handing them to the renderer subprocess |

## Test results

- 5 regression tests added: `pathsAlias` case variants (#274), `installRendererDiagnosticBoundary` 64 KiB pipe loop (#350), `finite_svg` true positive + false positive (#372), `defaultFromSide` / `defaultToSide` axis preference (#376), `compare` snapshot invariance (#400).
- New tests are deterministic and **pass on the test runner** when each fix is applied individually. (Some pre-existing CLI tests that depend on renderer stderr round-trips are sensitive to the local Node 22.22.2 × macOS pipe-buffer behaviour; they pass on the maintainer's Linux CI but fail on a macOS developer's machine regardless of these fixes. See the note below.)
- `npm test` overall: 1270/1305 pass, 35 fail on this dev machine; 1306/1306 pass, 0 fail on Linux CI. The 35 local-only failures are 100% `internal/unclassified` errors from the same Node subprocess-pipe truncation the project has been carrying since the renderer-diagnostic boundary was introduced.

### #274 — lowercase Windows drive letter

Strict `fs.realpathSync(gitRoot) !== realRoot` rejected user-supplied
`--repo-root c:\work` when git had recorded the worktree as
`C:\work`. The fix routes the comparison through the existing
`pathsAlias` helper from `archify/renderers/shared/output-path.mjs`,
which probes the directory's case sensitivity at runtime (the same
helper `resolveOutputPath` already uses to detect symbolic-link
aliases). On POSIX, the probe reports case-sensitive and the
comparison stays strict; on NTFS / macOS APFS / WSL with case=off,
the comparison lowercases before equality. No branch on
`process.platform` is added.

### #350 — stderr pipe truncation

`fs.writeSync(process.stderr.fd, payload)` was called once with the
full diagnostic JSON. macOS and Linux kernel pipe buffers are exactly
64 KiB; once the buffer fills, `writeSync` returns a short count and
the rest of the payload is silently dropped, leaving the parent CLI
unable to parse the diagnostics. Every diagnostic surfaces as
`internal/unclassified` instead of carrying the real schema rule and
supported fixes. The fix loops until every byte is accepted and
tolerates `EPIPE` / `ECONNRESET` from a closed reader.

### #372 — false-positive `finite_svg`

The check was a whole-SVG regex over the substrings `NaN`,
`undefined`, `Infinity`, `-Infinity`. Component `<text>` bodies that
happened to contain those literals — e.g. a `tag` reading "returns a
NaN leaf" — were reported as non-finite coordinates. The fix replaces
the regex with a per-attribute scan over the numeric attributes that
actually carry coordinates. CSS color tokens, function calls, and
hex strings are skipped because `Number.parseFloat` would return
`NaN` for them and they don't match the explicit non-finite literal
set. `details` is now populated with the offending attribute and
token, so the diagnostic is actionable.

### #376 — hub-and-spoke auto-route

`defaultFromSide` and `defaultToSide` chose left/right whenever the
two component centres had different `cx`, falling back to top/bottom
only when `cx` was exactly equal. A hub-and-spoke arrangement with the
hub above and the spoke offset to the side therefore rendered
horizontal connectors. The fix computes `|dx|` and `|dy|` and
prefers the dominant axis (`|dy| > 1.5 × |dx|` → vertical,
`|dx| > 1.5 × |dy|` → horizontal). Diagonals keep the historical
horizontal preference so existing diagrams do not shift when a
component is nudged by a few pixels.

### #400 — `compare` TOCTOU

`commandCompare` validated the user-supplied paths at the top, computed
the receipt hash and delta from those captured buffers, then handed
the original mutable paths to the renderer subprocess. The renderer
re-read the paths inside the architecture pipeline; a writer that
mutated either file between validate and render produced a diagnostic
that named bytes the receipt never saw. The diagnostic code disagreed
with the receipt; the Agent repair loop acted on bytes that were
not the bytes the diff was computed from. The fix snapshots the
captured buffers to deterministic staging files inside the existing
compare staging area, and the renderer reads from those snapshots
instead of the user-supplied paths.

## Local test infra note

On this dev machine (macOS, Node 22.22.2), 34 pre-existing CLI
integration tests fail with `internal/unclassified` from the renderer
subprocess because Node's `stdio: 'pipe'` drops bytes on
`process.exit(1)` timing when reading from a non-TTY stderr fd. The
same tests pass on the maintainer's Linux CI (Node 22.23.2 +
ubuntu-latest). The local-only failures are not caused by these fixes:
they appear in `archify/test/cli.test.mjs`,
`archify/test/architecture-delta.test.mjs`, and other renderer-stderr
round-trip tests, none of which were touched by this PR. If the
maintainer wants the local macOS suite to pass too, the underlying
cause is the renderer's `installRendererDiagnosticBoundary` losing
bytes to Node's pipe buffer — the same family as #350, but harder to
isolate because Node's stdio pipe and the OS pipe buffer are two
different layers with different behavior on Linux vs macOS. The #350
fix already drains the OS pipe buffer correctly; the Node stdio
truncation is a separate, larger problem worth its own issue.

Refs #274, #350, #372, #376, #400


tt-a1i:
Consolidating this multi-issue branch into the focused fixes during maintainer backlog cleanup. Three included problems are already fixed on main: renderer receipt truncation by #374, finite_svg false positives by #373, and compare input snapshots by #404. The remaining Windows repository-root problem stays open as #274 with #309/#276, and hub/spoke routing stays open as #376 with #379. We are preserving this PR’s pathsAlias-based filesystem-identity approach as review input for #274; closure does not claim those two remaining bugs are fixed. Keeping their focused integration paths avoids reintroducing alternate versions of the three already-shipped fixes.

PR #319, opened 2026-09-05
test: load workflow migration preload via a file URL

ouchihao:
## Problem and value

Passing a native Windows path directly to Node's `--import` prevents the workflow-migration cleanup regression from reaching the CLI (`ERR_UNSUPPORTED_ESM_URL_SCHEME`). Convert the existing fixture path with `pathToFileURL(...).href`, preserving the cleanup-failure injection and every committed-output/receipt/warning assertion. This also handles URL-significant characters in checkout paths.

## Stability impact

- Impact: local behavior, focused test correction. Only two lines in `archify/test/workflow-migration.test.mjs` differ from current main.
- No runtime, schema, package, or compatibility changes. The helper still loads the same fixture.
- Updated from reviewed head `16ef0cd` by merging main `1072200`; no additional feature/test logic was changed. Existing history is preserved.

## Tests run

Comparison base: `10722002bb8777ecb639d93c49586fae4adf3ae4`. Candidate: `bd61c2198f198158b49d47d65397130057361f55`.

On the updated candidate, Windows / Node 22.23.2:

- `node --test test/workflow-migration.test.mjs`: 20 passed, 1 failed, 0 skipped. The cleanup-warning regression passed; the remaining cyclic-symlink fixture fails before CLI execution at `fs.symlinkSync` with the existing host `EPERM` restriction.
- `git diff upstream/main --check`: passed. Current main is an ancestor of this head; the worktree is clean.

Reused evidence on `16ef0cd`: the target regression failed before the conversion and passed afterward. The test-helper/fixture change is unchanged in this update. [Independent Windows/Ubuntu verification](https://github.com/tt-a1i/archify/pull/319#pullrequestreview-5137129957) reports Ubuntu 21/21 and the Windows target passing, including spaces and `#` in paths; this is evidence for that original head, not a new-head execution.

No repeated local full renderer suite: current CONTRIBUTING.md permits targeted evidence for focused test-only corrections. Required remote CI still applies. On candidate `bd61c21`, [CI run 34227405336](https://github.com/tt-a1i/archify/actions/runs/34227405336) reports `action_required` and zero check runs: maintainer approval is still needed before CI can run. Branch status is now `BLOCKED` for required checks/review, not `BEHIND`; remote CI has not passed.

## Visual evidence

Not applicable: only the test helper's module specifier changes; no UI or diagram output changes.

## Generated artifacts

None regenerated for this correction. Runtime and package inputs match current main, including its existing `archify.zip`.

PR #409, opened 2026-09-14
test: fix Windows archive spawning and motion preference reloads

i-so-late:
## Problem and behavior

Two repository-only test harness corrections:

- The notifier archive gate directly spawned `build-zip.sh`, which fails to start on Windows. Reuse the existing `spawnBuildZip` helper so Windows invokes Bash and other platforms retain direct execution.
- The motion preference persistence test claimed to reload a page but changed its `file://` query on every navigation. After integrating main, Linux CI twice observed missing localStorage and `live` instead of `still`. Reload the exact same URL with `Page.reload`, retain fresh-document identity checks, and additionally assert that the URL and pre-startup preference survive all five reloads. File-URL localStorage behavior is browser-dependent: https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage

## Impact

Only `archify/test/release-package-gates.test.mjs` and `archify/test/motion-governor-browser.test.mjs` change. No runtime, renderer, schema, dependency, or packaged artifact change. Existing motion behavior assertions remain; the fix does not seed storage or skip failures. Test files are excluded from the distributable ZIP.

## Maintainer validation

Base main: `f631114`; candidate: `3c13b8a`.

- Official Node 22.21.1, macOS: the notifier archive target passed after main integration.
- Motion browser suite with real Chrome: 8 passed, 0 failed on the candidate. The previous harness also passed locally; the two failures were on Linux CI, so local success alone does not establish the Linux fix.
- Full CI-equivalent Viewer browser group on macOS, real Chrome, concurrency 2: 106 passed, 0 failed, 0 skipped.
- Final head `3c13b8a95ee8497ed7ae0c8bbe63f177b2a839af`: all nine required CI checks passed, including Linux WebM/Viewer browser checks, and CodeRabbit passed. CI run: https://github.com/tt-a1i/archify/actions/runs/34866644399. No open review threads.
- `git diff --check`: clean.

## Original Windows validation

Base `a07fa1d` -> candidate `c159256` (one commit). Windows 10 Pro 10.0.19045, Git for Windows 2.32.0, GNU bash 4.4.23 (msys), Node v22.21.1, npm 10.9.4, `unzip` from Git for Windows, clean checkout.

- Targeted, from `archify/`: `node --test --test-name-pattern="built archives contain the embedded notifier runtime" test/release-package-gates.test.mjs` -> base: 1 fail (`null !== 0`); candidate: 1 pass.
- `node --test test/release-package-gates.test.mjs` -> base: 23 tests, 20 pass / 2 fail / 1 skip; candidate: 23 tests, 21 pass / 1 fail / 1 skip. The test that stops failing is this gate. The remaining failure on both is `archive build excludes untracked files and external symlinks from the live working tree`, which needs symlink privilege (`EPERM` from `fs.symlinkSync`) and is unrelated. The skip on both is `archive build rejects non-canonical Node versions before publishing output`, which requires a Node major other than 22.
- `npm test` from `archify/` (after `npm ci`) -> base: 1348 tests, 1267 pass / 24 fail / 57 skip; candidate: 1348 tests, 1267 pass / 24 fail / 57 skip. The failing names differ by exactly one swap: `built archives contain the embedded notifier runtime` fails on base and passes on the candidate; `a last-good notice remains acknowledgeable after the refresh commits a new candidate` (`test/update-notifier.test.mjs`, unchanged by this PR) failed in the candidate's full run and passes 3/3 when run in isolation on the candidate. That test and its sibling `an empty precheck snapshot cannot start a second concurrent network request` are the same two update-notifier cases #349 reported as intermittent on Ubuntu; the sibling fails in both full runs here. The other 22 failures are identical on base and candidate and are Windows-environment causes: 16 need symlink privilege (`EPERM` from `fs.symlinkSync`), 3 are the `core.autocrlf` checkout cases (#318), and one each are the stager stat-mode assertion (#349), the installed-preview shutdown (#320), and the workflow-migration preload path (#319).
- `node --check` on the test file and `git diff --check`: clean.
- Linux/macOS: not run locally. The changed call site takes the non-win32 branch of `spawnBuildZip`, which is the same direct `spawnSync` the test used before, so the ubuntu `npm test` job exercises identical behavior.


tt-a1i (dismissed):
Reviewed the exact head. This narrowly reuses the existing spawnBuildZip helper for the notifier archive test, preserving non-Windows behavior and using the established Bash path on Windows. The targeted test passed locally on macOS with official Node 22.21.1; all required CI checks passed. No blocking findings.

PR #415, opened 2026-09-14
fix(delta): report brand and locale changes in architecture comparisons

kapelame:
## Problem and value

`compare architecture` reports a complete, empty delta when two valid snapshots differ only in a component's `brand` or in `meta.locale`. For example, changing a `Cache` node from `brand: "redis"` to `brand: "postgresql"` produces `summary.components.changed: 0`, no exact change row, and an unchanged node in the Delta view. Changing only the locale to `zh-CN` likewise leaves `presentationChanged: false`.

The comparator's explicit field lists omit these supported authored fields. This adds `brand` to the existing semantic component fields and `locale` to both presentation objects. Public-CLI regressions cover adding, replacing and removing a brand, plus a locale-only comparison. No separate planning issue is needed for this narrow reproduced defect under CONTRIBUTING.md.

## Stability impact

- **Impact class:** local comparison behavior. Affects the compare JSON receipt, exact change navigator, Delta node highlighting and summaries that consume the receipt.
- Existing component/relationship IDs, topology, geometry, evidence, schema validation and atomic output behavior are preserved. Brand changes use the existing semantic classification; locale changes use the existing presentation flag.
- No schema, version, defaults, runtime dependency or unrelated refactor changes.

## Tests run

Comparison base: `851b279f3710c3ed6f152f4b044a504ca4eb207c`; candidate: `d0630312b0a881d25b60a1c7fd2e760204270977`.

Current-base verification on macOS arm64:

- Before applying the runtime fix to this base: `node --test --test-name-pattern='brand-only|locale-only' archify/test/architecture-delta.test.mjs` — both new tests failed (`0 !== 1` and `false !== true`, Node 25.9.0).
- After applying the same comparator/test patch: `npm test` with Node 22.19.0 — all generated-input and golden checks passed; 1,359 tests, 1,311 passed, 0 failed, 48 skipped. This includes the 22 Delta tests and updated render-output/package checks. Skips cover the opt-in browser matrix, external pinned MCO checkout, the non-Node-22 archive check, Windows drive-path check and serialized site integration.
- `ARCHIFY_CHROME=<Chrome executable> node --test test/motion-governor-browser.test.mjs` with Node 22.19.0 — the upstream-modified browser suite passed: 8 tests, 0 failed, 0 skipped. This was run separately from the default suite above.
- `scripts/build-zip.sh <audit>/archify-rebuilt-refresh.zip && cmp archify.zip <audit>/archify-rebuilt-refresh.zip` with Node 22.19.0 — byte-identical repeat build. `<audit>` abbreviates the recorded audit output directory.
- After extracting the refreshed `archify.zip` outside the checkout: `node scripts/package-smoke.mjs <audit>/extracted-refresh/archify` with Node 22.19.0 — package smoke passed without installing dependencies.
- Re-ran public `compare architecture` with the exact brand-change fixture: one semantic `/brand` change, and the generated HTML byte-matches the previously browser-reviewed fixed artifact.
- `git diff --check` — passed; only comparator, regression tests and required archive differ from the current base.

Previously recorded evidence is retained with its original revision: on `50e2dccf` plus the same fix (commit `347ddbf`), the focused Delta suite passed 22/22, CLI/preview/output/update tests passed 170/170, and doctor/demo checks passed. These are not relabeled as fresh runs. The current full suite reruns the relevant automated coverage.

The unchanged WebM and DeepSeek Harness integrations were not exercised. Native Windows/Linux and the full opt-in browser matrix remain for their upstream CI jobs. No paid model calls were used.

## Visual evidence

The browser screenshots/DOM observations below were recorded on the original base plus the fix (commit `347ddbf`). After refreshing to the current base, the same CLI fixture produces byte-identical HTML; comparator, renderer, template and localization inputs also byte-match that candidate. The existing visual evidence is reused on that explicit basis.

The new public-CLI regression provides the reproducible minimal input: a single database node with stable ID `cache`, label `Cache`, position `[100, 100]`, size `[160, 80]`, and brands `redis` / `postgresql`. Run `compare architecture` against those snapshots before and after the fix.

- **Comparison conditions:** the same JSON files, Chrome 152.0.7977.83, 1440×900, dark theme, an isolated temporary profile, offline mode and HTTP/HTTPS blocked.
- **Before:** zero changes, no change row, `cache` marked `same`, and “No authored architecture changes.”
- **After:** one semantic change, an enabled `cache` row identifying `/brand`, and the existing changed-node outline/marker.
- **Automated browser checks:** receipt, node state and exact change row verified from the loaded artifact DOM.
- **Perceptual visual review:** passed for this fixed fixture; the changed-node outline and `/brand` row are visible. No renderer layout or Viewer interaction code changed. This is not a claim that the full browser matrix was run.

## Generated artifacts

Rebuilt `archify.zip` from current base `851b279f` plus the fix with Node 22.19.0. Both current-base and candidate archives have 79 entries; the only changed member content is `archify/delta/architecture-delta.mjs`. The new upstream `scripts/check-render-output.mjs` bytes are preserved exactly. Repeated builds byte-match.

Existing gallery/examples contain no newly detected brand-only or locale-only delta, so their outputs remain unchanged; the generated-content and golden checks passed. No release identity changed.


tt-a1i (approved):
Reviewed d0630312b0a881d25b60a1c7fd2e760204270977 against 851b279f3710c3ed6f152f4b044a504ca4eb207c. No blocking findings. The two new regressions fail against the base comparator; all 22 Delta tests and 9 additional comparator cases pass on the candidate. Full local suite: 1310 passed, 48 skipped, one archive byte check failed with Homebrew Node; that check passed with official Node 22.19.0. All 79 decompressed archive members match the rebuild, and only the comparator payload differs from base. Browser perceptual evidence is reused from the author, not independently rerun. Current remote checks pass.

PR #446, opened 2026-09-16
fix(text-fit): preserve conservative width estimates for VS15

tt-a1i:
## Problem and behavior

Integrate #411 while correcting its width explanation and refreshing it for dev. The previous estimator counted every VS15 sequence as one unit, including wide CJK bases whose font ignores the selector. A real-Chrome probe on dev rendered ten such characters at 109.97 SVG units inside a 100-unit node while the command returned success.

Keep the base character's conservative width estimate for VS15. The same CJK input now fails the existing layout check, with or without the selector. VS16 emoji estimates and text without selectors are unchanged. Actual glyph advance remains font-dependent; this is an estimate, not a promise that a text variant has the same rendered width as its base. Some VS15 symbol labels may now require more room even when a particular platform's glyphs happen to fit. See Unicode UAX #11: https://www.unicode.org/reports/tr11/.

## Integration and validation

Candidate `0bd5773` preserves the original author commits. Direct push to the author's fork was denied, so this maintainer branch carries the refresh. It also includes the pending #364 authoring clarification and #445 browser-job fix; merge those first or use the passing combined-head evidence and preserve history with a merge commit.

- 191 geometry/layout/output-check regressions passed, no failures/skips; golden checks passed after the behavioral and explanatory fixes.
- After integrating #364/#445, reran all 56 geometry tests: all passed, zero skips. Rebuilt the canonical ZIP with official Node 22.23.2.
- Fresh baseline Chrome measurements prove the CJK overflow. Public CLI probes confirm rejection after the fix. The author's symbol-only example measured 98.94 units on this machine, so it is not presented as proof of overflow; the CJK fixture supplies that evidence.
- git diff --check passed. Current-head full hosted CI remains required.

The change affects shared text estimates used by all five renderers. No font, schema, Viewer or renderer drawing code changes. Existing golden artifacts remain unchanged; archify.zip includes the updated shared estimator and integrated documentation.

Follow-up: the sequence browser gate now distinguishes an absent ARCHIFY_CHROME from an explicitly empty value. Verified unset = skip, empty = fail, configured Chrome = pass, so an invalid CI browser configuration cannot silently skip its check.


tt-a1i:
Current head 0bd577359b0e604602f0bb6e2db0b01094450b8a passed full CI 35123764165, including the real-browser job, and all review threads are resolved. The intermittent Motion storage failure observed on #364 is tracked separately in #447 with retained failing-run and controlled Linux evidence; this passing combined run is not presented as fixing that issue. Integrating this reviewed text-fit/authoring change into dev only, preserving original commits. Main promotion remains pending actual dev stability.

PR #455, opened 2026-09-17
fix(preview): close owned connections during forced shutdown

### LINKED ISSUE #453: [Bug]: Preview forced shutdown waits for unfinished HTTP connections
# Preview forced shutdown waits for unfinished HTTP connections

Base: `72c750bb070d95171dbb2244e5b62b1b7da69c12`.

## Reproduction and observed behavior

Start the public preview CLI with the shipped `web-app.architecture.json` and
`--no-open`, then wait until `/state` reports `verified`. Connect a local TCP
client to the printed loopback port and send only:

```js
socket.write('GET /state HTTP/1.1\r\n');
```

The request headers are intentionally unfinished. Send SIGINT twice to the CLI.
The second signal prints `forcing preview shutdown…`, but the process remains
alive at least 1000ms later, with its staging directory still present. Destroying
the client socket immediately allows normal exit with code 0.

This reproduces without a delivery subprocess still running, injected failures,
or source modifications. It was observed with the real CLI on Node 24.18.0 and
independently with `startPreview().stop({ force: true })` on Node 22.23.1.

## Cause

`closeServer()` calls `server.close()` and `closeIdleConnections()`. An incomplete
request is not idle and prevents the server-close callback from completing.
`stop({ force: true })` only force-kills the delivery child; it does not close
these HTTP connections. `finishStop()` consequently keeps waiting for
`serverClosed`, even after the explicit forced-stop request.

## Bounded fix

Force-stop should terminate the server's remaining connections as well as any
delivery child, then complete normal cleanup. Keep the first Ctrl-C's existing
graceful delivery-drain behavior. Add regression coverage for incomplete HTTP
requests, SSE clients, an active child, repeat stop calls, staging cleanup, and
preservation of the last delivered artifact.

The package declares Node >=18. `server.closeAllConnections()` was introduced
in Node 18.2, so the implementation should cover the declared minimum rather
than silently depending on that method being available. Tracking owned sockets
and destroying them during forced stop is one bounded option.

This is distinct from the Windows watcher issues and the existing tests of
delivery-child shutdown. It is an edge case involving an unfinished client
request, not a claim that every ordinary preview session hangs.

To3akaRin:
## Problem and value

Closes #453.

With a client holding an unfinished HTTP request, a second Ctrl-C prints `forcing preview shutdown…` but Preview keeps waiting for the server to close. Track Preview-owned sockets and destroy them on explicit forced stop, including connection events queued during shutdown.

## Stability impact

- Targets `dev`; integrates `d2f92292ada4c07cf06511392f5e1e6bbb19c607` with the original #455 fix. Candidate: `dcc1db983f1bc78305ee490b016db060c8ca2f15`.
- First Ctrl-C retains graceful delivery draining. Explicit force closes HTTP connections and the active delivery process. Repeated stop calls remain safe and the last verified output bytes are preserved.
- Preserve dev's identity-bound cleanup: remove only files whose ownership can be verified. Delivery interrupted before Preview receives its receipt may leave unconfirmed files and recovery material in private staging. Unknown contents are not recursively deleted.
- No new dependencies, CLI options, diagram schema or Viewer changes. Socket APIs remain compatible with Node >=18.

## Changes since the reviewed head

The dev audit found the original fake-child test did not cover real delivery interruption and the original cleanup promise was too broad. Add deterministic public Preview regressions at two real delivery boundaries: after writing the internal specification snapshot, and after successful delivery writes HTML/receipt but before the receipt reaches Preview. Both exercise an unfinished HTTP request and verify bounded shutdown, unchanged prior output, preservation of recovery bytes and an external claimant, removal of Preview's owned input snapshot, repeated stop, and listener closure. Update the canonical delivery contract accordingly.

## Validation

- macOS, Node 22.23.1: `node --test archify/test/preview-force-stop.test.mjs archify/test/preview.test.mjs`: **37 passed, 0 failed, 1 skipped** (Windows-only test).
- Ablation: removing explicit socket destruction produces **6 failures / 1 pass**, including both new real-delivery windows. Restored the necessary runtime logic; no permanent runtime deletion.
- Full local `npm test`: **1998 passed, 0 failed, 69 skipped**; generated-source checks and golden renders also passed. Optional browser skips are not passes.
- `git diff --check` passed.
- Final-head CI: https://github.com/tt-a1i/archify/actions/runs/35488539839. Remote browser and platform results are separate from local coverage.

## Generated artifacts

Rebuilt `archify.zip` from combined dev source using Node 22. Two independent builds are byte-identical. SHA-256: `15bf2eff6dfeecf66da216473a22ad79029fa61ce813d37749035ef3e4c3a862`.

Extracted outside the checkout and ran `node scripts/package-smoke.mjs /tmp/pr455-package-check/archify`: passed on macOS. No live Skill installation was changed.


tt-a1i (commented):
Reviewed `6cae7c9` against current main `72c750b`. The socket tracking is a small, appropriate fix for #453. It also handles connections queued as forced shutdown begins, while leaving an ordinary stop alone.

I reran `node --test test/preview-force-stop.test.mjs test/preview.test.mjs` from a clean snapshot on macOS with Node 26.8.1: 16 passed, no failures or skips. That covers the real unfinished HTTP request, second Ctrl-C, SSE clients, active child cleanup and preservation of the last artifact. The existing Node 18/20/22/24, package and browser/WebM CI checks are also green on this head.

I found no blocking issue in the proposed change.

PR #456, opened 2026-09-17
fix(brands): decode icon URLs and respect HTML head contexts

### LINKED ISSUE #454: [Bug]: Brand capture requests literal HTML entities in icon URLs
# Brand capture requests the literal HTML entity in an icon URL

Base: `72c750bb070d95171dbb2244e5b62b1b7da69c12`.

A page may declare its icon with a normally escaped HTML attribute:

```html
<link rel="icon" type="image/png" href="/mark.png?theme=dark&amp;size=32">
```

`archify brands capture <page-url> --json` requests
`/mark.png?theme=dark&amp;size=32` instead of `/mark.png?theme=dark&size=32`.
If the server requires the intended query and has no fallback favicon, capture
exits 2 even though the declared icon is available. This also affects later
digest-pinned capture verification because it uses the same discovery path.

## Reproduction

From the repository root, run this disposable loopback HTTP fixture with Node:

```sh
node --input-type=module <<'JS'
import http from 'node:http';
import { spawn } from 'node:child_process';
const requests = [];
const png = Buffer.from('iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAQAAAC1HAwCAAAAC0lEQVR42mNk+A8AAQUBAScY42YAAAAASUVORK5CYII=', 'base64');
const server = http.createServer((req, res) => {
  requests.push(req.url);
  if (req.url === '/plain' || req.url === '/escaped') {
    res.setHeader('Content-Type', 'text/html');
    res.end('<!doctype html><link rel="icon" type="image/png" href="/mark.png?theme=dark' + (req.url === '/escaped' ? '&amp;' : '&') + 'size=32">');
  } else if (req.url === '/mark.png?theme=dark&size=32') {
    res.setHeader('Content-Type', 'image/png');
    res.end(png);
  } else {
    res.writeHead(404);
    res.end('missing');
  }
});
await new Promise(resolve => server.listen(0, '127.0.0.1', resolve));
try {
  for (const page of ['/plain', '/escaped']) {
    const start = requests.length;
    const child = spawn(process.execPath, ['archify/bin/archify.mjs', 'brands', 'capture', `http://127.0.0.1:${server.address().port}${page}`, '--json'], {
      env: { ...process.env, ARCHIFY_BRAND_ALLOW_PRIVATE: '1' },
      stdio: ['ignore', 'pipe', 'pipe'],
    });
    let stderr = '';
    child.stdout.resume();
    child.stderr.on('data', chunk => { stderr += chunk; });
    const exit = await new Promise(resolve => child.on('close', resolve));
    console.log({ page, exit, requests: requests.slice(start), stderr: stderr.trim() });
  }
} finally {
  server.closeAllConnections?.();
  await new Promise(resolve => server.close(resolve));
}
JS
```

`ARCHIFY_BRAND_ALLOW_PRIVATE=1` is scoped only to fixture child processes to
permit the loopback test server. No real external website or credentials are
used.

Observed: `/plain` exits 0 and fetches the intended query; `/escaped` fetches the
literal `amp;` query, tries `/favicon.ico`, and exits 2 with HTTP 404.

## Bounded correction

Decode common escaped URL attribute references (`amp`, `quot`, `apos`, `lt`,
`gt` and decimal/hexadecimal numeric references) once after extracting `href`,
before resolving it against the page URL. Preserve URL percent encoding and
attribute-context rules for ambiguous bare ampersands. This need not introduce
a general HTML parser, new runtime dependency, or change the existing URL,
network, MIME, size, timeout, and digest checks.

Regression should check the actual requested URL and returned digest, including
numeric references, nested escapes, normal queries, and digest-pinned re-fetch.

To3akaRin:
## Problem and value

Closes #454.

On `dev`, a favicon href such as `/mark.png?a=1&amp;b=2` is requested with a literal `amp;`, and the streaming HTML reader mistakes `</head>` inside a comment or script string for the real head ending. Both cases miss an available declared icon.

Decode extracted href references once before URL resolution, and track comments, raw-text elements and quoted attributes when finding the explicit head boundary. The byte scanner retains linear scanning, bounded memory and early cancellation without adding runtime dependencies or a general HTML parser.

## Stability impact

- Shared behavior: `brands capture` and digest-pinned validation/rendering for architecture, workflow, sequence, dataflow and lifecycle use the same capture path.
- Preserve percent escapes, one-pass entity decoding and ambiguous attribute references. The named-reference set remains limited to the documented basic names.
- Preserve the 256 KiB HTML cap, image limits, total deadline, URL/destination/redirect checks, MIME checks and pinned SHA-256 verification. Oversized bodies after the head remain unread.
- No schema, CLI, release identity or Viewer layout changes.

## Latest dev integration

Head `a9996ea6` merges latest `dev` (`dd3a3547`, including PR #455). The only conflict was `archify.zip`; source and changelog merged automatically. Rebuilt the archive from combined source rather than selecting either old archive. Comparing both old ZIPs confirmed that the PR archive lacked the latest Preview fix and the dev archive lacked the brand fix; the rebuilt archive byte-matches both runtime files.

- Combined brand/Preview regressions: **99 passed, 0 failed, 1 Windows-only skip** (100 tests).
- Outside-checkout package smoke: passed on macOS.
- Two builds are byte-identical. Current ZIP SHA-256: `141fe496d016524970ae3b1c6c9272486cfd5c965df2e2d4c28cd6bfcc3afa25`.
- No new abstraction or runtime edit was introduced in the conflict resolution. Prior brand ablation and full-suite evidence below apply to `de9268f6`; they are not relabelled as a new-head full-suite run. Fresh remote CI is required for `a9996ea6`.

## Tests run (prior head `de9268f6`)

Comparison base: `dev` at `d2f92292ada4c07cf06511392f5e1e6bbb19c607`. Candidate: `de9268f6` (merges dev into the existing PR branch). Local environment: macOS, Node 22.23.1.

- Before the head fix, the new 30-case href suite produced **16 passes / 14 failures**. The audit's two comment/script reproductions both requested the fallback instead of the declared icon.
- `node --test test/brand-href-entities.test.mjs test/brand-marks.test.mjs`: **62 passed, 0 failed, 0 skipped**.
- `npm test`: **2021 passed, 0 failed, 69 skipped** across 2090 tests, including generated-source checks and golden renders. Browser/platform skips are not passes; hosted CI remains required.
- Regressions cover one-byte response chunks, split entities/end tags, comments, script/style/title text, quoted attributes containing `>` before `</head>`, raw end-tag prefixes, casing/spacing, unquoted hrefs, a body above the cap, and capture → pinned validate/render in all five diagram types.
- Ablation: removing comment handling caused 6 failures; removing raw-text handling caused 11; removing quote handling caused 2. All three were restored; nothing was permanently removed. The quote ablation exposed and prompted strengthening an initially insufficient fixture. Restored focused suite: 62/62.
- Syntax and diff checks passed.

## Generated artifacts (prior head `de9268f6`)

Rebuilt `archify.zip` from the final combined dev source. Two Node 22 builds are byte-identical; SHA-256: `d26ae5a1989938760ffa14be1cb1f576f2aaf46fa2ed9cb994ccc79854710f7b`.

Extracted outside the checkout: `node scripts/package-smoke.mjs /tmp/pr456-package/archify` passed on macOS. Re-ran all 10 PR #456 scenarios from the dev audit against the extracted package: **10 passed, 0 failed, 0 skipped**, including all five pinned renderer paths and both original head-boundary failures. Viewer/examples inputs are unchanged, so no Gallery regeneration is needed. Linux/Windows and real-browser coverage remain for hosted CI.


tt-a1i (commented):
Reviewed `9db5e31` against current main `72c750b`. Decoding after attribute extraction and before URL resolution is the right place for this. The one-pass behavior and explicit limit on supported named entities keep the fix reasonably small.

I reran `node --test test/brand-href-entities.test.mjs test/brand-marks.test.mjs` from a clean snapshot on macOS with Node 26.8.1: 42 passed, no failures or skips. The tests exercise the actual requests, subsequent pinned validation/rendering, percent escapes, nested escapes and rejection of a decoded non-HTTP URL. The existing Node 18/20/22/24, package and browser/WebM CI checks are green on this head.

I found no blocking issue in this fix.

PR #462, opened 2026-09-17
fix(workflow): apply shared cross-collection contracts in the compiler

### LINKED ISSUE #429: [Bug]: compileWorkflow accepts documents rejected by renderer semantic validation
Reproduced on live `main` at `d673e8300df60a5c8166abe78787fdc78f6b8000` with Node `v22.23.2` and `v26.8.1` on macOS (Git clone).

`compileWorkflow()` returns `ok: true` for duplicate `edges[].id` and for a guided view whose focus refers to a missing node. The workflow renderer rejects the exact same documents before compilation, with `relationship/duplicate-id` and `guided-view/invalid` respectively.

The compiler currently calls `validateSchema()` but does not apply the shared cross-collection validators used by `loadDiagram()`. This is a concrete validation-parity issue, independent of a broader supported library API. The compiler already has production callers such as workflow migration, in addition to its tests.

Expected: the compiler should report failure diagnostics for these invalid documents, consistently with the renderer. Preserve existing valid input and structured failure receipts.

Run from the repository root; the reproduction only creates and removes its own temporary directory:

```sh
node --input-type=module <<'JS'
import assert from 'node:assert/strict';
import fs from 'node:fs';
import os from 'node:os';
import path from 'node:path';
import { spawnSync } from 'node:child_process';
import { compileWorkflow } from './archify/renderers/workflow/workflow-compiler.mjs';

const baseline = {
  schema_version: 2,
  diagram_type: 'workflow',
  meta: { title: 'Validation parity', legend: { mode: 'hidden' } },
  lanes: [{ id: 'main', label: 'Main' }],
  nodes: ['a', 'b', 'c'].map((id, index) => ({
    id, lane: 'main', col: index * 2, type: 'backend', label: id,
  })),
  edges: [
    { id: 'first', from: 'a', to: 'b' },
    { id: 'second', from: 'b', to: 'c' },
  ],
};
const directory = fs.mkdtempSync(path.join(os.tmpdir(), 'archify-validation-parity-'));
try {
  for (const scenario of ['duplicate relationship ID', 'unknown guided-view focus']) {
    const workflow = structuredClone(baseline);
    if (scenario === 'duplicate relationship ID') workflow.edges[1].id = 'first';
    else workflow.meta.views = [{ id: 'view', label: 'View', focus: ['missing'] }];
    const compiled = compileWorkflow({ workflow, qualityProfile: 'standard' });
    const input = path.join(directory, 'input.json');
    fs.writeFileSync(input, JSON.stringify(workflow));
    const rendered = spawnSync(process.execPath, [
      'archify/renderers/workflow/render-workflow.mjs', input,
      path.join(directory, 'output.html'),
    ], {
      encoding: 'utf8',
      env: { ...process.env, ARCHIFY_DIAGNOSTIC_FORMAT: 'json', ARCHIFY_QUALITY_PROFILE: 'standard' },
    });
    const failure = JSON.parse(rendered.stderr.trim());
    console.log(JSON.stringify({
      scenario,
      compilerOk: compiled.ok,
      rendererExitCode: rendered.status,
      rendererDiagnosticCodes: failure.diagnostics.map((entry) => entry.code),
    }));
    assert.equal(compiled.ok, true);
    assert.equal(rendered.status, 1);
  }
} finally {
  fs.rmSync(directory, { recursive: true, force: true });
}
JS
```

Actual stdout (exit code 0):

```jsonl
{"scenario":"duplicate relationship ID","compilerOk":true,"rendererExitCode":1,"rendererDiagnosticCodes":["relationship/duplicate-id"]}
{"scenario":"unknown guided-view focus","compilerOk":true,"rendererExitCode":1,"rendererDiagnosticCodes":["guided-view/invalid"]}
```

Relevant code: `archify/renderers/workflow/workflow-compiler.mjs` around lines 773–802 and `archify/renderers/shared/cli.mjs` around lines 23–25.

A small fix could reuse one filesystem-independent validation entry point from both callers, with a parity regression for these two cases. No general-purpose renderer library extraction is required.

## Related work

#222 tracks a broader public renderer API. #418 isolates compiler imports from CLI diagnostic handlers and explicitly leaves loader validation and environment-independent configuration outside its scope. This report is a separate, bounded correctness issue in the existing compiler. Open and closed issues and related PRs were searched before filing.

gold-beyond:
## Problem and value

`compileWorkflow()` applied `validateSchema()` but skipped the cross-collection passes `loadDiagram()` runs, so it accepted documents the workflow renderer rejects. Both reproductions in #429 still hold on the current `dev` base (with the now-required `meta.output` added):

| input | `compileWorkflow()` before this change | `render-workflow.mjs` |
| --- | --- | --- |
| `edges[1].id` duplicates `edges[0].id` | `ok: true` | exit 1, `relationship/duplicate-id` |
| `meta.views[0].focus` names a missing node | `ok: true` | exit 1, `guided-view/invalid` |

The compiler is not test-only: `archify/migrations/workflow-v2.mjs` calls it from eight sites on migrated and probe documents. A document that can never render could therefore be produced and accepted by a migration path. Reusing one shared entry point (rather than copying the two checks) is what keeps the callers from drifting apart again. Closes #429.

## Stability impact

- **Impact class: shared behavior.** `archify/renderers/shared/cli.mjs` gains one exported entry point, `validateCrossCollectionContracts(diagramType, diagram)`, which runs the existing `validateGuidedViews()` and `validateRelationshipIds()`; `loadDiagram()` now calls that entry point instead of the two passes directly, so nothing is duplicated. `archify/renderers/workflow/workflow-compiler.mjs` imports it and calls it beside the existing `validateSchema()`.
- **Callers traced:** `renderers/workflow/render-workflow.mjs` (CLI) and `migrations/workflow-v2.mjs` (eight call sites). Both receive the narrower acceptance surface; migration tests pass unchanged.
- **Existing behavior preserved.** Valid input is untouched: the renderer suite, `test/golden.mjs` and the four `check:*` freshness gates all pass unchanged, so no rendered byte moved. The new pass runs on the same quality-resolved document `validateSchema()` already receives, before canonicalization, so it cannot see a transformed topology.
- **Failure behavior.** Newly rejected documents leave through the compiler's existing structured failure receipt — `ok: false`, `diagnostics[]`, `receipt.contract: readable-v2` — carrying the same stable codes the renderer reports, with `supportedFixes: []`. No new exit path, no new diagnostic code, no schema, default or acceptance-rule change.
- **Scope of the shared entry point.** Its comment names only the relationship-ID and guided-view checks; it does not promise parity with every renderer validation.
- **Deliberately out of scope:** `validateEngineeringProfile()` remains a `loadDiagram()`-only check; the compiler does not call it. Adding it would change compiler acceptance for engineering-profile documents, which is a broader compatibility decision than this issue's agreed scope. Happy to open a separate issue if you want that gap closed on the same footing.

## Tests run

The branch is one commit on top of the current `dev`. The archive conflict against a moved `dev` tip was resolved by rebasing onto `dev` and regenerating `archify.zip` from the combined source.

- **New regression** in `test/workflow-compiler-hard-contract.test.mjs`: `compileWorkflow applies the shared cross-collection contracts the renderer enforces`. For each case it spawns the real renderer CLI (`renderers/workflow/render-workflow.mjs`) on the mutated document and asserts exit 1 with the expected code in its JSON diagnostics, then that `compileWorkflow()` fails with exactly that code, no `internal/unclassified`, and `supportedFixes: []`.
- **Red on base:** with only the compiler-side call reverted, the CLI still rejects the document while `compileWorkflow()` reports `ok: true`, so the new test fails — it demonstrates the original defect rather than restating the fix.
- **End-to-end parity:** the reproduction script from #429 now reports `compilerOk: false` with diagnostic codes identical to the renderer's for both scenarios.
- **Targeted suites on this checkout** (`node --test test/workflow-compiler-hard-contract.test.mjs test/workflow-migration.test.mjs`, Node v22.23.1): 103 tests, 101 pass, 2 fail. Both failures are the symlink cases (`EPERM` on symlink creation) and the unmodified `dev` tree fails the same two (102 tests, 100 pass, 2 fail), so they are environment-specific rather than caused by this change. **Coverage gap:** this checkout cannot create symlinks; hosted CI remains the authority there.
- **CI on the final head:** all jobs green — the four Node versions, package smoke on ubuntu/macos/windows, `zip-freshness`, `published-update-manifest`, `webm-artifact` and the Windows portability checks (Pages is conditionally skipped).

## Generated artifacts

- **`archify.zip` regenerated**: both changed sources ship inside the packaged Skill. Built with `scripts/build-zip.sh` on Node v22.23.1 (the canonical major; the builder rejects other majors). This is also what the in-suite archive gate asserts, so a stale archive fails `test/release-package-gates.test.mjs` until it is rebuilt.
- **Toolchain check:** rebuilding the pristine `dev` tree with the same script in this Windows checkout reproduces the committed `archify.zip` byte-for-byte, and the branch archive was then rebuilt from the combined source the same way — so the regenerated archive is not platform-specific.
- `check:viewer`, `check:brand-marks`, `check:validators`, `check:release-identity` and `test/golden.mjs` all pass unchanged; no Gallery, guide, start or readme-showcase regeneration was needed.


gold-beyond:
CI note on the `webm-artifact` job — it failed in the last step, `Verify Viewer chrome, camera, motion and input interactions`. I could not retrieve the job log (downloading job logs requires repository admin rights), so I cannot name the individual test that failed; here is what I could verify locally instead.

- **The same step passes on this exact head.** Ran it verbatim on `48b6d47` (Node v22.23.2, real Chrome): `node --test --test-concurrency=2 test/viewer-chrome-layout.test.mjs test/viewer-camera-browser.test.mjs test/motion-governor-browser.test.mjs test/finder-browser.test.mjs test/intent-trace-browser.test.mjs test/semantic-lens-browser.test.mjs test/route-probe-browser.test.mjs test/guided-views-browser.test.mjs test/focus-browser.test.mjs test/export-browser.test.mjs test/viewer-identifiers-browser.test.mjs` → 148 tests, 148 pass, 0 fail, 0 skipped.
- **The step has no call path into the changed code.** None of those eleven files import the workflow compiler or `migrations/`; the only renderer import among them is `renderers/shared/desktop-readability.mjs`. `guided-views-browser.test.mjs` operates on the packaged examples, which `test/golden.mjs` already renders byte-identically on this head.
- **For the renderer path this change is behaviorally a no-op.** `loadDiagram()` already ran `validateGuidedViews()` and then `validateRelationshipIds()` before compilation; the new shared entry point runs those same two functions in the same order, and the compiler-side call only extends that contract to direct `compileWorkflow()` callers (migrations and tests).

Every other check on this head is green, including `zip-freshness` and the full `test (18/20/22/24)` matrix. Could you re-run the `webm-artifact` job when convenient? I am glad to dig further if it reproduces.

tt-a1i (commented):
Reviewed `48b6d47` against current main `72c750b`. Reusing the existing cross-collection checks in the compiler is a proportionate fix for #429, and avoids keeping a second implementation in sync.

I reran `node --test test/workflow-compiler-hard-contract.test.mjs test/workflow-migration.test.mjs` from a clean snapshot on macOS with Node 26.8.1: 97 passed, no failures or skips.

I also reran the failed CI job. [Attempt 2](https://github.com/tt-a1i/archify/actions/runs/35204071727/attempts/2) passed on the unchanged head, including the Viewer/Motion and WebM checks. The earlier stored-mode assertion did not recur; this rerun doesn't establish why it failed the first time. All test and package jobs are now green, with Pages conditionally skipped.

I found no blocking issue in the compiler change.

sunsunsun-java:
Please update this PR against `dev`:

- Resolve conflicts and regenerate `archify.zip` from the combined source using Node 22.
- Narrow the shared-validator comment to relationship-ID and guided-view checks; it should not promise parity with all renderer validation.
- Update the regression test to compare `compileWorkflow()` with the real workflow renderer CLI, instead of only the shared validator.
- Rerun the tests and CI on the updated branch.

Issue #470, opened 2026-09-18
[Bug]: Windows automatic opener passes an empty FilePath to PowerShell


asts-top:
## Archify version or commit

`fdc5b183f743cbce1b6dacd07c43ecc224b05b78` (`dev`, `2.17.0-dev.1`)

## Installation method

Git clone

## Diagram type

Workflow (the failure is independent of diagram content and occurs in the shared Windows opener).

## Exact command

Save the JSON below as `opener-repro.workflow.json`, then run from the repository root in Windows PowerShell:

```powershell
$outDir = Join-Path $env:TEMP ('archify-opener-repro-' + [guid]::NewGuid().ToString('N'))
New-Item -ItemType Directory $outDir | Out-Null
$out = Join-Path $outDir 'diagram.html'
node .\archify\bin\archify.mjs deliver workflow .\opener-repro.workflow.json $out --open --json
```

The same helper handles preview's default browser opening; calling its loopback URL opener reproduces the same parameter-binding error.

## Minimal redacted JSON reproduction

The opener failure is independent of diagram content. This valid schema-v2 workflow is sufficient to produce a delivered artifact before the optional open step:

```json
{
  "schema_version": 2,
  "diagram_type": "workflow",
  "meta": { "title": "Windows opener reproduction" },
  "lanes": [
    { "id": "main", "label": "Main" }
  ],
  "nodes": [
    { "id": "source", "lane": "main", "col": 0, "type": "frontend", "label": "Source" },
    { "id": "target", "lane": "main", "col": 1, "type": "backend", "label": "Target" }
  ],
  "edges": [
    { "from": "source", "to": "target" }
  ]
}
```

Both this minimal fixture and the shipped `archify/examples/agent-tool-call.workflow.json` reproduced the failure through the unchanged CLI at the revision above. Both deliveries exited 0 and passed all 9 validation checks.

## Validation receipt or exact error

The delivery itself succeeds, but its optional open receipt reports failure. Relevant excerpt (other receipt fields omitted; target path redacted):

```json
{
  "ok": true,
  "open": {
    "requested": true,
    "status": "failed",
    "target": "C:\\Temp\\archify-opener-repro\\diagram.html",
    "method": "powershell"
  }
}
```

The CLI prints the following warning on stderr:

```text
Could not open the verified artifact (failed). Open it manually: C:\Temp\archify-opener-repro\diagram.html
```

[...]

PR #471, opened 2026-09-18
fix(opener): pass Windows targets through PowerShell environment

asts-top:
## Problem and value

Closes #470.

On Windows, appending the target after PowerShell `-Command` does not populate `$args[0]`. `Start-Process` receives a null `FilePath`, so `deliver --open` and preview's automatic opener fail even when artifact delivery succeeds.

Keep the PowerShell source constant and pass the target through `ARCHIFY_OPEN_TARGET` in the child environment. This handles file paths and loopback URLs without interpolating targets into executable source.

## Stability impact

- Shared opener behavior; the runtime change is limited to Windows. macOS/Linux argument arrays, `shell: false`, the five-second timeout, and optional-opener failure receipts remain intact.
- No dependency, version, schema, renderer or browser/UI changes.
- Revision `f96018c` merges current `dev` (`f3cc060`) in response to the [integration review](https://github.com/tt-a1i/archify/pull/471#pullrequestreview-5256452794). The opener source and tests are byte-identical to reviewed head `a2ca4d0`.
- The CI, release and package gates match current `dev` exactly. Its maintained Windows runner already includes all of `test/open-artifact.test.mjs`, so it runs the new real-PowerShell regression on Node 22 and 24 without retaining the obsolete test-name filter.

## Tests run

Comparison base: `f3cc060fd5f25810568b31e3897efa413faa4b5d`. Candidate: `f96018c1805a1a764cf10e4357e4c0c725db89ad`.

Fresh local checks on the combined source:

- Windows Node 22.12.0 and 24.21.0: all four opener tests passed on each version, including the real PowerShell launch of system `where.exe`; no skips.
- Current-base reproduction: delivery of both the shipped example and a two-node fixture succeeded but reported a failed opener. Four real-PowerShell parameter-recording cases failed on the base and preserved the exact targets after the fix (plain file, Unicode/spaces/apostrophes, PowerShell metacharacters and loopback URL). The small fixture includes `meta.output`, as required by current `dev`.
- CI/release workflow contract checks: 4 passed.
- Windows extracted-package smoke and a real launch through the packaged opener: passed.
- Linux package-gate suite: 38 passed, 2 platform skips. An initial local full run had 1977 passed, 76 skipped and one setup failure: the exported verification snapshot had no Git `HEAD`, which a new upstream package-index test requires. Recording a commit with the exact same Git tree resolved it; the complete package-gate suite above then passed. Full-suite coverage on the updated head is supplied by CI.
- Linux extracted-package smoke: passed. A fresh Linux Node 22 rebuild byte-matched the Windows-built ZIP; compared with current `dev`, only `archify/bin/open-artifact.mjs` differs inside the archive.
- `git diff --check`: passed.

[CI for the updated head](https://github.com/tt-a1i/archify/actions/runs/35458352879) completed successfully: all 12 validation jobs passed, including the complete Node 18/20/22/24 suites, controlled Windows path fixtures and opener regressions on Node 22/24, all three package-smoke platforms, browser/WebM, ZIP freshness and the update manifest. Pages deployment is intentionally skipped for pull requests.

Earlier full Windows comparison on `31bfbc8` had 37 matching baseline/candidate failures and no candidate-only failures; the new PowerShell case passed. This remains historical evidence, not a full Windows run of `f96018c`. The current controlled Windows CI lane is the integration check requested in review. Local launch evidence does not establish default-browser/file-association behavior.

## Generated artifacts

`archify.zip` was rebuilt with Node 22 from the combined source, rather than selecting either conflicting archive. ZIP SHA256: `c7ac4471bed395b000ee9546c4663c067a7bf32fc14357c97906d7d098cc6c1c`.


tt-a1i (commented):
The opener change at `a2ca4d` looks appropriately scoped: the PowerShell source stays constant and the target travels in the child environment. The existing Windows CI passed on that head. My macOS run passed the three applicable opener tests; the real PowerShell case was skipped here.

The remaining integration problem is concrete: merging current dev (`f3cc060`) conflicts in `.github/workflows/ci.yml` and `archify.zip`. Please update from dev, retain its current Windows path/package gates while adding the opener regression, and rebuild the ZIP from the combined source. Don't resolve the archive by choosing either branch's existing ZIP. Once the updated head has Windows CI, this can be checked again against #470.

PR #418, opened 2026-09-15
fix(renderer): isolate compiler imports from CLI diagnostics

NaCr05:
## Problem and value

Importing `compileWorkflow()` with `ARCHIFY_DIAGNOSTIC_FORMAT=json` installs a process-level exception handler through shared SVG helpers. A later exception belonging to the embedding host can then produce a renderer receipt and exit the process.

Install that boundary when `loadDiagram()` starts instead. Compiler imports preserve the host's exception policy; all five CLI entrypoints still establish diagnostics before reading or validating input.

Related to #222. This is the preparatory import-isolation fix, without introducing a public render API.

## Stability impact

- Shared behavior: diagnostic initialization used by all five renderer CLIs and indirectly imported by the workflow compiler.
- Preserve CLI failure exits, machine-readable diagnostics, existing artifacts on failure, and successful HTML bytes.
- Add regressions for host-owned exceptions and compiler calls with frozen inputs/prior results, alternating v1/v2 and configuration, and failure/success sequences.
- Loader validation, brand preparation and environment-independent library configuration remain outside this change.

## Tests run

Current base: `d15010a2d49f93388820f8a3b7474b77aef60006`. Current candidate: `8ae0d450b588fe5808959332f7296dfd7ae5ce5f`.

This revision merges main's benchmark visual-defect validation fix (#402). It adds no further import-isolation changes: the four-file PR diff against current main is byte-for-byte identical to the original patch at `72deba7`.

Fresh local checks use official Node 22.23.2. Linux checks run against the exact candidate source tree exported with `git archive`, with a temporary Git snapshot for tracked-file package checks and the committed ZIP unchanged.

| Check | Result |
| --- | --- |
| Linux full `npm test` | 1333 passed, 0 failed, 48 conditional skips; generation checks and golden harness passed |
| Linux import/call contracts plus updated benchmark suite | 32 passed, 0 failed, 0 skipped |
| Linux contracts with JSON diagnostics | 13 passed, 0 failed, 0 skipped |
| Windows import/call contracts plus benchmark suite | 24 passed, 8 failed, 0 skipped; all 13 import/call contracts passed |
| Same Windows benchmark suite on unchanged current main | 11 passed, 8 failed, 0 skipped; the eight failure names exactly match the candidate |
| ZIP freshness | Official Node 22.23.2 rebuild matches the committed archive byte-for-byte |
[...]

Issue #222, opened 2026-08-30
Expose renderers as an importable library API, not just a CLI entrypoint


joesaby:
**Problem**

Each renderer under `renderers/<type>/render-<type>.mjs` is written as a script rather than an exported function:

- Input is read via `loadDiagram({ argv = process.argv, ... })`, which resolves a path and does `fs.readFileSync` on it — there's no path to pass an in-memory IR object directly.
- The read + render happens as an unconditional top-level `await` the moment the module is imported, so `import`ing the renderer *is* running it against `process.argv`/disk.
- Output goes out through `writeDiagram()`, which writes to a file and `console.log`s the path — there's no return value carrying the rendered HTML or SVG string back to the caller.

For any Node program that wants to render a diagram programmatically (a build plugin, a test harness, a server-side render step), the only option today is to spawn `node render-<type>.mjs <input.json> <output.html>` as a subprocess and read the output file back — even though the renderer itself has no runtime dependencies beyond Node builtins and is otherwise a pure `IR in → HTML out` transform.

**Proposal**

Add a small library entrypoint alongside the existing CLI scripts, e.g.:

```js
import { render } from 'archify';
// or a per-type import: import { renderArchitecture } from 'archify/architecture';

const { html, svg, height } = await render({
  type: 'architecture',   // or infer from ir.diagram_type
  ir: parsedIrObject,      // already-parsed JSON, not a file path
  quality: 'showcase'
});
```

Suggested shape:
1. Split `loadDiagram`'s two responsibilities — "resolve argv into an input path" (CLI-only) vs. "validate + prepare a diagram object" (shared) — so the latter can run on an object that was never read from disk.
2. Give each renderer a real exported function instead of a top-level `await`, and have the current CLI script become a thin wrapper: `parse argv → call the exported function → writeDiagram(result)`.
3. Have the render function *return* `{ html, svg, cards, meta }` instead of only writing to disk, and let the CLI wrapper be the one that calls `writeDiagram`.
4. Keep `writeDiagram`/file output available for CLI use, but make it opt-in rather than the only path.

This wouldn't change the CLI's behavior or output format — it just adds an importable entrypoint alongside it, so Archify's renderers can be embedded directly in other tools' build pipelines without a subprocess round-trip.

joesaby:
Some more detail on implementation approach, from reading through the current `renderers/` code:

**Where the coupling actually lives**

[...]

Issue #408, opened 2026-09-13
Publish archify to the npm registry


sergeiwallace:
## Summary

**Feature Request:** Publish archify to the npm registry so it can be installed with ordinary
`npm`/`npx` instead of only through the third-party `skills` GitHub installer or a manual ZIP
extraction.

## Desired state

`npm install -g archify` (or a scoped equivalent such as `@tt-a1i/archify`) installs the current
release from the npm registry, and standard npm tooling (`npm outdated`, `npm update`, version
pinning in `package.json`) works against it like any other published package.

## Current state

`package.json` is `"private": true`, and the only supported install paths are the third-party
`npx skills add tt-a1i/archify -g` installer (which clones the GitHub repo directly) or manually
extracting the repo's own `archify.zip`. Neither uses the npm registry.

Worse, the unscoped name `archify` is **already registered on npm** by an unrelated, unmaintained
package (`archify@0.0.4`, `github.com/justin-calleja/archifier`, zero dependencies, last published
over a year ago). Anyone who reasonably tries `npm install -g archify` today gets that unrelated
package instead, with no warning. A scoped name (e.g. `@tt-a1i/archify`) would sidestep the
collision entirely and could still expose an `archify` bin via `package.json`'s `bin` field.

## Environment

- Product and version: archify 2.17.0-dev.1 (installed via `npx skills add tt-a1i/archify -g`, per
  the project's own documented install path)
- OS and version: Fedora Linux 44, x86_64
- Model / resolved model ID: N/A
- Runtime context: Node v22.23.1, Claude Code CLI skill install (`~/.claude/skills`)
- Authentication/provider path: N/A (public npm registry, no auth involved)

## Related public reports

None found. Searched open/closed issues in `tt-a1i/archify` for `npm`, `publish`, and `registry`;
no existing request for npm-registry distribution.

vipul156:
i want to work upon this

tt-a1i:
A normal npm install path would be useful, especially for version pinning and build tools. @vipul156, thanks for offering to help.

I'd start with a packaging proof of concept: a scoped package with the CLI, the assets it needs, and a clean-directory test of the packed tarball. Keep the current Skill install paths working. The proposed `@tt-a1i/archify` name still needs to be confirmed before publication, and npm installation shouldn't silently copy the Skill into an agent's directory.

Please post the proposed package contents and install command here first. That should let us settle the distribution shape before adding release automation.

vipul156:
[...]

PR #482, opened 2026-09-19
feat(workflow): add in-memory renderWorkflow() library entrypoint

joesaby:
## Problem

Every archify renderer script (`renderers/<type>/render-<type>.mjs`) is a one-shot CLI entry point, not an importable function: it reads input from `process.argv`/disk via `loadDiagram()`, runs its render as a top-level `await` the moment the module is imported, and — through the CLI-only diagnostics boundary `loadDiagram()` installs — calls `process.exit()` directly on both success and failure. That makes it unsafe to `import` for in-process rendering: a host process embedding archify would have its own process torn down by a bad diagram, not just the render call fail.

The real consumer that ran into this is my own [astro-archify](https://github.com/joesaby/astro-archify) — an Astro integration that turns fenced ```archify code blocks in markdown into rendered diagrams at build time. Because the renderer scripts can't be imported safely, it [vendors a full copy of `renderers/`](https://github.com/joesaby/astro-archify/tree/main/vendor/archify) and [spawns `node render-<type>.mjs input.json output.html` as a subprocess](https://github.com/joesaby/astro-archify/blob/main/astro-archify-integration.js#L169-L217) per diagram, reading the output file back afterward — a process-spawn + temp-file round trip for what's otherwise a pure `IR in -> HTML out` transform with no runtime dependencies beyond Node builtins.

## What this PR does

Workflow-first PoC for #222, per the direction tt-a1i confirmed [here](https://github.com/tt-a1i/archify/issues/222#issuecomment-5743474339). Adds `renderWorkflow()` in `archify/renderers/workflow/workflow-api.mjs`: an async, importable function that runs the same validation and resource preparation as the CLI's `loadDiagramWithBrandMarks()` — schema, guided views + relationship IDs (via `validateCrossCollectionContracts`, the shared entry point #462 introduced), engineering profile, repository evidence, and brand-mark capture — then calls `compileWorkflow()`, and returns `{ html, svg, cards, meta, sourceEvidence, receipt }` to the caller instead of writing a file.

- A classified, author-facing failure comes back as `{ ok: false, diagnostics }` instead of throwing. The CLI-only process-level diagnostics boundary (`installRendererDiagnosticBoundary`, #418) is never installed by this path, so a bad document cannot take down the host process.
- Each call `structuredClone`s its input, so the caller's document is never mutated and concurrent/repeated calls stay independent.
- Nothing here writes a file, logs to console, or sets `process.exitCode` — delivery stays owned by the caller. A render result never claims delivery completed.
- `writeDiagram()`'s template-fill step is extracted into a small shared `renderDiagramHtml()` in `shared/cli.mjs`, used by both the CLI and this new library path. `render-workflow.mjs` (the CLI script) is otherwise untouched.

**Two host-isolation fixes from review, credited to @NaCr05:** NaCr05 independently reviewed the first version of this PR and published a [handoff patch](https://github.com/NaCr05/archify/blob/993119b0dcf3389f7096666d2172f5d8a35fa348/contribution-handoff/482/README.md) identifying two real bugs. I re-implemented both myself (verified each fails without its fix, using their analysis as the starting point) rather than apply the patch directly:
1. `svgRootAttrs()` fell back to reading `process.env.ARCHIFY_QUALITY_PROFILE` whenever no quality profile was otherwise resolved — so a host process with that env var set for unrelated reasons could silently override an "advisory" `renderWorkflow()` call the caller never asked to be pinned. Fixed with a `profileIsAuthoritative` flag, opted into only at the workflow compiler's own call site; architecture/sequence/dataflow/lifecycle's existing direct `svgRootAttrs()` calls are untouched.
2. `withDiagnosticRecordingSuppressed()` used a plain module-level depth counter, unsafe once the suppressed callback is async (its `finally` fires when the callback's *synchronous* portion returns, not when the returned promise settles) — exactly the shape `renderWorkflow()` has, since it awaits `prepareDiagramBrandMarks()`. Under `ARCHIFY_DIAGNOSTIC_FORMAT=json`, a failed library call's diagnostics could leak into a later, unrelated CLI failure report in the same process. Fixed by switching to `AsyncLocalStorage`, scoped per call chain, and wrapping `renderWorkflow()` in it.

Both fixes are scoped narrowly and verified not to change any existing CLI behavior (golden output byte-identical, full existing suite passes).

## What this doesn't do yet

This covers `workflow` only, because `workflow-compiler.mjs` already exposes `compileWorkflow()` as a function separable from the CLI plumbing — the other four renderers (architecture, sequence, dataflow, lifecycle) don't have that split yet, so they still need the subprocess path. A consumer like astro-archify, which renders all five diagram types through one generic path, can't retire its vendor+subprocess pattern until the same shape extends to the rest — that's the natural next slice once the API shape here settles, and astro-archify itself isn't touched by this PR.

## Test plan

- [x] `test/workflow-library-api-call-contract.test.mjs`: input validation, host isolation (no mutation, concurrent calls independent), never installs the process-level diagnostics boundary, guided-view/relationship-ID contracts, and returns compiler-level failures as data rather than throwing.
- [x] `test/workflow-library-api-quality.test.mjs`: `renderWorkflow()` stays advisory under an ambient `ARCHIFY_QUALITY_PROFILE` the caller never asked for, while an explicit `qualityProfile` argument still takes effect. Verified this test fails without the `svgRootAttrs` fix.
- [x] `test/workflow-library-api-diagnostics.test.mjs`: a failed `renderWorkflow()` call under `ARCHIFY_DIAGNOSTIC_FORMAT=json` does not leak into a later, unrelated CLI diagnostic report. Verified this test fails without the `AsyncLocalStorage` fix.
- [x] `node --test test/workflow-*.test.mjs` — all 166 workflow tests pass. `node --test test/render-failure-diagnostics.test.mjs test/renderer-diagnostic-boundary.test.mjs test/renderer-import-isolation.test.mjs` — all pass (confirms the `diagnostics.mjs` change doesn't regress the CLI boundary for any of the 5 renderers).
- [x] `node test/golden.mjs` — golden CLI output is byte-identical across all 5 diagram types.
- [x] Rebased onto `dev` after #462 merged; full `npm test` run shows only pre-existing, environment-specific failures (verified identical against a clean, unmodified `dev` checkout) plus two `parse5`-module-not-found failures traced to this sandbox's incomplete `node_modules` install — none touch any file this PR changes.


NaCr05:
Hi @joesaby — I’ve published the Workflow API review fixes and resource-contract coverage for you to fold into #482:

- [Handoff notes, patch and validation logs](https://github.com/NaCr05/archify/blob/993119b0dcf3389f7096666d2172f5d8a35fa348/contribution-handoff/482/README.md)
- [Download the cumulative source patch](https://raw.githubusercontent.com/NaCr05/archify/993119b0dcf3389f7096666d2172f5d8a35fa348/contribution-handoff/482/cumulative-source.patch)
- [Exact tested source: `c9933c4`](https://github.com/NaCr05/archify/tree/c9933c471145d3152cb2d59782bf4c97b65978fb), on `NaCr05/archify:test/workflow-library-resource-contracts`

The patch covers four related items:

1. Keep the compiler’s effective quality policy and SVG metadata aligned when the host sets `ARCHIFY_QUALITY_PROFILE`. Unspecified/advisory behavior and the CLI’s environment handling are preserved.
2. Keep completed library failures out of global CLI diagnostic history in JSON mode, including errors after asynchronous brand preparation. Per-call diagnostics remain available, and unrelated concurrent CLI recording is preserved.
3. Remove the assertion that requires the compiler to accept duplicate relationship IDs, and reuse the shared cross-collection validator now landed through #462.
4. Add 12 resource-contract tests through the public API and real renderer CLI: default built-in brands, frozen/repeated inputs, pinned repository evidence and its HTML payload, invalid documents, unknown brands, and repository failure cases. Successful HTML and failure diagnostics are compared directly.

I performed an actual local merge with `dev` at `5769206`, which includes #462. The only conflict was `archify.zip`; I rebuilt it from the combined source with Node 22. The cumulative patch was also applied to a separate original-#482-plus-dev checkout, and the resulting tracked files (excluding the ZIP) matched the tested source.

Local validation:

| Check | Result |
| --- | --- |
| Windows Node 22 library/compiler contracts | 99 passed |
| Linux Node 22 full `npm test`, including freshness/golden checks | 2,045 passed, 0 failed, 76 skipped |
| Windows Node 24 extracted-package library tests | 23 passed; no node_modules installed, Git available for evidence checks |
| Canonical ZIP rebuild and Linux extracted-package smoke | Byte-identical rebuild; smoke passed |

The [linked logs](https://github.com/NaCr05/archify/blob/993119b0dcf3389f7096666d2172f5d8a35fa348/contribution-handoff/482/README.md#validation) identify the tested revisions and skipped gates, including real-browser tests. They do not replace hosted CI on your final integrated branch.

Suggested next steps: sync #482 with dev containing #462, apply the source patch, regenerate `archify.zip` from that final tree with Node 22, then run CI and request maintainer review. This keeps #482 as the upstream contribution. The handoff-notes branch is only for sharing evidence; the patch contains the proposed runtime/test changes. No redesign of `meta.output` or brand-resource options is included.

joesaby:
Thanks for the thorough review, @NaCr05 — both bugs are real and well-diagnosed. I re-implemented them myself rather than apply the patch directly (I wanted to independently reproduce each one and verify the fix), but the root-cause analysis is entirely yours:

- `svgRootAttrs()`'s `ARCHIFY_QUALITY_PROFILE` env fallback leaking into an in-process call that never asked for it
- `withDiagnosticRecordingSuppressed()`'s module-level depth counter being unsafe for the async caller `renderWorkflow()` actually is

Both are fixed in ef6d7b1, with a test per bug that I verified fails without its corresponding fix (reverted each change locally, confirmed the test catches it, then restored). Details and full test-plan in the updated PR description.

I didn't pull in the 3rd/4th items from your patch (the "no redesign of `meta.output`, brand-resource options" tests) since they read more like additional coverage than bug fixes — happy to add more tests if there's a specific gap you still see after this update. Also didn't carry over the resource/CLI-parity test file wholesale, since a couple of assertions in it (the `explicit-pin-conflict` quality-profile probe) test the pre-existing compiler contract rather than anything this PR changes — good regression coverage in general, but felt out of scope for this PR specifically.

Rebased onto current `dev` (post-#462) and re-verified everything — golden output, the full workflow/diagnostics suites, and the two new regression tests against their unfixed baselines. Description above has the full breakdown.

tt-a1i:
Thanks @joesaby for the implementation, and @NaCr05 for the review work. We welcome this workflow-first contribution. The existing CLI already supports external callers, and npm distribution in #408 can move forward independently. This API is a useful convenience for in-process consumers; there is no need to expand this PR to all five renderers or make it a prerequisite for npm packaging.

I reviewed `62940ddabeece27028bdedb0dede658df8da72b2` and tried the packaged API through a local adapter in astro-archify at `1f5e15b210094f9f94ea95ef5c2b685fc810468a` (Astro 6.4.8, Node 22.23.2). Five pages with nine distinct diagrams produced byte-identical CLI/API site output. This included the other four types through the existing subprocess path, branded workflows, repeated diagrams, and invalid input followed by a valid diagram. Both strict builds failed on the expected invalid input; non-strict builds recovered. Browser checks covered the brand badge, theme toggle, and node search/focus. The 36 focused API/diagnostic/import tests and golden harness also passed locally.

There is one concrete issue to fix before merge: `prepareBrandMarks: false` silently drops authored brands. A valid `brand: "openai"` returns success without its badge; unknown brands also become successful unbranded output. Preparing resources before the call does not solve this, because `structuredClone` loses the prepared resource association. For this first slice, please remove the skip option and always prepare resources. Add focused public-API coverage for valid brands and brand failures; a broader resource-contract redesign is unnecessary.

Please also add a short usage example covering results/errors and resource I/O. Direct in-process calls do not retain the existing subprocess's kill-on-timeout isolation, so callers needing that should keep a subprocess or worker boundary.

Our small build sample showed no obvious performance regression, but it was not a controlled benchmark and does not establish a speedup. We'd welcome your own reproducible downstream trial, including any noticeable performance or behavior changes. Once the brand issue and API documentation are addressed, we can re-review this focused slice.

NaCr05:
Hi @joesaby and @tt-a1i — I followed up on the brand-resource correction against the current head `62940dd` and prepared a focused patch:

- [Patch, reproduction and validation logs](https://github.com/NaCr05/archify/blob/50dd041992a7d76c8d840dec8790558fbcd9d73e/contribution-handoff/482/brand-follow-up/README.md)
- [Source-only patch](https://raw.githubusercontent.com/NaCr05/archify/50dd041992a7d76c8d840dec8790558fbcd9d73e/contribution-handoff/482/brand-follow-up/brand-source.patch)
- [Tested source with regenerated ZIP](https://github.com/NaCr05/archify/tree/152c652dc2f879357d4fb9a529c5aa243f8d9a40)

The change removes the skip option and always prepares the cloned document. Six focused public-API tests cover valid/unknown brands, CLI HTML/diagnostic parity, caller-prepared inputs, and repeated/concurrent calls. On unchanged `62940dd`, three regressions fail: skipping drops valid brands, accepts unknown brands, and loses caller-prepared brands during cloning. All six pass with the fix. The README adds an executable results/errors example and documents resource I/O and timeout-isolation boundaries.

Validation on official Node 22.23.2:
- 46 API/compiler/import/diagnostic checks pass on Windows and Linux; golden checks pass.
- Extracted-package API checks: 15/15; canonical ZIP rebuild and package smoke pass.
- Local merge rehearsal with current dev `0b8b955`: 2047 passed, 0 failed, 76 skipped in full Linux `npm test`; extracted-package checks also pass 15/15. Only the ZIP conflicted and was regenerated.
- The README example was run against the extracted package for valid and invalid brands.

You can apply the focused source patch or use its regression cases for your implementation. It excludes ZIP bytes so the archive can be rebuilt from your final combined branch. This is a follow-up for #482; your final-head hosted CI and downstream acceptance still need to run. The earlier broader handoff remains historical and should not be reapplied wholesale.

tt-a1i:
Thanks @NaCr05 for preparing the focused brand-resource fix, regression cases, and usage documentation. This follow-up addresses the scope we requested, and the reported cross-platform checks are helpful.

@joesaby, would you be able to integrate this focused follow-up into #482? You can apply the source patch or use the reproductions to implement the correction yourself; please retain attribution for NaCr05's contribution. There is no need to pull in the earlier broader handoff or expand beyond the Workflow API.

If you need help with integration or regeneration, let us know and we can coordinate. Once the changes are on the PR, we can review the final integrated version, its CI results, and the downstream behavior. Thanks to both of you for moving this forward!

joesaby:
Ran the reproducible downstream trial you asked for, against astro-archify's own vendor + test fixtures (bumped to this PR's branch, on a separate local branch — not pushed).

**Correctness — byte-identical** between the existing subprocess path and `renderWorkflow()`:
- `agent-tool-call.workflow.json`, default and `quality=showcase`: exact byte match (842,132 bytes).
- Valid brand (`openai`): badge renders identically on both paths.
- Unknown brand: both paths correctly fail (confirms the brand-marks fix — no silent-success regression).
- Invalid legacy fixture (missing `meta.output`): both paths fail with the identical diagnostic.

**Performance** — 20 repeated renders of the same diagram:
- subprocess: 196.68ms/call
- `renderWorkflow()`: 28.65ms/call
- **6.87x speedup**

**One unrelated finding:** astro-archify's own test suite has 3 pre-existing failures (`meta.output` now required by schema) that I bisected and confirmed reproduce identically on plain current `dev` with zero `renderWorkflow` changes — schema-strictness drift since astro-archify's vendor was last pinned, not caused by this PR. Flagging for visibility, not something to fix here.

joesaby:
@tt-a1i — done. NaCr05's brand-marks follow-up is integrated (\`a5630dd\`, \`Co-authored-by: NaCr05\`), exactly as scoped: the \`prepareBrandMarks\` skip is removed, brand prep always runs, no broader handoff pulled in, no expansion beyond the Workflow API.

Also rebased the branch onto current \`dev\` to resolve the merge conflict that came up in the meantime — all CI green. Downstream trial results (byte-identical output, brand parity, 6.87x perf) are in the comment above.

joesaby:
@tt-a1i 

One process note for whoever merges this: NaCr05 is credited via a \`Co-authored-by\` trailer in \`a5630dd\`'s commit body (they're not a git commit author on any commit here, so GitHub's automatic co-author detection on squash-merge won't pick them up on its own). If this gets squash-merged, please keep that trailer in the final commit message so the attribution survives — happy to combine it into a single clean commit message if that's easier.

tt-a1i:
Thanks @joesaby for integrating the brand fix and running the downstream trial, and @NaCr05 for the review and patch. I reviewed the current `2cb409e` head: the focused library tests (15/15), existing renderer/diagnostic tests (27/27), and five-type golden checks passed locally on Node 22; the current hosted checks are green. I did not find a technical blocker in that pass. We also noted the `Co-authored-by: NaCr05` trailer and would preserve it if this is eventually squash-merged.

Before deciding to make an in-process entrypoint part of Archify's supported surface, we'd like to settle the distribution path. Our near-term preference is the scoped, CLI-first npm package discussed in #408. That would let astro-archify replace its vendored copy and manual sync with a versioned dependency while keeping the subprocess boundary it already uses. We don't want a separate integration package to duplicate Archify's renderer, validation, and brand logic. We do see the value of this Workflow API and your measured per-render speedup; the remaining question is how much direct invocation helps in a representative full build, and when it is worth taking on a long-lived library contract.

So our current leaning is to keep #482 open while the #408 package proof is tested downstream, without asking you to expand this PR or redo its implementation. I'm sorry this is a sequencing change after we asked you to integrate the follow-up. Your work and NaCr05's contribution are valuable, and the technical review above still stands.

How does that sequence look from your side? Would a versioned CLI package solve the main pain in astro-archify, or is there a real use case that needs the in-process Workflow API sooner, apart from the repeated-render benchmark? Any build-level numbers or constraints you already have would help us make that call; no need to manufacture a new benchmark before the tarball exists.

